PR #36, opened 2019-09-27
Add CMake support

f-tischler:
Added support for building with CMake aswell as a template project which integrates WickedEngine using find_package().


turanszkij:
Hi,

What is your purpose by adding CMake? I'm happy to support only one version of Visual Studio at the moment. And to be honest, I didn't want to support CMake, as it is just an other complication and dependency that needs to be maintained. 

I glanced through the commits quickly, and saw that you also added compiled shader objects into the source control, which I want to avoid.

Sorry if that sounded rude, that was not my intention. I am happy that someone is interested in the project and we can continue the discussion.

turanszkij:
And some other concerns related to CMake:

- Currently, the way to build is simple: Open solution, then press F5. Building with CMake will be much more complicated, can it be optional?
- I am not used to CMake, not interested to learn. Someone needs to support it as project grows.
- Project needs to build for 32-bit, 64-bit, ARM. For all projects, including Windows 10 and UWP version. There are also HLSL shaders. All of these would need CMake configuration and testing.
- Need to keep supporting the Appveyor nightly build system

f-tischler:
No no, thats fine. It is your project and you decide whats added and what not.

Uh, wasn really my intention to add the shader objects. You are right, that does not belong into source control.

The idea was thst many people use CMake as their (meta) build tool as do I. By providing CMake support they feel right at home and have many possibilities to add your engine to their project. Building with Visual Studio Projects is cumbersome when it comes to adding dependencies. You have to manually add paths and libraries to use other libraries. CMake handles that way more elegantly by using find_package().

Actually, adding CMake support does not add any dependences. It should not add much work at all

f-tischler:
With Visual Studio 2019 (and 2017 for that matter) it is the same: Open Folder where the CMakeLists.txt resides and press F5. Yes, it can be optional. My changes allow to choose between building with Visual Studio directly or with CMake

I was in the same position some years ago. I was completely fine with Visual Studio Solutions and such. But now I am happy that i learned CMake, its a great alternative in some cases. Since I am planning to eploy your engine at work, I can also support CMake as long as I use it.

I would need to look into that but building for ARM should not be a problem. About UWP and HLSL i am not sure.

Since it is optional, it does not affect AppVeyor builds. I can offer you to set up a CMake nightly build in TravisCI

f-tischler:
It is just an offer from my side, if your are not fine with it Ill keep it in my fork and won't bother you with it anymore. But I think there are people which would appreciate it like they would appreciate linux support

turanszkij:
I see, it is quite easy to use CMake from Visual Studio. There are some issues however:

- Simply pressing F5 didn't do anything, I had to select for example "Editor.exe" from the top menu as startup project. That's not a big issue, I think I need to specify a startup project right now as well.
- Then, the shaders were not building, so immediately I get errors when the program starts.
- The asset paths/working directory are not correct, so assets won't load like this
- target platform configurations are not set up properly (x86/x64/ARM)
- I noticed that you also added External_Template_Windows/fonts/arial.ttf again, just like the shader binaries. These should be in the WickedEngine folder, and implementing projects reference them instead. 
- When building a project, Visual Studio output keeps spamming huge text for each source file which makes it very hard to read.
- I no longer see the filters I have set up in Visual Studio when I'm browsing a cmake project. This is important, because I manage the solution with filters, instead of folders where I can. This simplifies c++ include files a lot.

Sorry, I won't be merging in the current state. If you are fine with using Cmake just in your project, please do. If you bring it up to be more complete and fix these issues, I will look at it again.

f-tischler:
> Simply pressing F5 didn't do anything, I had to select for example "Editor.exe" from the top menu as startup project. That's not a big issue, I think I need to specify a startup project right now as well.

Yep, thats right but its the same when opening the Visual Studio Solution for the first as the startup project is part of the user config of VS which is/should not be added to source control

> Then, the shaders were not building, so immediately I get errors when the program starts.
Ah yes, I will look into that. Did not recognize that these are built by VS because I build it via the solution first before adding CMake.

> The asset paths/working directory are not correct, so assets won't load like this (well, this would also be responsible for the above issue, as shaders are assets)

The asset directories are curently a hardcoded, relative paths which is not good. Projects outside of the root folder need to adjust this manually or by copying the assets to their own directory. For the template project I did the latter but I see that for the template project the first approach may be better.

> target platform configurations are not set up properly (x86/x64/ARM)

This is inherent with how CMake works. You select the target architecture when configuring with CMake and  I added instructions on how to do that in the main ReadMe. It would be possible to also include the CMakeSettings.json into source control but thats not optimal.

> When building a project, Visual Studio output keeps spamming huge text for each source file which makes it very hard to read.

This is also configurable through the CMakeSettings.json and is a user preferences which should not go into source control

> I no longer see the filters I have set up in Visual Studio when I'm browsing a cmake project. This is important, because I manage the solution with filters, instead of folders where I can. This simplifies c++ include files a lot.

Yes that is true and is also inherent with the open folder approach of the CMake integration in VS. For most people this does not matter as they will only open the project, build it and run the editor for example. I get that you like them and thats why I would see the CMake integration as an optional feature for the people who need it, living side-by-side with the original Solution. Also it would make it way easier to port the engine to Linux once the Vulkan backend is stable. I would be happy to assist you with porting if you like.

> Sorry, I won't be merging in the current state. If you are fine with using Cmake just in your project, please do. If you bring it up to be more complete and fix these issues, I will look at it again.

That is completely fine, the purpse of the PR is to find and fix problems before thinking about merging it. I will fix everything that is fixable and then it would be great if you can take a look it again.

Thanks for the feedback.

turanszkij:
> The asset directories are curently a hardcoded, relative paths which is not good. Projects outside of the root folder need to adjust this manually or by copying the assets to their own directory. For the template project I did the latter but I see that for the template project the first approach may be better.
That is intentional. The only assets the engine uses are shaders and fonts. Those paths are configurable per application with wiRenderer::GetShaderPath() and wiFont::GetShaderPath() which are writable strings. Any other asset should be responsibility of the application project. For example there are images inside the Editor project which are specific to Editor, but the editor also uses engine assets such as shaders and the arial font.

I like how you see this thing as something optional. I like that you are also doing a sample project for it that shows how to use CMake.

As for the linux port, yeah that would be nice. But there is not only the Vulkan issue, but input and sound as well, and probably many more things I don't see coming. There is an issue for that already.

PR #60, opened 2019-11-22
Add LINESTRIP PRIMITIVETOPOLOGY

amerkoleci:
Add LINESTRIP PRIMITIVETOPOLOGY and remove UNDEFINED_TOPOLOGY as vulkan has no concept of undefined topology.


turanszkij:
Actually, there is a problem with this. Removal of UNDEFINED_TOPOLOGY was not working correctly in DX11 and DX12. Graphics glitches and debug layer warnings were present. I will fix this, next time please run the tests with debug layer enabled and chack for issues.

amerkoleci:
Vulkan doesn't have that behavior, need to dig what is the correct behavior, oh, probably it is pointlist by default

turanszkij:
I just readded UNDEFINED to the enum, and it works fine. The problem is the guards that are setting this in BindPipelineState() only if it actually changes for dx11 and dx12. Can be improved, but will need more testing.

PR #62, opened 2019-11-22
Use standard types where possible instead of WIN32 types.

### LINKED ISSUE #54: Refactor WinAPI int types and use stdint instead.
Types like UINT should be uint32_t, etc. to follow c++ standards instead of WinAPI types. There are a lot more types like this, ULONG, LONG, BOOL, FLOAT, etc, remove all.

WinAPI and DirectX code should however keep to use and cast to these types.

amerkoleci:
Fixes #54

### DIFF
--- a/WickedEngine/wiDirectInput.cpp
+++ b/WickedEngine/wiDirectInput.cpp
@@ -27,7 +27,7 @@ BOOL IsXInputDevice(const GUID* pGuidProductFromDirectInput)
 	BSTR                    bstrClassName = NULL;
 	DWORD                   uReturned = 0;
 	bool                    bIsXinputDevice = false;
-	UINT                    iDevice = 0;
+	uint32_t                    iDevice = 0;
 	VARIANT                 var;
 	HRESULT                 hr;
 
--- a/WickedEngine/wiGraphicsDescriptors.h
+++ b/WickedEngine/wiGraphicsDescriptors.h
@@ -324,7 +324,7 @@ namespace wiGraphics
 
 	// Structs /////////////////////////////////////////////
 
-	struct ViewPort
+	struct Viewport
 	{
 		float TopLeftX = 0.0f;
 		float TopLeftY = 0.0f;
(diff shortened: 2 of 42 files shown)


WickedEngine/wiDirectInput.cpp
```diff
@@ -27,7 +27,7 @@ BOOL IsXInputDevice(const GUID* pGuidProductFromDirectInput)
 	BSTR                    bstrClassName = NULL;
 	DWORD                   uReturned = 0;
 	bool                    bIsXinputDevice = false;
-	UINT                    iDevice = 0;
```
turanszkij: Let's keep it for DirectInput as this is what DirectInput uses natively.

WickedEngine/wiGraphicsDescriptors.h
```diff
@@ -553,42 +553,42 @@ namespace wiGraphics
...
 	};
 	struct IndirectDrawArgsIndexedInstanced
 	{
-		UINT IndexCountPerInstance = 0;
-		UINT InstanceCount = 0;
-		UINT StartIndexLocation = 0;
+		uint32_t IndexCountPerInstance = 0;
+		uint32_t InstanceCount = 0;
+		uint32_t StartIndexLocation = 0;
 		INT BaseVertexLocation = 0;
```
turanszkij: INT should be int32_t

turanszkij: ..can be fixed later though

WickedEngine/wiDirectInput.cpp
turanszkij: ..probably DirectInput wrapper will be refactored anyway, so not important.

turanszkij:
Actually, when testing there turned out to be a problem with GPUQueryResult, those BOOL -> bool refactors lost the 4 byte size of the BOOL and the queries were written incorrectly. This caused occlusion queries to malfunction. This should be a lesson for me too not to merge these without testing first.

amerkoleci:
Was indeed a little bit sceptic about GPUQueryResult as well, maybe instead of bool we could use uint32_t to fit in size of 4

turanszkij:
I removed the bools, occlusion predicate will just have its result in result_passed_sample_count, and not really interested in result_disjoint.

PR #81, opened 2020-01-21
A new implementation for Screen Space Reflections

Kliaxe:


WickedEngine/ShaderInterop_Renderer.h
```diff
@@ -231,6 +231,9 @@ CBUFFER(FrameCB, CBSLOT_RENDERER_FRAME)
 	float3		g_xFrame_WorldBoundsExtents;			float pad2_frameCB;		// world enclosing AABB abs(max - min)
 	float3		g_xFrame_WorldBoundsExtents_rcp;								// world enclosing AABB 1.0f / abs(max - min)
 	float		g_xFrame_CloudSpeed;
+	
+	uint		g_xFrame_TemporalAAEnabled;
```
turanszkij: Could you put this on a previous padding value instead? For example replacing the pad0_frameCB in a couple lines above. And emove the new padding you added. 
Also, (uint, float3) will not pad correctly in Vulkan SPIRV. Instead (float3, uint) should be used. But if you just replace a previous padding you don't have to mind this.

WickedEngine/ssrCS.hlsl
```diff
@@ -6,6 +6,9 @@ TEXTURE2D(texture_lineardepth_minmax, float2, TEXSLOT_ONDEMAND1);
 
 RWTEXTURE2D(output, float4, 0);
 
+//#define USE_RAYMARCH
+
+#ifdef USE_RAYMARCH
```
turanszkij: I tried it, looks amazing. Slower than the old one but much nicer. Can add sliders later to reduce sample count and other params.

Kliaxe:
Some changes have been made with the padding and version. Everything should be set up now! :)

turanszkij:
Thanks a lot for this cool update!

Issue #80, opened 2020-01-21
SSR Rework


Kliaxe:
Hi turanszkij!

So i stumbled upon your engine not too long ago and I got to say, your engine is a huge inspiration! It's well written, and has an amazing engine structure which is easy to understand.

I played a bit around with the build, and wanted to rework the current Screen Space Reflections. The current SSR in the engine is pretty simple, and I wanted to take it a step further. It's another way of doing SSR, and thought if you may be interested :)

turanszkij:
Hi, 
That's a great idea! Actually, recently I was looking at implementing the hierarchical Z-tracing approach, but haven't got it working yet. What method are you interested in? 
For reference, I was reading this on HiZ tracing: https://github.com/greje656/Questions/blob/33aa607f6bb0847331ce09ba56981e30f69920fd/hiz.md
It has implementation code too at the bottom. This is using a depth mipchain to skip a lot of empty space.
It this engine currently there is a linear depth mipchain (texture_lineardepth_minmax), but they are using regular depth mip chain (only min) I think.

Let me know what you think and thanks for stopping by.

Kliaxe:
Hi again, and thanks for the quick answer :D

I was looking at an implementation based on 2D marching, from this article from Morgan McGuire and Michael Mara's "GPU Screen-Space Ray Tracing":
http://casual-effects.blogspot.com/2014/08/screen-space-ray-tracing.html
It uses a "Digital Differential Analyzer" (DDA) line algorithm, with perspective-correct interpolation, which really caught my attention.
I played around with their algorithms and ended with something that looks really promising, in the current version.

turanszkij:
Yes, I'm familiar with it. If you feel like integrating it here, feel free to. I would be curious to try it.

Kliaxe:
Pull request should be up now. Hopefully it should work for you like it did on my setup, without any weird behaviors. *fingers crossed*

turanszkij:
Yes, it works perfectly, I only have one minor request before merging, see the pull request. Thanks!

Kliaxe:
Finally had the time to make the small changes. Pull request should be updated now.
Btw, I'm really glad I could contribute to the engine!

turanszkij:
Merged in #81 
Thanks for contributing! I will leave this open for a while. If me or someone else finds a way to optimize it or compare with a hierarchical depth approach, I'll post here.

Kliaxe:
[...]

PR #85, opened 2020-01-24
SSR bug fix

Kliaxe:
A quick bug fix for the SSR. The way of dealing with small artifacts didn't work well with other far planes, which caused nasty artifacts.


turanszkij:
Thanks for the fix!
#80

PR #87, opened 2020-02-09
Set the correct type on the texture returned from GetBackBuffer

BenV:
This resolves the assertion error caused when trying to take a screenshot. I'm fairly new to the codebase so please let me know if there is a better/more appropriate way of fixing this.


turanszkij:
This is the appropriate way to fix his, thanks a lot!

PR #96, opened 2020-03-10
Stochastic Screen Space Reflections

Kliaxe:
Everything should hopefully compile without any problems.

A quick thing to mention is the CopyTexture2D function produces NAN which can be a problem in this case. I didn't change the issue in this pull request because you probably might change a few things, but do notify me if there is anything.


turanszkij:
This is truly excellent work, very nice effect and you followed the coding style to maximum extent. 
I've opened a pull request into your repo with some minor fixes, if you merge that, hopefully this pull request will update and I will merge it to the master.

Kliaxe:
Nice catch! Everything should now be set!

turanszkij:
Indeed! Appreciate all the effort!

PR #100, opened 2020-03-12
Minor SSR improvements

Kliaxe:
Should hopefully compile fine.


WickedEngine/stochasticSSRCS_raytrace.hlsl
```diff
@@ -224,7 +224,7 @@ void main(uint3 DTid : SV_DispatchThreadID)
     // Everything in view space:
     const float3 P = reconstructPosition(uv, depth, g_xCamera_InvP);
     const float3 N = mul((float3x3)g_xCamera_View, decodeNormal(texture_gbuffer1.SampleLevel(sampler_point_clamp, uv, 0).xy)).xyz;
-    const float3 V = normalize(P);
+    const float3 V = normalize(-P);
```
turanszkij: Did you intend this? Reversed the V, then you reverse it again after this two times.

Kliaxe: Yes it was intentional, I just didn't want to confuse anybody with the way I did it before. The view direction was incorrect, but the result was still the same.
From my understanding, you'd calculate view direction from world position like so:
`normalize(cameraPosition - worldPosition)`
but since the camera position is always at (0, 0, 0) in view space we would simply do this:
`normalize(-viewPosition)`

turanszkij: Makes sense!

PR #107, opened 2020-04-19
Use wiResource structs instead of texture pointers for deferred mip g…

BenV:
…eneration

Taking a quick stab at fixing a crash I was running into pretty rarely with deferred mipmap generation. I modeled this after the [bvh fix](https://github.com/turanszkij/WickedEngine/commit/a92d5e7665776bf9eb60e0e7f00f952ef0647c7a#diff-7f6c8a0f0c1e358ba75c62db5bc000acL77) to use wiResources instead of raw texture pointers

A stack trace of the crash is attached.
deferred-mip-crash


turanszkij:
This is good, thanks!

PR #114, opened 2020-05-18
Linux

Je06jm:
Started the initial Linux port. It currently builds correctly, but has no system depended functionality and will not run. See issue #9


CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+cmake_minimum_required(VERSION 3.13)
+project(WickedEngine)
+
+set(CMAKE_CXX_STANDARD 17)
```
nemerle, line 4: Could also add 
```cmake
set(CMAKE_CXX_STANDARD_REQUIRED ON)
```

CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+set (WK_INC_DIRS "")
+foreach (_headerFile ${WK_HEADERS})
+    get_filename_component(_dir ${_headerFile} PATH)
+    list(APPEND WK_INC_DIRS ${_dir})
+endforeach()
+list(REMOVE_DUPLICATES WK_INC_DIRS)
+
+add_definitions(-DNDEBUG -DCMAKE_GENERATOR_PLATFORM=x64)
+add_library(WickedEngine ${WK_SOURCES} ${WK_SOURCES_C})
+target_include_directories(WickedEngine PRIVATE ${WK_INC_DIRS} ${VULKAN_SDK_DIR}/include /usr/include/vulkan)
```
nemerle, line 21: could also consider adding the following bit to help out when using CMake + MSVC 
```cmake
set_property(GLOBAL PROPERTY USE_FOLDERS ON) # enables project folder support on MSVC
set(AUTOGEN_TARGETS_FOLDER ON) # on MSVC puts all auto targets into a single folder - helps organize things.
# re-create the directory layout in generated MSVC projects
source_group(TREE ${CMAKE_CURRENT_SOURCE_DIR} FILES ${WK_SOURCES} ${WK_SOURCES_C} ${WK_HEADERS}) 
```

WickedEngine/RenderPath3D.cpp
```diff
@@ -402,8 +402,12 @@ void RenderPath3D::RenderReflections(CommandList cmd) const
 		}
 
 		wiRenderer::SetClipPlane(water, cmd);
-
+#ifdef _WIN32
 		device->Barrier(&GPUBarrier::Image(&rtReflection, IMAGE_LAYOUT_SHADER_RESOURCE, IMAGE_LAYOUT_RENDERTARGET), 1, cmd);
+#else
+		GPUBarrier image = GPUBarrier::Image(&rtReflection, IMAGE_LAYOUT_SHADER_RESOURCE, IMAGE_LAYOUT_RENDERTARGET);
```
nemerle, line 408: No idea if GPUBarrier has complex assignment operator but if it does then we can make this minimally more efficient by doing:
```cpp
GPUBarrier image(GPUBarrier::Image(&rtReflection, IMAGE_LAYOUT_SHADER_RESOURCE, IMAGE_LAYOUT_RENDERTARGET));
```

WickedEngine/wiArchive.h
```diff
@@ -5,6 +5,10 @@
 #include <string>
 #include <vector>
 
+#ifndef _WIN32
+#include <cstring>
+#endif
+
```
nemerle, line 11: We could remove the `#ifdef` here, even if the Windows string/vector already includes `<string>` beforehand?

WickedEngine/wiGraphics.h
```diff
@@ -479,22 +495,9 @@ namespace wiGraphics
...
-			} image;
-			struct Buffer
-			{
-				const GPUBuffer* buffer;
-				BUFFER_STATE state_before;
-				BUFFER_STATE state_after;
-			} buffer;
+			struct Memory memory;
+			struct Image image;
+			struct Buffer buffer;
```
nemerle, line 500: This is interesting, what errors were You getting on Linux trying to compile the old version?

WickedEngine/wiGraphicsDevice.cpp
```diff
@@ -3,8 +3,13 @@
 
 // These will let the driver select the dedicated GPU in favour of the integrated one:
 extern "C" {
+#ifdef _WIN32
 	_declspec(dllexport) DWORD NvOptimusEnablement = 0x00000001;
 	_declspec(dllexport) int AmdPowerXpressRequestHighPerformance = 1;
+#else
+	int NvOptimusEnablement = 0x1;
```
nemerle, line 10: we could maybe use a macro here?
```
#ifdef _WIN32
#define DLL_PUBLIC __declspec(dllexport)
#else
#define DLL_PUBLIC __attribute__ ((visibility ("default")))
#endif
```

WickedEngine/wiGraphics.h
nemerle: Ok, tried it in Godbolt, looks like it complaining about declaring types in anonymous union.
If the Memory/Image/Buffer types aren't used anywhere else, that union could be rewritten as:
```cpp
		union
		{
          	struct 
            {
                const GPUResource* resource;
            } memory;
            struct
            {
                const Texture * texture;
                IMAGE_LAYOUT layout_before;
                IMAGE_LAYOUT layout_after;
            } image;
            struct 
            {
                const GPUBuffer *buffer;
                BUFFER_STATE state_before;
                BUFFER_STATE state_after;
            } buffer;
		};
```

CMakeLists.txt
Je06jm: I'm not very familiar with CMake. Once I can get the pull request through, I'm hoping that someone who knows more about CMake can work on it

WickedEngine/RenderPath3D.cpp
Je06jm: That's definitely something to look into

WickedEngine/wiGraphics.h
Je06jm: Probably. Most of what I did was to get it to compile, so that we can start working on add the Linux layer to files like wiRawInput.cpp. There is definitely some syntax polish that can be done.

WickedEngine/wiGraphicsDevice.cpp
Je06jm: That's probably a good idea

WickedEngine/wiArchive.h
Je06jm: We'll have to try that after I get the initial pull request accepted.

CMakeLists.txt
Je06jm: Also a good idea

CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+cmake_minimum_required(VERSION 3.13)
+project(WickedEngine)
+
+set(CMAKE_CXX_STANDARD 17)
+set(VULKAN_SDK_DIR "/home/jeremy/VulkanSDK/1.2.135.0/x86_64")
```
turanszkij, line 5: That's probably not a good path. On Windows I set it from a system variable which is created by Vulkan SDK installer: $(VULKAN_SDK)/Include 
Is there something like that for Linux?

SConstruct
```diff
@@ -0,0 +1,9 @@
+VULKAN_SDK = "/home/jeremy/VulkanSDK/1.2.135.0/x86_64"
```
turanszkij, line 1: Same as above, is the path not retrievable automatically from SDK installation?

WickedEngine/MainComponent.cpp
```diff
@@ -55,17 +55,25 @@ void MainComponent::Initialize()
 		}
 		else if (wiStartupArguments::HasArgument("dx12"))
 		{
+#if _WIN32
```
turanszkij, line 58: I think it should be #ifdef

WickedEngine/MainComponent.cpp
```diff
@@ -55,17 +55,25 @@ void MainComponent::Initialize()
...
 			wiRenderer::SetDevice(std::make_shared<GraphicsDevice_DX12>(window, fullscreen, debugdevice));
+#else
+			assert(false);
+#endif
 		}
 
 		// default graphics device:
 		if (wiRenderer::GetDevice() == nullptr)
 		{
+#if _WIN32
```
turanszkij, line 72: #ifdef

WickedEngine/wiGraphics.h
```diff
@@ -469,6 +469,22 @@ namespace wiGraphics
 		PRIMITIVETOPOLOGY			pt = TRIANGLELIST;
 		uint32_t					sampleMask = 0xFFFFFFFF;
 	};
+	struct Memory
```
turanszkij, line 472: I'd prefer these structs declared within the GPUBarrier struct

WickedEngine/wiHelper.cpp
```diff
@@ -791,3 +790,116 @@ namespace wiHelper
 		}
 	}
 }
+
```
turanszkij, line 793: Most of these will not have to be reimplemented for Linux probaboly, so instead of new functions, just #ifdef for Linux specific parts within the existing functions where it's required.

WickedEngine/wiInput.cpp
```diff
@@ -581,3 +582,56 @@ namespace wiInput
 	}
 
 }
+
+#else
```
turanszkij, line 586: I'd prefer just extending the functions with #ifdefs for Linux where it's required, instead of reimplementing every function from scratch. (Like I suggested with wiHelper.cpp)

WickedEngine/wiLua.h
```diff
@@ -3,9 +3,15 @@
 
 extern "C"
 {
+#ifdef _WIN32
```
turanszkij, line 6: It's better to just use forward slash everywhere

WickedEngine/wiMath.h
```diff
@@ -17,66 +17,111 @@ namespace wiMath
 	}
 	inline float Distance(const XMVECTOR& v1, const XMVECTOR& v2)
 	{
+#ifdef _WIN32
```
turanszkij, line 20: Yeah, the reference doesn't really make sense here anyway, so your edit is better and should be the only one kept.

WickedEngine/wiMath.h
```diff
@@ -17,66 +17,111 @@ namespace wiMath
...
 		float Distance = 0.0f;
 		XMStoreFloat(&Distance, length);
 		return Distance;
 	}
 	inline float DistanceEstimated(const XMVECTOR& v1, const XMVECTOR& v2)
 	{
+#ifdef _WIN32
 		XMVECTOR& vectorSub = XMVectorSubtract(v1, v2);
 		XMVECTOR& length = XMVector3LengthEst(vectorSub);
+#else
```
turanszkij, line 51: Can we just be using the DirectX namespace globally as it is right now? It will be a pain in the ass to ifdef every single math usage per platform.

WickedEngine/wiNetwork_Linux.cpp
```diff
@@ -0,0 +1,32 @@
+#include    "wiNetwork.h"
+
+namespace wiNetwork
+{
+    void Initialize() {
```
turanszkij, line 5: Just calling out that I start the brackets on new lines for functions. If somewhere is not, it's a mistake.

WickedEngine/wiProfiler.cpp
```diff
@@ -178,7 +178,11 @@ namespace wiProfiler
 
 		lock.lock();
 
+#ifdef _WIN32
```
turanszkij, line 181: The non-reference path should be kept only

WickedEngine/wiRawInput.cpp
```diff
@@ -1,7 +1,21 @@
 #include "wiRawInput.h"
 #include "wiPlatform.h"
 
-#if defined(_WIN32) && !defined(PLATFORM_UWP)
+#ifdef _WIN32
```
turanszkij, line 4: For platforms where there is no Raw input API, we don't care so all these changes should be removed. Raw input is only Windows-desktop specific API as far as I am aware.

WickedEngine/wiResourceManager.cpp
```diff
@@ -69,7 +69,11 @@ namespace wiResourceManager
 
 		// dynamic type selection:
 		{
+#ifdef _WIN32
```
turanszkij, line 72: I think it makes sense to not use reference here for either platform

WickedEngine/wiVersion.cpp
```diff
@@ -7,7 +7,7 @@ namespace wiVersion
 	// main engine core
 	const int major = 0;
 	// minor features, major updates
-	const int minor = 42;
+	const int minor = 43;
```
turanszkij, line 10: That's right, these changes will be a new minor version, and revision need to set to 0 after minor version changes.

WickedEngine/wiXInput.cpp
```diff
@@ -112,9 +112,20 @@ namespace wiXInput
 #else
 namespace wiXInput
 {
-	void Update() {}
-	int GetMaxControllerCount() { return 0; }
-	bool GetControllerState(wiInput::ControllerState* state, int index) { return false; }
-	void SetControllerFeedback(const wiInput::ControllerFeedback& data, int index) {}
+	void Update() {
```
turanszkij, line 115: XInput will never be on Linux, so these empty functions don't make sense, just remove.

CMakeLists.txt
Je06jm: Done. On Linux, you now have to export VULKAN_SDK=".../path/to/vulkan_sdk"

SConstruct
Je06jm: Removed SConstruct

WickedEngine/MainComponent.cpp
Je06jm: You are 100% right

WickedEngine/wiGraphics.h
Je06jm: Done! Just for future reference, you can't declare a struct like `struct Memory { ... } memory;` in a union. Instead, use `struct { ... } memory;' or define them seperetly.

WickedEngine/wiLua.h
Je06jm: Indeed. I'm not sure why Windows even uses \\ when it supports /.

WickedEngine/wiMath.h
Je06jm: Yes. That was just to get my ide to stop complaining, so I could find actual problems easier.

WickedEngine/wiNetwork_Linux.cpp
Je06jm: Of course. We really need to get a stylesheet made up

WickedEngine/wiRawInput.cpp
Je06jm: Alright. I'll #ifdef it out of the Linux build

WickedEngine/wiGraphics.h
turanszkij: That's a little bummer, but no big deal. :(

turanszkij:
Closing this for not addressed issues remaining. I started over the linux build from scratch here: #125 
I might revisit the Cmake system later.

PR #133, opened 2020-07-02
Event system

turanszkij:
The event system will be used to intercept resolution change, DPI change, shader reload events from multiple systems. It is also intended for user defined events.

+ other interface changing refactors


turanszkij:
Dropped "renderpass callback" functionality for the time being.

PR #136, opened 2020-07-11
WIP: Initial linux support using SDL

portaloffreedom:
In this branch I'm trying to port the engine on linux through the use of the SDL2 library.
I've decided to not include the extra libraries needed to run the engine.

Up to this point, the SDL version of the engine is supposed to work on windows as well. Potentially on MacOS as well through the use of MoltenVK.

What works:
- [x] cmake build system
- [x] create a window with a vulkan surface
- [x] rendering demos (sometimes I get deadlocks)
- [x] mouse and keyboard input (needs refinement about DPIs and a check on reliability)
- [ ] gamepad input
- [ ] touch input
- [x] file picker
- [ ] audio
- [x] Editor
- [x] Demos
- [ ] Template projects

...will add more things todo as they come up


turanszkij:
Submitted PR to fix windows build

WickedEngine/wiRenderer.cpp
```diff
@@ -1308,7 +1308,7 @@ void LoadShaders()
 	wiJobSystem::Execute(ctx, [](wiJobArgs args) { LoadShader(PS, pixelShaders[PSTYPE_CAPTUREIMPOSTOR_ALBEDO], "captureImpostorPS_albedo.cso"); });
 	wiJobSystem::Execute(ctx, [](wiJobArgs args) { LoadShader(PS, pixelShaders[PSTYPE_CAPTUREIMPOSTOR_NORMAL], "captureImpostorPS_normal.cso"); });
 	wiJobSystem::Execute(ctx, [](wiJobArgs args) { LoadShader(PS, pixelShaders[PSTYPE_CAPTUREIMPOSTOR_SURFACE], "captureImpostorPS_surface.cso"); });
-	wiJobSystem::Execute(ctx, [](wiJobArgs args) { LoadShader(PS, pixelShaders[PSTYPE_CUBEMAP], "cubemapPS.cso"); });
```
turanszkij, line 1311: I noticed that you renamed some shaders, why?

trolley813: @turanszkij That's probably because the HLSL shader file in the repository is named `cubeMapPS.hlsl`. Unlike Windows, file names in Linux are (usually) case-sensitive.

turanszkij: Good catch, makes sense!

turanszkij:
Submitted pr to resolve conflicts

PR #141, opened 2020-07-13
Fix compiler warning: delete[] on void* is undefined behaviour

JohanAR:
There were so many things that differed between 16 and 32 bit indices so it felt more elegant to move them inside the if-else. But if you prefer the other style the warning can ofc. be fixed in another way


turanszkij:
This is very good

Issue #145, opened 2020-07-19
New resource binding model


turanszkij:
The new resource binding model aims to be more flexible and efficient. It will allow the developer to specify root signatures, root descriptors, root constants and descriptor tables. With this, comes the ability to update only a subset of descriptors dynamically, while the static descriptors are kept alive. It will also allow to have unbounded descriptor array and descriptor indexing within the shaders. It will allow to set secondary root signatures for ray tracing pipelines. The aim is to let DX12 and Vulkan devices work with this model, while optionally the DX11 device would emulate it under the hood with falling back to slot based binding model. The slot based model will not be maintained, so that the use of shader reflection can be removed.

turanszkij:
Added in 9723d94d906fe6dda527cc4a8e5b1c86ec54c374

PR #146, opened 2020-07-19
New resource binding model

turanszkij:
See issue #145


turanszkij:
Draft for new resource binding model:
- developer will declare root signature layout
- developer will allocate descriptor tables
- developer will declare root signatures for PSOs and compute shaders
- developer will write descriptors into tables with simple interface
- developer will bind root signatures
- developer will bind descriptor tables, root descriptors, root constants
- system will ensure that latest changes in descriptor tables are shader visible

TODO:
- implement for Vulkan device
- implement for DX11 device optionally if feasible
- modify all engine code to comply

#146

turanszkij:
- Developer will be responsible to manage CPU threading and descriptor table writes

turanszkij:
Problems:
- root constant / push descriptor syntax changes are maintainable between dx12 and vulkan with ConstantBuffer<T> syntax, but not feasible with dx11, which doesn't support ConstantBuffer<T> syntax
- shader visibility flags very differently set between dx12 and vulkan. Probably best to just use always all stages?
- dx12 descriptor spaces / vulkan descriptor set numbers mismatch: don't know yet how to handle nicely

amerkoleci:
Not sure if all this stuff simplifies developers life, maybe a fallback to easier mode will be welcome probably, for dx12 and vulkan experts all this stuff is easy, newcomes will suffer

turanszkij:
> Not sure if all this stuff simplifies developers life, maybe a fallback to easier mode will be welcome probably, for dx12 and vulkan experts all this stuff is easy, newcomes will suffer

Thanks for the feedback, that is definitely a concern. I will try to provide a simple interface here that is possibly simpler than the old model, however, it will require a change of mindset.

turanszkij:
Changes:
- removed root descriptors (vulkan has dynamic buffers instead, works very differently)
- WriteDescriptor() will handle arrayIndex for descriptor arrays
- WriteDescriptor() can take buffer offset
- static samplers now part of DescriptorTable
- DescriptorTable bind point = HLSL descriptor "space" (to keep consistent model between dx12 and vulkan)
- DescriptorTable can have both samplers and resources inside it (dx12 will handle the separation in the backend)
- Root constants: specify byte size (instead of uint32_t count) at declare time, not when binding (to comply with vulkan)
- User doesn't need to bind RootSignature, it is bound automatically with PSO/ComputeShader/RaytracingPipelineState

Problems:
- Root descriptor (dx12) is different than dynamic buffer (vulkan). Root descriptor is a lot more convenient to use, but need to figure out how to handle efficiently with vulkan. It could be just a vkCmdBindDescriptorSets() with the offsets, but how to detect if the table needs allocation?

turanszkij:
Changes:
- Root descriptors implemented: developer declares them within DescriptorTable (to match vulkan), but actually binding is following the more convenient dx12 model. 
- refactors

Problems:
- Indexing the root descriptors could be weird (index them just within themselves, not counting regular descriptors). Some more testing will reveal if this is a good behaviour or not.
- How to best ensure that very frequent small draws don't always start a new gpu visible heap entry (new descriptorset)? If this possible from backend side, that would be the best solution.

turanszkij:
Implemented in 9723d94d906fe6dda527cc4a8e5b1c86ec54c374 
The old binding model can still be used

PR #148, opened 2020-07-27
Volumetric Cloudscapes and Atmospheric Scattering

Kliaxe:
Also some minor SSR improvements, it should be better in motion now.


Kliaxe:
Alright, everything should finally be set up!
And of course if/when I find a better way to deal with things I'll make some improvements.

The atmosphere are still WIP. There are plenty of ways to improve it, for instance I have to make it integrate properly with the scene, so we can apply some aerial perspective, and not to mention performance stuff.
Protip: Try flying into space

Right now the volumetric cloud pass is just applied to the postprocessing chain to avoid messes. I found multiple ways to integrate it with the rest of the renderpaths, but I couldn't decide what the best pratical way was, so I reckoned you knew how apply it properly.

turanszkij:
Thanks for this, it's great! I opened a pull request into this branch with a few changes that I would like to make.

Plus some other topics:
1) Realistic sky:
- It is very heavy, I can't run just the sky rendering in 4k/60FPS on RTX 2060. It is maybe not a very good idea to have this heavy function in every object rendering shader (plus raytracing, etc, whatever uses sky). If it reduces performance of other shaders while not enabled, then it will have to remain as compile time option to switch to realistic sky.
- It is darker than I would expect with the base exposure setting and no eye adaption (maybe eye adaption is not very good implementation, because it overbrightens a lot)
- The horizon silhouette line is not anti aliased (or I don't know how to describe it)
- The light scattering is also visible on the other side of the sky when the sun is about to go down the horizon (I don't know if that is intentional or not)
- The sun ring is also visible when the sun goes below the horizon

2) Volumetric Clouds
- When going above clouds, there is a point when there is a sudden popping, and above that point, the shape of them is slightly different, and they don't extend to infinity, just around the camera
- I think the volumetric clouds are best kept as post process because they have an other feature which the simple clouds don't. The simple clouds are only running on part of the screen where nothing was rendered, but the volumetric clouds can also occlude rendered geometry (which looks very cool).

Let me know what you think of the above.

Kliaxe:
Those are some good thoughts!

1. 
- Oh wow I only now noticed. I think when I tested it I might have disabled some things. As I mentioned it's in its early stages and performance changes will be made. Should we exclude it for now or mark it experimental for awareness?
- Since it's based on the primary sun light intensity, by default it is a bit underexposed, but I thought it would be nice if the atmosphere brightness would change according to the rest of the scene, the same with the clouds since they're considered a 3D object. Should we perhabs change this?
- Yes I noticed that too. If we enable some AA it does hide it a lot. Right now the atmosphere renders an artificial planet, which is what you see below the horizon. This could perhabs change later on though, and maybe relax the line.
- I'm not entirely sure what you mean. Could you provide an example?
- Oops that shouldn't happen. I'll fix that.
2.
- The popping is something I did on purpose and we can definitely change this. Before there was a seamless transition, but I noticed if you were inbetween the clouds at that moment, you could see the quality difference between the top and bottom view, which i found a bit obnoxious (but that's just me). That was some time ago and things may be different now. I could add an option if you'd like, and see if I can improve this. The fade is a bit funny atm when above the clouds, but I'll also work on that too.
- That was also what I was thinking. The only thing that's a bummer, is that it can't be applied to the SSR or transparent objects, etc. without cheating a bit.

turanszkij:
1)
- For now, let's leave it as it is
- Ah of course, changing sun energy will affect sky brightness, that makes a lot of sense.
- I was thinking that a smoothstep could be applied to the planet sphere, like the sun ring, to make it smooth even without anti aliasing turned on. I will see if I can make this work.
- It's a bit hard to make a picture that shows it. It can be seen when sun is just above horizon, so it's pretty dark. If camera faces the sun, you can see the sun and increased brightness of the sky. When you start turning sideways, 90 degress is where the sky is darkest, but if you continue turning the camera to 180 degrees, the sky brightness starts increasing again slightly. It might be physically correct though.

turanszkij:
I was messing around trying to eliminate the hard line between planet and atmosphere, but failed. I want to disable the realistic sky at compile time for now because I'm not comfortable to include that heavy code in other shaders. This is in the PR. I am happy to merge it whenever you say so. :)

Kliaxe:
Alright should be merged now! It is probably best to keep it disabled for now. Incase you're curious, to optimize the atmosphere we'd precompute some assets into textures, which we just sample at runtime. This repository is a good example and great source of inspiration:
https://github.com/ebruneton/precomputed_atmospheric_scattering/

I think you can merge. I'm not near any pc atm, but I'd like to make some changes to the clouds and especially the atmosphere, but I think I'll do them in a separate PR once I get the time.

However, thanks once again for letting me contribute to the project!

turanszkij:
Thanks a lot for contributing!

PR #160, opened 2020-10-10
0.49

turanszkij:
This will contain various updates for 0.49 including:
- material shaders
- custom shaders
- gui


turanszkij:
dropping forward, tiledforward, deferred, tileddeferred render paths and instead have a "default" render path that supports everything. The path tracing will be kept

PR #161, opened 2020-10-12
Renderpath upgrade

turanszkij:


turanszkij:
Will be continued in #160

PR #165, opened 2020-10-28
Precomputed Atmospheric Scattering

Kliaxe:
And a small ssr change.


Kliaxe:
I made a change to the SSR temporal pass, because the previous way of dealing with min and max responses felt in the end quiet unpreditable at times. At first i thought it worked pretty well, but soon after I noticed some problems with the temporal pass and by changing it, I got better results and control. I don't know if we should change the other denoising stuff too or not.

turanszkij:
Hi, most of it is good, but I fixed a few things:
- Linux build didn't build because of incorrect file encoding of skyAtmosphere.hlsli, so changed it to UTF8
- raytraced reflections root signature was set separately and the atmosphere luts were not bound
- resized realistic sky button in the editor
- refactored the GetDynamicSky() function in shaders, internally it uses the global lut textures instead of expecting them as parameter

For some reason I couldn't create a PR into your repo, so you should merge my changes into this PR from this branch: https://github.com/turanszkij/WickedEngine/pull/167

Kliaxe:
Everything should be merged now

turanszkij:
I also noticed some shader warnings about uninitialized variables in the sky shaders. I will fix them later, they are probably bogus, since I already seen things like this that were incorrectly reported by fxc (and not reported by dxc).

PR #177, opened 2020-11-05
Improvements

amerkoleci:
Fix RESOURCE_MISC_RAY_TRACING not mapping correctly.
Vulkan: Correctly handle swapchain resize logic and semaphore handling.
Vulkan: Enable VK_EXT_debug_utils only if supported and enable debug utils messenger instead of deprecated VK_EXT_Debug_Report.


WickedEngine/Shaders_SPIRV.vcxproj
```diff
@@ -55,7 +55,7 @@
       <AdditionalOptions>-flegacy-macro-expansion -spirv -fvk-use-dx-layout %(AdditionalOptions)</AdditionalOptions>
     </FxCompile>
     <PreBuildEvent>
-      <Command>compile_shaders_spirv.py
+      <Command>python compile_shaders_spirv.py
```
turanszkij, line 58: This doesn't let this project compile on Appveyor, would you remove the "python" please?

turanszkij (changes requested):

WickedEngine/Shaders_SPIRV.vcxproj
amerkoleci: Reverted, sorry

turanszkij (approved):

PR #180, opened 2020-11-06
Add Morph Target

### LINKED ISSUE #179: Morph Target Support
Morph targets can be useful in character customization and facial animation. According to glTF 2.0 specification, the format support morph target as well but has not been implemented yet in Wicked Engine.
[https://github.com/KhronosGroup/glTF-Tutorials/blob/master/gltfTutorial/gltfTutorial_017_SimpleMorphTarget.md](url)

I have implemented this feature. There is no conflict with master branch and no code has been removed. However, the archive version will be 53 or above to prevent breaking assets. I also add the tool for adjust weight in editor. 

If you're interested, I will make the pull request so you can check the code.
image

XthemeCore:
The sample model is not include due to the license. If you want to try, you can download from [here](http://www.mediafire.com/file/z38ymsowdsj3yyd/MorpthTargetSample.zip/file).

Resolves #179

### DIFF
--- a/Editor/MeshWindow.cpp
+++ b/Editor/MeshWindow.cpp
@@ -513,6 +513,32 @@ void MeshWindow::Create(EditorComponent* editor)
 	AddWidget(&terrainGenButton);
 
 
+	morphTargetCombo.Create("Morph Target:");
+	morphTargetCombo.SetSize(XMFLOAT2(100, hei));
+	morphTargetCombo.SetPos(XMFLOAT2(x + 280, y += step));
+	morphTargetCombo.OnSelect([&](wiEventArgs args) {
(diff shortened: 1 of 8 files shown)


turanszkij:
I tried it, it's pretty good and will be a useful feature! It is not recommended to always call MeshComponent::CreateRenderData() for performance reasons, this will recreate every GPU buffer of the mesh. Instead, only the required buffers will need to be updated via GraphicsDevice::UpdateBuffer() call. It's not that urgent though, I can do this optimization later if you'd like.

An other thing I noticed that the normals are not updating with the morph, is that expected?

turanszkij (approved):

turanszkij:
Actually, dynamic update will be important when we have morph target animation implemented. It will make more sense to implement at that point when we know more requirements. I will merge this now.

PR #183, opened 2020-11-07
Fix - add normal for morph target

### LINKED ISSUE #179: Morph Target Support

XthemeCore:
I don't want to increase archive version so here is a quick fix for morph target. Hopefully no one has created the wiscene asset using last committed version.

In blender export, glTF file contains both position and normal (without tangent) which I think might be used in updating the normal for morph target. There is no implementation for normal yet. Just prepare the file to prevent breaking the assets.

Fix #179

### DIFF
--- a/Editor/ModelImporter_GLTF.cpp
+++ b/Editor/ModelImporter_GLTF.cpp
@@ -812,6 +812,15 @@ void ImportModel_GLTF(const std::string& fileName, Scene& scene)
 								mesh.targets[i].vertex_positions[vertexOffset + j] = ((XMFLOAT3*)data)[j];
 							}
 						}
+						else if (!attr_name.compare("NORMAL"))
+						{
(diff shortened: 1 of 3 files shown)


turanszkij (approved):
Usually it's not a good way to not increase archive version in this case. But since this fix comes quickly, I think it's ok. I wouldn't want to add the version check in the middle of morph serialization. Thanks.

PR #185, opened 2020-11-07
Font outline

CaminoOnPc:
Added simple font outline code, in theory, can be drawn more efficiently but performance impact from this method is basically negligible so shouldn't matter for now

### DIFF
--- a/Editor/MeshWindow.cpp
+++ b/Editor/MeshWindow.cpp
@@ -535,7 +535,7 @@ void MeshWindow::Create(EditorComponent* editor)
 	    if (mesh != nullptr && morphTargetCombo.GetSelected() < mesh->targets.size())
 	    {
 			mesh->targets[morphTargetCombo.GetSelected()].weight = args.fValue;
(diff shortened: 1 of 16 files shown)


turanszkij (commented):
I don't really like this solution. It produces many draw calls, but more importantly it doesn't look good (in my opinion), see attached image.


Quality can be improved by adding more draw calls that expand the outline diagonally too.

But I'd rather not add outline support at this point.

BluePrintRandom:
in blender they have a system to extrude / bevel text

you can use this to produce normals on the bevel - and then use

dot(normal, local_up) to pull out 2 tones
https://www.youtube.com/watch?v=MFcUVU_JPV0
that is how these icons are made

turanszkij:
It is a texture based font being used here, so I don't see how that would work.

BluePrintRandom:
ah - blender tesselates a svg or font to create a model

this may be a neat shader to have (3d text from 2d font etc)

PR #190, opened 2020-11-14
Automatic shader dependencies in cmake targets

portaloffreedom:
I got the shader compilation in cmake improved by automatically appending the included headers (recursively) into its dependencies.
In short, if you change an header it will also recompile all shaders that use that header.

Downsides:
- It's extremely slow (because I have to compile a shader to know its header dependencies. It's a dxc limitation but that we can get the developers to improve it)
- including/removing headers will need to re-run the cmake generation to make the building process consistent again (reload cmake in clion / run cmake in the terminal)


turanszkij (approved):

turanszkij:
Awesome, thanks!

Issue #197, opened 2020-11-30
Subsurface scattering rewrite


turanszkij:
Instead of post process SSS, investigate the wrap lighting method. 

Pro:
- Better fit for forward renderer, don't need to write out light buffer for blurring
- More control over subsurface amount and color

Con:
- Shadow edges won't be blurred (could be solved by soft shadow like approach, but is it important?)

Additionally implement back-scattering by sampling shadow maps.

PR #198, opened 2020-12-01
Subsurface Scattering Rewrite

turanszkij:
Based on #197


turanszkij:
- removed lightbuffers from gbuffer
- removed deferred composition
- removed sss post process

- added subsurface scattering: amount, color parameters. The technique is based on wrap diffuse lighting in forward shader

Problems:
- shadow cull mode switched because otherwise SSS material has very ugly self shadow. But this shadow cull mode is not as good as the normal one, results in some disconnects. Need to enable this only for SSS materials.

turanszkij:
-shadow culling is not nicely fixable. For now, SSS materials that have self shadow problem should disable "cast shadow"

PR #199, opened 2020-12-04
disable jobsystem in editor loading screen

turanszkij:
There is an infinite loading screen issue with a computer with only 2 cores, so try disabling editor job system.


turanszkij:
Fixed in 0292d190d66f5cd517ad30058ff7dd42602dbed1

PR #201, opened 2020-12-07
Update D3D12MA.

jdswebb:
D3D12MA now requires the adapter be passed when initializing. The highest performance adapter is chosen explicitly before device creation and provided as an argument to D3D12CreateDevice. DXGIFactory::EnumAdapterByGpuPreference requires the latest DXGI header so I have switched to it - if that's not ok will need to do it the older way.


turanszkij:
Could you please remove the D3D12MemAlloc.cpp line from WickedEngine\WickedEngine\Utility\CmakeLists.txt? That is incorrectly included for the linux build

turanszkij (approved):

turanszkij:
Thanks for contributing!

PR #203, opened 2020-12-18
SSR Changes

Kliaxe:
Hi turanszkij!

A bit unexpected PR here. The SSR in the current state was unstable and unpolished, and felt it needed some attention. Great changes has been made and introduces some new features. The implementation should now be alot closer to PBR. It now features importance sampling of visible normals, and blue noise. 
To take full advantage of the new imporance sampling, we had to use a bigger range of rays, but that introduced more noise. That's when I realised the current noise wasn't good enough, and that's where blue noise comes in.

Bug fixes, refactors, and performance changes has also been made.

Here is a preview:
sc_18-12-2020 23-46-28


turanszkij (commented):
This sounds great! You could also bump the revision version number in version.cpp by 1

Kliaxe:
Oops will do! 

I also wish I could experiment with these things with raytracing, but unfortunately I don't own a raytracing capable GPU... yet. Waiting to get my hands on one of those new GPUs.

turanszkij:
Thank you for the update, it looks great! Can't wait until you start improving the ray traced reflections. ;)
Until then, I will see if any of your changes can be integrated to the rt reflections too.

turanszkij:
I've integrated the blue noise ray directions into raytraced reflections, it looks much better now. Plus some refactors to share more code between SSR and rt reflections. Also, the Structured Buffer is better to use than a Typed buffer usually, so changed that for the blue noise stuff.

Kliaxe:
Oh that looks really good! Even though we don't have a "high-end" denoiser, it still looks great!
Just curious, what roughness did the floor have in that picture?

turanszkij:
The roughness was set to 0.276 on the slider. Also, here is a comparison between SSR and ray traced reflections if you are curious:

PR #210, opened 2020-12-30
D3D12 - fix swapchain resize and rework WaitForGPU() to wait for copy and direct queues to actually complete.

jdswebb:
I moved the common fence stuff to a struct - looks a little hacky for the frame fence so perhaps that could be improved a bit.

Swapchain resize works, shutdown is now clean and everything seems to be ok.


turanszkij:
I don't really like that there is an abstraction for the fence. I think using plain fences is fine, it will make it easier to refactor usage of a single fence if needed.

jdswebb:
Ok, will close this one then.

turanszkij:
Sorry about that. I don't think fences always need locks and event handles in some cases either, only in specific cases.

jdswebb:
No worries, I did have the same reservations about the unneeded locks/event handles. It was just having the same duplicated code everywhere for (for example) waiting didn't seem very nice either.

PR #211, opened 2020-12-30
D3D12: Fix OBJECT_DELETED_WHILE_STILL_IN_USE errors upon swapchain resize and shutdown

jdswebb:
GraphicsDevice_DX12::WaitForGPU() now waits for the copy and direct queues to complete rather than waiting for the frame fence which was not sufficient.

This fixes OBJECT_DELETED_WHILE_STILL_IN_USE errors upon swapchain resize and shutdown.


jdswebb:
Attempt number 2! 😃

turanszkij:
Looks good now. I think there is a better way to resize swapchain without blocking the GPU (like Vulkan), but this will do now for safety.

PR #214, opened 2021-01-07
fixed SDL2 discoverability issues under Manjaro

mzjaworski:


mzjaworski:
@portaloffreedom I could not add you as reviewer, so I decided to notify you like that.

mzjaworski:
I have just noticed a small issue with my submitted CMake, I will quickly fix it

turanszkij:
Thanks for submitting the PR. Sorry that I auto merged into your branch, I didn't think it was actually possible but tried pressing that button on Github anyway.
@portaloffreedom Does this fix also solve the SDL problem for your distro? Do you agree with this change?

portaloffreedom:
It's definetely not complete. When using pkg_config instead of cmake just including the target is not enough. We need to include also the Include directories, compiler definitions and linking directories for that library.

But out of curiousity, why pkg_config? Do you not have in your system the cmake files for SDL2 at all?
(for me)
```
sdl2 /usr/lib/cmake/SDL2/SDL2Config.cmake
sdl2 /usr/lib/cmake/SDL2/SDL2ConfigVersion.cmake
sdl2 /usr/lib/cmake/SDL2/SDL2Targets-noconfig.cmake
sdl2 /usr/lib/cmake/SDL2/SDL2Targets.cmake
```

mzjaworski:
@portaloffreedom I do have them in the same directory 
image

Problem simply is that when i am using find_package it does not link against SDL2. and when I have changed it to pkg_config it did.

portaloffreedom (approved):
it works well also on my computer and it seems to pass the CI as well, for the its good :)

turanszkij:
Thanks for the updates!

PR #221, opened 2021-01-17
Velocity buffer should be written by prepass

turanszkij:
This would work better with ray traced reflection, ssr, and other things that need denoising/refprojection. Having the ability to reproject earlier gives several benefits, like staring the effects earlier.


turanszkij:
todo: fix linux, voxelizer, tessellation, and test a lot

PR #224, opened 2021-01-23
Volumetric Clouds Updates

Kliaxe:
Hi turanszkij!

Great changes to the volumetric clouds system has been made. It should now be a lot closer to PBR.
Bug fixes, refactors, and performance changes has also been made.

I'm curious what performance you're getting with the clouds on. My laptop (gtx 1050) with fullscreen enabled is getting around 10 ms at worst. Even though the clouds are well optimized, it's in a very high graphics preset. Why? Because it looks gorgeous! However by tweaking some performance values I was able to get 4-5 ms and still looks fantastic.

Here is a preview:
sc_23-01-2021 23-22-30

### DIFF
--- a/WickedEngine/RenderPath3D.cpp
+++ b/WickedEngine/RenderPath3D.cpp
@@ -1293,12 +1293,9 @@ void RenderPath3D::RenderPostprocessChain(CommandList cmd) const
 	{
 		if (getVolumetricCloudsEnabled())
 		{
-			const Texture* lightShaftTemp = nullptr;
-
 			wiRenderer::Postprocess_VolumetricClouds(
 				*rt_read,
 				*rt_write,
-				*lightShaftTemp,
 				rtLinearDepth,
 				depthBuffer_Copy,
 				cmd
--- a/WickedEngine/volumetricCloud_finalCS.hlsl
+++ b/WickedEngine/volumetricCloud_finalCS.hlsl
@@ -3,7 +3,6 @@
 
 TEXTURE2D(texture_input, float4, TEXSLOT_ONDEMAND0);
 TEXTURE2D(texture_reproject, float4, TEXSLOT_ONDEMAND1);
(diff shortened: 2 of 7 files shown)


turanszkij:
Thanks, this is really good! The performance is a bit slower, in previous version, from default camera it took ~2ms to render clouds in 2848x1429 resolution. The new ones take ~3ms with same setup. But the new ones do look nicer! Just a minor observation, is that you always compute velocities from camera matrices, but the velocity buffer already should contain proper velocities even for the sky. It's not important though and I can update this later too.

turanszkij (approved):

Kliaxe:
For some reason I didn't think of that. I just assumed by using the cloud position it would be more accurate. If it doesn't make a real difference I think we can safely remove it and use velocity buffer instead.

Btw would it be possible for you to add a 2x downsampled depth buffer?
https://github.com/turanszkij/WickedEngine/blob/fc96b2a3b3593f73f1484dda0954588b8cbdb7bb/WickedEngine/volumetricCloud_renderCS.hlsl#L668 
At this line I'm using the linear depth buffer instead, because it wasn't available at the time. I wasn't sure what the best way to implement it would be.

turanszkij:
Yes, I think it would probably be a good idea to add a downsampled native depth buffer too. Will need to watch out because this takes non trivial amount of time, so need to investigate exporting both at the same time if possible.

turanszkij:
I added two extra levels for `texture_depth`, so mip1 now contains 2x downsampled depth and mip2 contains 4x dowsampled depth. The depth was downsampled with MAX operation (so unlike linear depth it takes closest to camera depth sample of neighbors). Is that good for your case?

Kliaxe:
That looks good! I'll probably apply the changes next time I make another update. My next goal is to add some environment luminance with the realistic sky, to apply to the clouds and fog.

PR #226, opened 2021-01-30
sheen and clearcoat

turanszkij:


turanszkij:
TODO: 
- clear coat normal somehow messed up
- validate sheen and clearcoat against gltf reference

PR #229, opened 2021-02-04
sky and cloud updates

Kliaxe:
Hi turanszkij!

A small update has been made to the sky and clouds. Before, the fog was using the uppermost pixel as it's source of color, which wasn't accurate with the realistic sky. The same with the clouds, as it just used a custom color as it's ambient light. Now the engine is equipped with a texture that captures the surrounding sky color, when realistic sky is enabled. That now fixes the current problems.
Some changes has also been made to the clouds so it performes a bit faster, and distributed the sample count differently as it looks nicer. Should also now use the downsampled depthbuffer as discussed last time.

Also, moving the resource indexes by 1 in Resourcemapping.h, would that change anything? Just want to be 100% sure. 

Btw, the sheen and clear coat looks REALLY good!


Kliaxe:
Should be merged now!
I noticed the clouds doesn't work well with reflections, refractions or light shafts. That's simply because it's a postprocessing effect and is scheduled to draw after the scene. Would it be possible to add the clouds earlier in the pipeline? One idea I had, was to add an entry to DRAWSCENE_FLAGS with volumetrics, and make it run once per frame like now, and once per env probe. For the probes we could make a very cheap version of the clouds. 
These are just speculations, but I'm curious to hear what you think of this.

turanszkij:
Yes, it's possible to move the clouds earlier. The SSAO/SSR is also being run pretty early now, I think the volumetric clouds should be somewhere around there to be correctly refracted at least. The reflections sound more complicated. I don't see what you meant by adding it to DRAWSCENE_FLAGS, so you can show it if you want.

Kliaxe:
By rendering the clouds as early as SSR/SSAO doesn't sound bad and should be straight forward to implement. That should be enough for light shafts (additional occlusion texture) and transparents. But if we want to take it a step further, we could also add the clouds to the SSR, planar reflections and probes. That's where the idea came with the DRAWSCENE_FLAGS. I actually messed up, because I meant RENDERPASS, where we then could add the clouds into the drawscene function. In my head, I thought it could work out if we made the clouds into the rtSceneCopy texture, which the SSR and perhaps other effects uses. All together I'm not sure though.

turanszkij:
Yeah if the clouds would be rendered as early as SSR, they would be in the rtSceneCopy texture, so if the SSR can sample as far as the sky, then it would work.
Implementing it into the envmap sounds difficult or not straight forward.
Rendering it for planar reflections would work probably by just rendering it again with the planar reflection camera, but no additional changes.
I don't think it needs to be a RENDERPASS or part of the DrawScene, because that's a place for object rendering, not full screen effects.

Kliaxe:
Interesting stuff! Now that I think about it, you're probably right about the probes. It would be difficult to add, since the clouds has lots of other passes like, noises, upsampling and temporal, and not to mention it is using compute shaders to render as a fullscreen effect, etc.
Makes sense about the DrawScene for object rendering. You mentioned if we added the clouds as early as SSR, we could add the cloud texture into the rtSceneCopy texture. However, I can't really get the image in my head of how we could/would do it atm.
Would you try moving the clouds earlier in the process?

Sorry for the delay btw.

turanszkij:
Yes, I'll try moving the clouds immediately after the main color pass, I think that way it will be automatically part of the rtSceneCopy.

turanszkij:
I moved the clouds before the scene mipchain pass. It works correctly now with refraction and SSR.

Kliaxe:
That looks really good! I thought it would be more difficult than that. Oh well

PR #230, opened 2021-02-04
barriers fixing

turanszkij:


turanszkij:
GENERAL layout is not allowed, as there is no dx12 equivalent. Most optimal layouts will be encouraged.

PR #232, opened 2021-02-07
SDL Keyboard Input, Unix File Loading and SDL Window Config

megumumpkin:
Added keycode mapping from the engine's keycode enum to the appropriate SDL2 keycode enum mapping.
Unix filesystem support by changing the win32 path branching backslashes to the appropriate unix file branching slashes.
Editor with window configuration support for SDL version.


WickedEngine/wiInput.cpp
```diff
@@ -548,6 +548,115 @@ namespace wiInput
                 case wiInput::KEYBOARD_BUTTON_PAGEUP:
                     keycode = SDL_SCANCODE_PAGEUP;
                     break;
+				case 48:
```
portaloffreedom, line 551: just one small thing, could you include in a comment a link where this codes come from? for future refence

WickedEngine/wiHelper.cpp
```diff
@@ -382,7 +382,13 @@ namespace wiHelper
 	bool FileRead(const std::string& fileName, std::vector<uint8_t>& data)
 	{
 #ifndef PLATFORM_UWP
+#ifdef SDL_FILESYSTEM_UNIX
+		std::string unixPath = fileName;
+		std::replace(unixPath.begin(), unixPath.end(), '\\', '/');
+		ifstream file(unixPath, ios::binary | ios::ate);
+#else
```
portaloffreedom, line 389: do we really need the else part? we can just do the replace in the `#ifdef SDL_FILESYSTEM_UNIX` section and load the file anyway

portaloffreedom (changes requested):

WickedEngine/wiInput.cpp
megumumpkin: Done!

portaloffreedom (dismissed):

megumumpkin (commented):
wiInput and wiHelper code has been revised

WickedEngine/wiHelper.cpp
```diff
@@ -382,7 +382,11 @@ namespace wiHelper
 	bool FileRead(const std::string& fileName, std::vector<uint8_t>& data)
 	{
 #ifndef PLATFORM_UWP
-		ifstream file(fileName, ios::binary | ios::ate);
+		std::string filepath = fileName;
```
turanszkij, line 385: There is no need to make a copy of fileName for the windows version, which doesn't modify it. So I preferred your very first solution actually.

turanszkij:
Thanks everyone for the fixes!

PR #233, opened 2021-02-08
Supports for building using CMake on Windows.

amerkoleci:
- Cleanup CMake scripts to better support Windows and Linux build
- Shader compilation using dxc.exe for shader model 6.0 or higher (no dx11 shaders ATM)
- Correct build of dependencies
- Correct handle of Win32 entry point and rc compilation


turanszkij (approved):

turanszkij:
I've tried it, works for me. Thanks!

PR #237, opened 2021-02-10
CMake improvements

amerkoleci:
Add legacy shader compilation (DX11) and improve CMake logic.

### DIFF
--- a/WickedEngine/wiRenderer.cpp
+++ b/WickedEngine/wiRenderer.cpp
@@ -55,11 +55,7 @@ GPUBuffer			resourceBuffers[RBTYPE_COUNT];
 Sampler				samplers[SSLOT_COUNT];
 
 // Specify shader folder using macro when running from CMake
-#ifdef WICKED_DEFAULT_SHADERS_DIRECTORY
-string SHADERPATH = WICKED_DEFAULT_SHADERS_DIRECTORY;
-#else
 string SHADERPATH = wiHelper::GetOriginalWorkingDirectory() + "../WickedEngine/shaders/";
-#endif
 
 LinearAllocator renderFrameAllocators[COMMANDLIST_COUNT]; // can be used by graphics threads
 inline LinearAllocator& GetRenderFrameAllocator(CommandList cmd)
--- a/WickedEngine/wiRenderer.cpp
+++ b/WickedEngine/wiRenderer.cpp
@@ -54,7 +54,6 @@ GPUBuffer			constantBuffers[CBTYPE_COUNT];
 GPUBuffer			resourceBuffers[RBTYPE_COUNT];
 Sampler				samplers[SSLOT_COUNT];
 
-// Specify shader folder using macro when running from CMake
 string SHADERPATH = wiHelper::GetOriginalWorkingDirectory() + "../WickedEngine/shaders/";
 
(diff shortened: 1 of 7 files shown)


turanszkij:
I still have problems, repro steps:
- download your repo as zip, unzip to desktop
- open the folder in visual studio, cmake is automatically generated
- when finished, I select Editor_Windows as startup project and press run (F5)
- after everything compiles fine, it crashes in PSO creation, because shaders are missing.

The console output contains messages like:
`\BUILD\x64\Debug\Editor_Windows/../WickedEngine/shaders/tonemapCS.cso`

This suggests to me that working directory is not correctly set to `Editor` directory

amerkoleci:
Can you test now after last commit please?

amerkoleci:
Now the shaders should be OK when opening in Visual Studio, still the working directory is not set correctly to folder.

WickedEngine/wiRenderer.cpp
```diff
@@ -54,7 +54,12 @@ GPUBuffer			constantBuffers[CBTYPE_COUNT];
 GPUBuffer			resourceBuffers[RBTYPE_COUNT];
 Sampler				samplers[SSLOT_COUNT];
 
+// Specify shader folder using macro when running from CMake
+#ifdef WICKED_DEFAULT_SHADERS_DIRECTORY
+string SHADERPATH = WICKED_DEFAULT_SHADERS_DIRECTORY;
```
turanszkij, line 59: This is not good, this would lead to a lot of confusion. You shouldn't modify this. 
Apart from that, it doesn't fix the issue anyway. If it would fix the shaderpath issue, it would still not fix the working directory issue.

turanszkij (changes requested):

amerkoleci:
Finally, the battle was won :) Now it works correctly, I think now wicked supports building on windows from Visual Studio Code as well.

turanszkij:
No change for me with the previously described workflow, do I need to do anything else on my side?

WickedEngine/wiRenderer.cpp
```diff
@@ -54,6 +54,7 @@ GPUBuffer			constantBuffers[CBTYPE_COUNT];
 GPUBuffer			resourceBuffers[RBTYPE_COUNT];
 Sampler				samplers[SSLOT_COUNT];
 
+// Specify shader folder using macro when running from CMake
```
turanszkij, line 57: What does this mean?

amerkoleci:
I'm doing:
- Open folder from Visual Studio, selection WickedEngine root folder
- Chosing x64-Debug or x64-Release from Configuration, then chosing Tests.exe or Editor.exe from target to run.
- Make sure to perform fresh checkout maybe?

Thanks

WickedEngine/wiRenderer.cpp
amerkoleci: Rest from my previous commit, removed.

turanszkij:
I always download your whole branch as zip and build from scratch, the content is never found that way. We can merge this PR though if you think it's done and improves things on your end. I appreciate you looking into this.

amerkoleci:
For me is ok to merge, would love to make it work for you as well

korfriend:
2021년 2월 15일 (월) 오전 5:51, Turánszki János <notifications@github.com>님이 작성:

PR #244, opened 2021-03-02
Bindless Descriptors

turanszkij:
Implementation for bindless descriptors. This will replace descriptor management with RootSignature/DescriptorTable. The user can query an `int` descriptor index from any resource that was created with `GetDescriptorIndex()`, for any subresource. The index will be valid for as long as the resource is valid. This largely simplifies descriptor management and improves performance. The old slot based binding model will still work, that is useful for binding a small amount of descriptors and not needing to rewrite existing shaders.

TODO:
- remove descriptor management API: DescriptorTable, RootSignature
- create simple API for push/root constants
- rewrite heavy resource usage shaders with bindless
- rewrite raytracing shaders with bindless


turanszkij:
Added simple push constants API:
C++:
```cpp
void PushConstants(const void* data, uint32_t size, CommandList cmd);
// ...
struct PushConstants
{
	uint materialIndex;
	int textureindex;
};
PushConstants push;
push.materialIndex = device->GetDescriptorIndex(materialCB, CBV);
push.textureindex = device->GetDescriptorIndex(texture, SRV);
device->PushConstants(&push, sizeof(push), cmd);
```
HLSL:
```
Texture2D<float4> bindless_textures[] : register(t0, space5);
struct PushConstants
{
	uint materialIndex;
	int textureindex;
};
PUSHCONSTANT(push, PushConstants);
// ...
float4 color = bindless_textures[push.textureindex].Sample(sam, uv);
```
- Only 1 push constant block allowed per pipeline for ease of use.
- Push constant will be bound at the time of next draw/dispatch command for the invoking pipeline.
- Push constant is limited to 128 byte data.
- Push constant is minimal requirement to use bindless API if not using the slot-based descriptor binder, to provide descriptor index from CPU to shader.

turanszkij:
Remaining todo:
- better subset data uploading, maybe rewrite shader side scene interface
- vulkan hangs sometimes
- move more things to bindless

PR #245, opened 2021-03-04
Fixed incorrect mouse delta position with SDL2

AlexAUT:
Mouse delta position was incorrectly calculated in the current version. To reproduce just open the Editor and try to adjust the camera, it will just spin randomly.
This pull request fixes the issue and makes the Editor usable under Linux.

### DIFF
--- a/WickedEngine/wiSDLInput.cpp
+++ b/WickedEngine/wiSDLInput.cpp
@@ -37,8 +37,8 @@ namespace wiSDLInput
 //        {
 //            internal_controller.state = wiInput::ControllerState();
 //        }
-        mouse.position.x = saved_x;
-        mouse.position.y = saved_y;
+        // mouse.position.x = saved_x;
+        // mouse.position.y = saved_y;
 
         std::vector<SDL_Event> events(1000);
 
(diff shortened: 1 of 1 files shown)


turanszkij:
@portaloffreedom Hi Matteo, could you take a look at this when you have the chance?

WickedEngine/wiSDLInput.cpp
```diff
@@ -80,18 +80,6 @@ namespace wiSDLInput
...
-                        mouse.delta_position.x += event.motion.xrel;
-                        mouse.delta_position.y += event.motion.yrel;
-                        // TODO you can update mouse buttons by reading event.motion.state
-                        mouse.left_button_press = event.motion.state & SDL_BUTTON_LMASK;
-                        mouse.right_button_press = event.motion.state & SDL_BUTTON_RMASK;
-                        mouse.middle_button_press = event.motion.state & SDL_BUTTON_MMASK;
-                        //mouse.x1_button_press = event.motion.state & SDL_BUTTON_X1MASK;
-                        //mouse.x2_button_press = event.motion.state & SDL_BUTTON_X2MASK;
-
-                        break;
```
portaloffreedom, line 94: I think we should not remove this `break;`

portaloffreedom:
@turanszkij This week I'm extremely busy with work, so I don't think I will be able to test it

WickedEngine/wiSDLInput.cpp
AlexAUT: Just wanted to be consistent with the source file. Because all other events are "grouped" and have one break per group.

But I can insert it, doesn't change the behaviour anyways.

turanszkij:
Thanks for making this fix and sorry for taking so long to merge. An other user also verified that it solves the mouse delta issue.

AlexAUT:
No worries, I'm happy to give something back to the project :).

PR #248, opened 2021-03-12
CMake speed-up: cache SPIRV output to prevent rebuild each config, make test and editor build optional

efedo:
Reduces the time required for subsequent CMake configurations from approx. 10-20 seconds to <1 second. Caches the last modified time for each shader source and only rebuilds if source is updated or SPIRV output is missing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,5 +22,13 @@ elseif(UNIX)
 endif()
 
 add_subdirectory(WickedEngine)
-add_subdirectory(Tests)
-add_subdirectory(Editor)
+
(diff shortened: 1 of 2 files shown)


portaloffreedom:
Cool!
I really like this improvement!

efedo:
It doesn't check whether includes are updated, but it appears that neither do the CMake DX11/12 pipelines. I can change it so that it does check for modified includes if you'd prefer.

turanszkij:
I've tested the linux build and everything works as expected. Modifying a shader include will correctly trigger the rebuild for the affected files. I will merge what's currently here. Thanks for contributing!

PR #250, opened 2021-03-15
Unified Shader Compiler

turanszkij:
Shader compile system is scattered across cmake, visual studio, and python scripts, all having some serious drawbacks. The unified shader compiler aims to provide every shader compilation need under one interface:
- compile shaders straight from c++ code with simple interface
- efficiently detect what needs to be recompiled (include dependencies...)
- efficient use of threads, job system is possible
- interop between engine code and shader compiler: controlling permutations will be easier
- can implement offline compilation or runtime compilation, while having context of the engine codebase
- background shader rebuild detection will be possible


turanszkij:
Current state:
- Windows can compile missing shaders on demand (vulkan, dx12)

Todo:
- detect include file changes
- implement background worker [editor] that recompiles when detecting outdated shader, then reloads shaders
- maybe hlsl5 for dx11, but that requires different compiler backend (not urgent, can wait)
- linux support would be nice (what about the comptr and atlbase.h dependency?)

turanszkij:
-TODO: improve wiBackLog, as mass shader compile message spamming reduces usability in the current form.

turanszkij:
Added shader dump feature which will generate wiShaderDump.h file that contains every shader as embedded binary data formatted as a C++ header file.

PR #251, opened 2021-03-17
Vulkan fix - use the destroy lock in DescriptorBinder::destroy() 

jdswebb:
Uuse the destroy lock in DescriptorBinder::destroy() as it can be called from multiple threads via pool growth in DescriptorBinder::validate().


turanszkij:
Thanks a lot!

PR #254, opened 2021-03-27
ssr and volumetric clouds improvements

Kliaxe:
Hi turanszkij!

I noticed a few minor issues that have been fixed in this PR.
Fixed a bug that was causing incorrect lighting and performance issues for the clouds.
I also noticed a bug where the cloud motion would become unstable, the further you went from origin, which also caused extreme 
lines at the edges when moving. Therefore, I had to redo some calculations.
In addition to this, a few tweaks and corrections has also been made.

Keep up the good work!


turanszkij (approved):

turanszkij:
Thanks a lot, always looking forward to your updates!

Issue #256, opened 2021-04-14
Multi swapchain support


turanszkij:
The device and swapchain will be logically separated. This will allow creating multiple swapchains and drawing to multiple windows.

turanszkij:
Problem:
- There is extensive usage of GetScreenWidth() and such functions that assume a single global swapchain/window, this would need lot of refactor.

turanszkij:
merged https://github.com/turanszkij/WickedEngine/commit/f1ced24f05ec65cfee25c189c588440d2bbfb4bf

PR #257, opened 2021-04-14
Multi swapchain support

turanszkij:
Testing #256 

Main changes:
- SwapChain is separated from device creation, CreateSwapChain allows creating multiple swapchains
- PresentBegin(), PresentEnd() removed. Instead, RenderpassBegin will have a version that accepts a SwapChain.

Problems:
- The device was written with the assumption of a single swapchain, so any uses of GetResolution, SetResolution, etc will have to be made on a SwapChain now. This will be a very large change. 
- Vulkan implementation is very messy right now because there are many places where surface - physical device - swapchain support is queried together (for example creation of present queue)


turanszkij:
Mostly everything in place at the lower levels. It will take more time to remove global resolution queries from higher level systems.

Issue #258, opened 2021-04-22
FidelityFX-Denoiser


turanszkij:
Integrate https://github.com/GPUOpen-Effects/FidelityFX-Denoiser to replace the custom denoisers for raytraced shadows and reflections

turanszkij:
Video presentation: https://youtu.be/RTtVC-Zk5yI

Problem is their data layout is different and only support 1 ray per pixel

turanszkij:
The FidelityFX reflection denoiser results were not good enough, current one is smoother and faster #265

PR #263, opened 2021-05-15
Raytraced shadow updates

turanszkij:
refactor raytraced shadows, and add better denoiser


turanszkij:
shadows and rtAO denoise done #258

Issue #261, opened 2021-05-05
Need support in HairParticle


cailuming:
![image](https://user-images.githubusercontent.com/19427259/117100583-ac30fc80-ada6-11eb-86ab-92354f0e798e.png)
Hi, I don't know how to get the  index of specified Entity, For example ,when I want to create a HairParticle, I must create a hair entity and get the wiHairParticle component from the hair entity. And then I have to specify a base mesh entity to the meshID of HairParticle. As you can see from the picture above, I want to load something like a plane, I use the wiScene::LoadModel, But here comes the problem, I have no idea about how to get the entity I just load. And I found that the return value of wiScene::LoadModel is type of entity, but it may be invalid if we pass false args.

turanszkij:
Hi, you can use `scene.Entity_FindByName("meshName")`. This will return an entity if one exists with that name. If multiple entities exist with the same name, the first one will be returned (so I recommend to have unique entity names). 

You can also set up hair particle systems inside the editor, and use the paint tool to place particles onto the mesh surface. Let me know if you need more info.

cailuming:
I got it, and I works as I expected finally. Another issue is the jagged edge on the texture alpha blending, or may be alpha testing? 
![image](https://user-images.githubusercontent.com/19427259/117131469-0dbb9000-add4-11eb-8450-a261e4572a18.png)
The original texture looks like this
![image](https://user-images.githubusercontent.com/19427259/117131656-478c9680-add4-11eb-98c2-62c8e4a5d651.png)
It seems that the original texture does not looks worse than on the final rendering. Maybe it's just a matter of Anti-aliasing?

By the way, I have tried to use paint tool in editor, but still confusing about how to use, such as add/remove trangle on HairParticle , when I select it and paint on the focused grass, nothing happened, Or I need to go through the related source code

turanszkij:
The jaggedness is due to alpha testing. There is no alpha blending supported now for the grass (it is too expensive to have sorting, blending and multi layered lighting). You can modify the alpha testing threshold by selecting the hair particle system in the Editor, then going to the Material window, where you can adjust the "AlphaRef" slider.

The paint tool for hair is a bit hard to use, I admit.  There are three modes for HairParticle:
- Add triangle: selected triangles of the mesh will generate particles (shown in purple debug color if the "wireframe" checkbox is enabled)
- Remove triangle: selected triangles will be removed from the hair, they will no longer generate particles (so remove triangles shown in purple debug color)
[...]

PR #264, opened 2021-05-15
atmosphere and volumetric cloud updates

Kliaxe:
Hi turanszkij!

Yet another clouds and atmosphere update. In this PR I added ground contribution lighting for the clouds. Normally a lot of light is hitting the surface of earth and bounces up onto to clouds, which we haven't accounted for, until now. This makes the clouds look even more realistic, and it's actually beginning to look a lot like what you'd see outside your window, which is quite impressive for real-time graphics. I'm impressed of how this has turned out, but I'm still very concerned about performance. Volumetric clouds in general is a very expensive effect. Should we let this effect be enabled by default, or disabled where the user manually can turn it on at will?

Another stuff in this PR is regarding ghosting and cutout artefacts which appears when geometry is moving in front of clouds. I also noticed an issue with the atmosphere, where the sun would be below the planet, but you'd still see a red solid color. These issues has been fixed in this PR.

Anyway I'm happy to contribute yet again! Keep up the good work!

Comparison of ground contribution effect: (ON/OFF)
clouds ON
clouds OFF

Comparison of atmosphere fix: (ON/OFF)
atm ON
atm OFF

### DIFF
--- a/WickedEngine/shaders/skyAtmosphere.hlsli
+++ b/WickedEngine/shaders/skyAtmosphere.hlsli
@@ -357,6 +357,15 @@ float3 GetTransmittance(AtmosphereParameters atmosphere, float pHeight, float su
 
 float3 GetAtmosphereTransmittance(float3 worldPosition, float3 worldDirection, AtmosphereParameters atmosphere, Texture2D<float4> transmittanceLutTexture)
 {
+	// If the worldDirection is occluded from this virtual planet, then return.
+	// We do this due to the low resolution LUT, where the stored zenith to horizon never reaches black, to prevent linear interpolation artefacts.
+	// At the most shadowed point of the LUT, pure black with earth shadow is never reached.
+	float2 sol = RaySphereIntersect(worldPosition, worldDirection, float3(0.0f, 0.0f, 0.0f), atmosphere.bottomRadius);
+	if (sol.x > 0.0f || sol.y > 0.0f)
+	{
+		return 0.0f;
+	}
+	
 	float pHeight = length(worldPosition);
 	const float3 UpVector = worldPosition / pHeight;
 	float SunZenithCosAngle = dot(worldDirection, UpVector);
--- a/WickedEngine/shaders/volumetricCloud_renderCS.hlsl
+++ b/WickedEngine/shaders/volumetricCloud_renderCS.hlsl
@@ -105,6 +105,8 @@ static const float g_LODDistance = 25000.0; // After a certain distance, noises 
 static const float g_LODMin = 0.0; // 
 static const float g_BigStepMarch = 3.0; // How long inital rays should be until they hit something. Lower values may ives a better image but may be slower.
 static const float g_TransmittanceThreshold = 0.005; // Default: 0.005. If the clouds transmittance has reached it's desired opacity, there's no need to keep raymarching for performance.
+static const float g_ShadowSampleCount = 5.0f;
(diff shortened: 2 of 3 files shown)


turanszkij:
This is really good improvement. The ghosting is basically gone, and the directional light behaves more consistently.

> Volumetric clouds in general is a very expensive effect. Should we let this effect be enabled by default, or disabled where the user manually can turn it on at will?

The expensive effects like most post processes are turned off by default to get the fastest performance by default, and this falls into this category. But perhaps it would be better if the volumetric clouds could act on the weather's cloudiness factor, so by default (cloudiness = 0) the post process would be a pass-through/disabled, and it would properly scale itself with higher values of the cloudiness parameter. (so cloudiness = 1 would be totally overcast). I'm not sure such a dynamic cloudiness is possible to be implemented this way. (for example now the depth of field post process is enabled by default, but only really running when camera parameters are set up for depth of field)
Let me know what you think about this.

Thanks a lot, I will merge this now as it is already a great improvement!

Kliaxe:
> Should we let this effect be enabled by default, or disabled where the user manually can turn it on at will?

Oh I actually meant the contribution effect haha. I read somewhere in #261 that you were considering exposing some of the cloud parameters, which I think could be interesting! If you do expose some of them, do you think there would be enough space inside the weather window? I had a crazy idea if the engine could support tabs UI for a window, but thats just loose thoughts.

> But perhaps it would be better if the volumetric clouds could act on the weather's cloudiness factor ...

Right now the clouds has two parameters, `g_CoverageAmount` and `g_CoverageMinimum`, which controls the cloudiness. `g_CoverageAmount` is just a multiplier for our weather map, where we probably could use that as the cloudiness, but `g_CoverageMinimum` is at the same time also needed to create totally overcast. So a single cloudiness value (0 to 1) from clear to overcast is a bit difficult to do with this setup, but as far as I can tell it's totally possible to implement.

turanszkij:
> I actually meant the contribution effect
So the new effect has to be enabled through code now? I missed it then.

Somehow unifying the weather and volumetric clouds is planned, but true that the weather window is one of the problems. Having a window tab support would be good, I am not sure when I get around to implementing it. I'll try playing with the cloud coverage sometime and see how it works.

Kliaxe:
Sorry I miscommunicated the whole thing. All in all I just suggested if performance was a concern, we could make the contribution effect something the user could enable themselves with eg. a checkbox, to save some performance with clouds enabled.

If you do get around to test the coverage parameters, I found that values like 1.5 for `g_CoverageMinimum` looks nice for overcast clouds. But for the most part, the weather map plays a significant role of how to clouds look. These values I've referenced, for instance, manipulates that pre-generated weather map. We can make all sorts of weird shapes inside the weather map generator, and adjust it accordingly with the global parameters from the cloud render pass.

turanszkij:
I think it would make sense to expose some post process settings, like the samplecount and such that control performance. It is also an option to reduce the rendering resolution of the effect. 

I've also tried the coverage settings quickly, they should definitely be exposed too as part of the weather. 

I'm working on some other stuff at the moment, but I'd like to get around to it soon.

turanszkij:
Updated the volumetric clouds:
- added light shaft occlusion support
- moved into weather
- parameter access from c++, serialization

I only added the enable, coverageamount and coveragemin controls to the weather window as there are a lot of other parameters I didn't experiment with.

PR #266, opened 2021-05-23
Fix compilation for VS2019 Windows 10

kimkulling:
- Use XAUDIO2_DEFAULT_PROCESSOR instead of XAUDIO2_USE_DEFAULT_PROCESSOR

### DIFF
--- a/WickedEngine/wiAudio.cpp
+++ b/WickedEngine/wiAudio.cpp
@@ -76,7 +76,7 @@ namespace wiAudio
 			hr = CoInitializeEx(NULL, COINIT_MULTITHREADED);
 			assert(SUCCEEDED(hr));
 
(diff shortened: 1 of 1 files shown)


turanszkij:
Hi, the current one should be used according to the header:

// This definition is included for backwards compatibilty. Implementations targeting Games and WIN10_19H1 and later, should use
// XAUDIO2_USE_DEFAULT_PROCESSOR instead to let XAudio2 select the appropriate default processor for the hardware platform.
#define XAUDIO2_DEFAULT_PROCESSOR Processor1

I think you might have an outdated windows sdk

kimkulling:
Hm, possibly. I need to check!

kimkulling:
Right, outdated WIn-SDK. Thanks!

PR #275, opened 2021-07-04
exponential height fog

Kliaxe:
Everything should now be set up. I've made a small change in relation to the discussion. Instead of using the existing fog height, I've added two separate parameters to control the height fog.

Let me know if anything needs to be changed.


turanszkij:
Thanks, I think this makes sense. It looks good, too. The weather window is getting out of hand though, I really need to do something about this soon :)

turanszkij:
Also, a shot I made

PR #277, opened 2021-07-07
Included OpenImageDenoise in cmake builds (linux)

portaloffreedom:
OpenImageDenoise is optional, will build support for the library if found installed.


turanszkij (approved):

turanszkij:
Thank you for this improvement!

PR #279, opened 2021-07-11
Editor text input fix for Linux

megumumpkin:
Added functions for text input under main_SDL2.cpp


portaloffreedom:
Would it be possible to squash all commits into only one? This is quite a mess of patches and they seems to modify only one file.

turanszkij:
Yes, I usually use the "squash and merge" for pull requests, so it doesn't need to be done as part of the commits.

turanszkij (approved):

turanszkij:
Thank you for the fixes!

PR #282, opened 2021-07-14
Updates

amerkoleci:
Updates various third party libraries and vulkan to latest heades.


turanszkij:
Could you add the dxcompiler update also to this branch too so that it's updated as one? Also, this is a lot of changes, how much did you test that it doesn't break things?

amerkoleci:
> Could you add the dxcompiler update also to this branch too so that it's updated as one? Also, this is a lot of changes, how much did you test that it doesn't break things?

I tested by running all samples and editor, loading scenes etc

turanszkij:
Thanks. It should also be tested by running with the different rendering APIs too. However, I will also do some test runs on my side.

amerkoleci:
> Thanks. It should also be tested by running with the different rendering APIs too. However, I will also do some test runs on my side.

Yes, I tested d3d11, d3d12 and vulkan, loaded sponza and other scenes, run tests etc

turanszkij (approved):

turanszkij:
Thanks for the updates and testing!

PR #285, opened 2021-07-14
D3D12 improvements

amerkoleci:
This time is D3D12:
- Enable DXGI debug only in debug mode (DXGIGetDebugInterface1 is not supported in UWP) and improve ID3D12InfoQueue logic.
- Use IDXGIFactory4 and check for variable refresh rate display support (tearing) and improve adapter detection.
- Improve SwapChain creation, handle supported formats and correctly create RTV with wanted format.

### DIFF
--- a/WickedEngine/wiGraphicsDevice_DX12.cpp
+++ b/WickedEngine/wiGraphicsDevice_DX12.cpp
@@ -2486,31 +2486,35 @@ using namespace DX12_Internal;
 				return dxgiFactory->EnumAdapters1(index, ppAdapter);
 		};
 
-		ComPtr<IDXGIAdapter1> candidateAdapter;
+		ComPtr<IDXGIAdapter1> dxgiAdapter1;
 		for (uint32_t i = 0;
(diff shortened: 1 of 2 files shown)


turanszkij:
I'd like that the dxgi debug device would be created not for `_DEBUG` build, but when the `debugdevice` is specified (as it is currently).
Also, could you explain the problem with the swapchain, I didn't fully get the reason for the rewrite.
But thanks for working on this.

amerkoleci:
> I'd like that the dxgi debug device would be created not for `_DEBUG` build, but when the `debugdevice` is specified (as it is currently).
I see OK 
> Also, could you explain the problem with the swapchain, I didn't fully get the reason for the rewrite.
If you want to create swap chain with unsupported format, let say B8G8R8A8_SRGB it would fail, you can try before my commits, but you can create swapchain with B8G8R8A8_UNORM and the RTV with B8G8R8A8_SRGB.
> But thanks for working on this.

turanszkij (approved):

turanszkij:
Thanks, nice updates!

PR #290, opened 2021-07-19
Volumetric clouds reprojection update

Kliaxe:
Hi turanszkij!

In this PR the clouds reprojection method has been improved.
I was playing a bit around with the sliders to adjust the shape of the clouds and realised the performance was dropping fast in some cases, where the clouds are thin and require more samples. No wonder why you reduced the resolution scale from 1/2 to 1/4, and this also makes sense if it has to work properly with planar reflections. There is only so much we can do with the performance settings, so adjusting the resolution would indeed be the best we could do here.
The new reprojection method allows the clouds to be upscaled from 1/4 to 1/2, etc. This technique was the primary reason volumetric clouds was able to run in the game Horizon: Zero Dawn on the PS4, with great framerate.

Let me know if anything need to be changed!
Btw the FSR was a nice addition. Didn't think it would be included this early!


turanszkij:
Thanks, sounds good I will check it out soon!

turanszkij (approved):
Very good, I was especially looking out for the quality at the geometry edges especially with alpha test + TAA, which was very poor with my downsampling to quarter res. Now it's improved a lot. Thank you!

PR #291, opened 2021-07-29
Surfel GI

turanszkij:
Diffuse GI with surfels.

TODO:
- find better coverage distribution for surfels
- find a way to reduce noise, especially after lighting changes dramatically
- optimize the hash table: try radix sorting, try to find better hash function
- implement surfels sticking to moving and skinned objects: will need to use bindless primitive IDs (rewrite instance buffer system required)
- use a gbuffer normal that is not normal mapped: GI is larger scale than normal maps, so normal maps introduce discontinuity and noise


turanszkij:
surfels on animated objects now supported, but it doesn't work with dx11 anymore because it will use bindless fetching from vertex buffers using primitiveID and barycentrics

turanszkij:
From now, surfel GI must be enabled from Renderer window -> surfel GI

PR #294, opened 2021-08-05
Samplers: Correctly implement BorderColor

amerkoleci:
- Add new SAMPLER_BORDER_COLOR
- Correctly implement under D3D11
- Correctly implement under D3D12 and handle static samplers as well
- Correctly map to VkBorderColor


turanszkij (approved):

turanszkij:
Thanks for the updates!

PR #300, opened 2021-09-06
Disable enum bitmask operators by default

qy-wang:
Should disable bitmask operators for most enum type.


turanszkij:
Hi, was this an issue for you? What was the problem?

qy-wang:
> Hi, was this an issue for you? What was the problem?

Not a bug now, but potentially, it could be.
If `enable = true` by default, can apply `|` & `|=` for all enums, include those we don't expect to do.

turanszkij:
You're right, I didn't realize it actually worked for all by default.

PR #305, opened 2021-09-08
Volumetric cloud sample count slider for QOL improvement.

TheIronLefty:


turanszkij:
Hi, thanks for the interest. I'd like to not expose the sample count in this way. First of all, your slider places it to 256 samples, but by default the clouds are using 128. The slider doesn't take into account what's the current weather's cloud step count, so if I load a scene, it's still sitting at 256. You could do this in the WeatherWindow::Update() function, calling the slider's SetValue() with the current weather's right parameters, that would be better.

Second, sample counts above 128 are not usable realistically, the render will take too long. I don't want users to set a sample count very much, because it will cause performance issues and saved into the scene, then it will be hard for them to find the problem. But maybe a tooltip would be a good warning.

Also, I didn't understand why you want to set power of two sample counts, I think it would be better to instead find a better range, like 64 - 256 is maybe reasonable, with default being 128.

PR #306, opened 2021-09-10
- Improvements round 2

amerkoleci:
- Remove SHADERFORMAT_HLSL5 and rename SHADERFORMAT_HLSL6 to SHADERFORMAT_DXIL which makes more sense.
- Remove SHADERMODEL_5_0 and legacy D3DCompiler shader support.
- GetDescriptorIndex is abstract as we always support bindless index.
- Vulkan: Add new VK_CHECK to check and log error during calls.


WickedEngine/wiShaderCompiler.cpp
```diff
@@ -443,139 +438,6 @@ namespace wiShaderCompiler
 	}
 #endif // SHADERCOMPILER_ENABLED_DXCOMPILER
 
-#ifdef SHADERCOMPILER_ENABLED_D3DCOMPILER
```
turanszkij, line 446: I actually wanted to keep this here, to have a reference of how the old d3dcompiler interface can be used. This will come in handy when somebody searches for d3d compile functions on github. Keeping it here and disabled is fine.

WickedEngine/wiShaderCompiler.cpp
```diff
@@ -595,19 +457,6 @@ namespace wiShaderCompiler
 			}
 		}
 #endif // SHADERCOMPILER_ENABLED_DXCOMPILER
-
-#ifdef SHADERCOMPILER_ENABLED_D3DCOMPILER
```
turanszkij, line 599: Same, just keep the d3dcompiler code parts and keep them disabled. It's good if we need to test something that we can just re enable it.

turanszkij (changes requested):

amerkoleci:
I see, will actually close this PR then :)

PR #313, opened 2021-09-17
Basis Universal

turanszkij:
-KTX2 transcoding implemented (BC1 or BC3 destination so far)
-added KHR_texture_basisu gltf extension support

TODO: 
-basis transcode
-encoders
-resource manager could request specific transcode destination format from outside (would be good for normal maps)
-lightmap compression


turanszkij:
lightmap compression will need to be handled differently as this library doesn't seem to support BC6H

turanszkij:
Added some parts of DirectXTex for missing BC compressors, eg BC6 for lightmaps

PR #316, opened 2021-09-20
Vulkan improvements.

amerkoleci:
- Cleanup core 1.2 feature checks.
- Improve SwapChain creation logic and handle capabilities in better way.

### DIFF
--- a/WickedEngine/wiGraphicsDevice_Vulkan.cpp
+++ b/WickedEngine/wiGraphicsDevice_Vulkan.cpp
@@ -1894,6 +1894,10 @@ using namespace Vulkan_Internal;
 				debugUtils = true;
 				instanceExtensions.push_back(VK_EXT_DEBUG_UTILS_EXTENSION_NAME);
 			}
+			else if (strcmp(availableExtension.extensionName, VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME) == 0)
+			{
+				instanceExtensions.push_back(VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME);
+			}
 		}
 		
 		instanceExtensions.push_back(VK_KHR_SURFACE_EXTENSION_NAME);
@@ -1994,8 +1998,10 @@ using namespace Vulkan_Internal;
 					continue;
 
 				features2.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_FEATURES_2;
+				features_1_1.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_VULKAN_1_1_FEATURES;
 				features_1_2.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_VULKAN_1_2_FEATURES;
-				features2.pNext = &features_1_2;
+				features2.pNext = &features_1_1;
+				features_1_1.pNext = &features_1_2;
 				void** features_chain = &features_1_2.pNext;
 				acceleration_structure_features = {};
 				raytracing_features = {};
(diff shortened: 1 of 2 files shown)


WickedEngine/wiGraphicsDevice_Vulkan.cpp
```diff
@@ -2283,10 +2267,7 @@ using namespace Vulkan_Internal;
 		allocatorInfo.physicalDevice = physicalDevice;
 		allocatorInfo.device = device;
 		allocatorInfo.instance = instance;
-		if (features_1_2.bufferDeviceAddress)
```
turanszkij, line 2286: I think this should still be checked for

WickedEngine/wiGraphicsDevice_Vulkan.cpp
```diff
@@ -2962,12 +2956,9 @@ using namespace Vulkan_Internal;
 			bufferInfo.usage |= VK_BUFFER_USAGE_ACCELERATION_STRUCTURE_BUILD_INPUT_READ_ONLY_BIT_KHR;
 			bufferInfo.usage |= VK_BUFFER_USAGE_SHADER_BINDING_TABLE_BIT_KHR;
 		}
-		if (features_1_2.bufferDeviceAddress == VK_TRUE)
```
turanszkij, line 2965: This needs to be checked too

turanszkij (changes requested):

WickedEngine/wiGraphicsDevice_Vulkan.cpp
amerkoleci: Sure? Even if it is core feature in vulkan 1.2?

turanszkij: It's not an extension any more but the device will not necessarily support it, that's why you can check support in the features_1_2

WickedEngine/wiGraphicsDevice_Vulkan.h
```diff
@@ -46,15 +46,13 @@ namespace wiGraphics
 		VkQueue copyQueue = VK_NULL_HANDLE;
 
 		VkPhysicalDeviceProperties2 properties2 = {};
-		VkPhysicalDeviceVulkan11Properties properties_1_1 = {};
```
turanszkij, line 49: also, pls don't remove these (`properties_1_1` and `features_1_1`) , it's useful for looking into these properties with debugger currently, and if we want something from here in the future.

turanszkij (changes requested):

WickedEngine/wiGraphicsDevice_Vulkan.cpp
```diff
@@ -2494,12 +2506,30 @@ using namespace Vulkan_Internal;
 		dynamicStateInfo.dynamicStateCount = (uint32_t)pso_dynamicStates.size();
 		dynamicStateInfo.pDynamicStates = pso_dynamicStates.data();
 
-		allocationhandler->bindlessSampledImages.init(device, VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE, properties_1_2.maxDescriptorSetUpdateAfterBindSampledImages / 4);
-		allocationhandler->bindlessUniformTexelBuffers.init(device, VK_DESCRIPTOR_TYPE_UNIFORM_TEXEL_BUFFER, properties_1_2.maxDescriptorSetUpdateAfterBindSampledImages / 4);
-		allocationhandler->bindlessStorageBuffers.init(device, VK_DESCRIPTOR_TYPE_STORAGE_BUFFER, properties_1_2.maxDescriptorSetUpdateAfterBindStorageBuffers / 4);
-		allocationhandler->bindlessStorageImages.init(device, VK_DESCRIPTOR_TYPE_STORAGE_IMAGE, properties_1_2.maxDescriptorSetUpdateAfterBindStorageImages / 4);
-		allocationhandler->bindlessStorageTexelBuffers.init(device, VK_DESCRIPTOR_TYPE_STORAGE_TEXEL_BUFFER, properties_1_2.maxDescriptorSetUpdateAfterBindStorageImages / 4);
-		allocationhandler->bindlessSamplers.init(device, VK_DESCRIPTOR_TYPE_SAMPLER, 256);
+		if (features_1_2.descriptorBindingSampledImageUpdateAfterBind == VK_TRUE)
```
turanszkij, line 2509: Thanks for fixing the others. These ifs are probably ok to be removed (but they can stay too), because descriptorIndexing is required, and without these the renderer won't work properly anyway.

turanszkij (approved):

WickedEngine/wiGraphicsDevice_Vulkan.cpp
```diff
@@ -2513,30 +2494,13 @@ using namespace Vulkan_Internal;
 		dynamicStateInfo.dynamicStateCount = (uint32_t)pso_dynamicStates.size();
 		dynamicStateInfo.pDynamicStates = pso_dynamicStates.data();
 
-		if (features_1_2.descriptorBindingSampledImageUpdateAfterBind)
-		{
```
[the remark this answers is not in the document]
turanszkij: These checks are not necessary, because the renderer won't work correctly without these bindless features.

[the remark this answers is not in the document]
turanszkij: But it's also ok to keep them, not sure what's the best way

PR #317, opened 2021-09-20
Linux Audio Support

megumumpkin:
Finalized version of Linux Audio, this one is using FAudio.
Added FAudio lib into the engine (found in WickedEngine/Utils/FAudio)


turanszkij:
The windows build is failing because `#include <SDL_stdinc.h>` is included in the windows build (wiAudio.cpp). Please fix.

PR #320, opened 2021-09-22
bloom and eye adaption updates

turanszkij:
-eye adaption is now histogram-based
-bloom affected by expose and eye adaption
-bloom is no longer in post process chain, it's applied in tonemap shader instead
-path tracing bloom enabled


savegame:
@turanszkij  lokkslike this PR broke render on linux Vulkan     

изображение    
it looks normal only close ibjects    
изображение  
littlebit_far

savegame:
when i downgrade to previous commit, it looks normal

turanszkij:
Did you rebuild shaders? This update got some mipmap generation shaders rewritten, so you will need to rebuild them.

savegame:
no, iam not =) ok, what i should run for rebuild shaders only& or what delete? for build it on cmake/make step ?

turanszkij:
The shaders will be built with cmake on Linux. I am not really familiar with it, so refer to how you build the project with cmake. You might need to delete your shaders too so that they are rebuilt from scratch.

savegame:
ok, thanks! i'll delete them all )

turanszkij:
Yes, it's the .cso files, right

PR #322, opened 2021-09-25
Added more settings for backlog

flipdp:
Added things like disabling lua execution, setting log levels or completely disabling the backlog

### DIFF
--- a/WickedEngine/wiBackLog.h
+++ b/WickedEngine/wiBackLog.h
@@ -23,8 +23,7 @@ namespace wiBackLog
 
 	std::string getText();
 	void clear();
-	void post(const std::string& input);
-	void post(const std::string& input, LogLevel level);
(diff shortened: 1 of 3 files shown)


WickedEngine/wiBackLog.h
```diff
@@ -16,6 +24,7 @@ namespace wiBackLog
 	std::string getText();
 	void clear();
 	void post(const std::string& input);
+	void post(const std::string& input, LogLevel level);
```
turanszkij, line 27: This could have been just a default parameter instead of two functions.

turanszkij (approved):
Everything looks ok, there is one minor thing that you can fix if you want to, with the default parameter.

turanszkij (approved):

turanszkij:
Thanks, it's in!

PR #332, opened 2021-10-11
Vulkan Pipeline Cache

amerkoleci:
- Implement pipeline cache support, verifies and loads cache validation.
- wiHelper: Add GetTempDirectoryPath and use it in vulkan cache lookup.

Next step: Implement ID3D12PipelineLibrary  for D3D12 backend

### DIFF
--- a/WickedEngine/shaders/brdf.hlsli
+++ b/WickedEngine/shaders/brdf.hlsli
@@ -419,8 +419,7 @@ struct Surface
 
 		if (material.options & SHADERMATERIAL_OPTION_BIT_ADDITIVE)
 		{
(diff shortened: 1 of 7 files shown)


amerkoleci:
> > > I haven't found any issues, tested on Windows only. It works very well.
> > 
> > 
> > Did you notices the performance improvements?
> 
> Yes, definitely. The second start of application is faster, and when loading a scene after second application launch, there is no more freeze. Also, when placing env probe, there is no more freeze.

Exactly, thats what the pipeline cache is doing, prewarm and created faster pipeline after cache being created, we need D3D12 pipeline cache now :)

PR #337, opened 2021-11-02
Linux Network Implementation

megumumpkin:
Implemented using standard socket library provided by the Linux system.
Reference: [https://www.binarytides.com/socket-programming-c-linux-tutorial/](https://www.binarytides.com/socket-programming-c-linux-tutorial/)

Also deleted the destroy function in the wiNetwork header and the Linux code since socket is destroyed automatically when it is out of scope.

### DIFF
--- a/WickedEngine/wiNetwork_Linux.cpp
+++ b/WickedEngine/wiNetwork_Linux.cpp
@@ -1,22 +1,27 @@
 #include "wiPlatform.h"
-#include <cstring>
 
 #ifdef PLATFORM_LINUX
 #include "wiNetwork.h"
 #include "wiBackLog.h"
 #include "wiTimer.h"
 
-#include <sstream>
-#include <string>
 #include <unistd.h>
 #include <errno.h>
 
 #include <sys/socket.h>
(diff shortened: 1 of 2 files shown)


WickedEngine/wiNetwork_Linux.cpp
```diff
@@ -1,45 +1,184 @@
...
+		return true;
 	}
 
 	bool Send(const Socket* sock, const Connection* connection, const void* data, size_t dataSize)
 	{
+		if (sock->IsValid()){
+			sockaddr_in target;
+			target.sin_family = AF_INET;
+			target.sin_port = htons(connection->port);
+			std::stringstream ipss;
```
turanszkij, line 67: This part will use heap allocation (due to `std::string` objects), which is not very good if the Send() will be called frequently. I think the Send() and Receive() functions are probably performance critical in how a networking code might use them.
But this looks like pretty similar to the win32 implementation, where string objects are not used. 
The same for the Receive() function.

turanszkij (changes requested):

turanszkij (approved):

PR #339, opened 2021-11-03
Added runtime shader compiler on linux

portaloffreedom:


turanszkij (commented):
I reverted the dxcapi.h modifications along with the path change. This is needed to more easily update the dxcapi.h file in the future. Let's see if it still works correctly for linux build.

WickedEngine/wiJobSystem.cpp
```diff
@@ -73,13 +86,13 @@ namespace wiJobSystem
 		{
 			std::thread worker([] {
 
-				while (true)
+				while (jobs_alive)
```
portaloffreedom, line 89: Isn't this thread unsafe?

turanszkij: Yes, it should be atomic for one, and not be destroyed until any workers are alive. Probably need a better approach.

turanszkij (approved):

PR #341, opened 2021-11-05
HDR display support

turanszkij:
Currently sort of replicates vulkan's automatic HDR behaviour. Needs more testing. Difference is that vulkan will somehow manage to force the HDR on for a monitor, while dx12 respects the monitor setting.


turanszkij:
Todo: shaders need to handle HDR and SDR elements blending correctly

amerkoleci (approved):

PR #342, opened 2021-11-05
Add imgui sdl2 adapter

portaloffreedom:


portaloffreedom:
keyboard input does not work, probably WickedEngine eats all events and they are not parsable by imgui. Need some time to figure this part out. Mouse input works no problem.

turanszkij (approved):

turanszkij:
I think this will be an improvement already so can be merged, or it can be updated further too.

amerkoleci:
> keyboard input does not work, probably WickedEngine eats all events and they are not parsable by imgui. Need some time to figure this part out. Mouse input works no problem.

Make sure you call ImGui_ImplSDL2_ProcessEvent and everything will be cool :)

Just like here: https://github.com/ocornut/imgui/blob/master/examples/example_sdl_directx11/main.cpp#L100

turanszkij (approved):

PR #352, opened 2021-11-09
Fixed crash on wayland exit

portaloffreedom:
The engine would crash on exit because SDL handles the deletion of the swapchain on his own, making the engine delete the swapchain twice, which would result in a crash on wayland.
This fix checks if the Video Subsystem is not running any more, if not it will assume there is no need to destroy the swap chains any more.

If there is a better way to test if the swapchain is already destroyed, even better :)

### DIFF
--- a/WickedEngine/wiGraphicsDevice_Vulkan.cpp
+++ b/WickedEngine/wiGraphicsDevice_Vulkan.cpp
@@ -929,7 +929,7 @@ namespace Vulkan_Internal
 			// If so we would delete the swapchain twice, causing a crash on wayland.
 			if (SDL_WasInit(SDL_INIT_VIDEO))
 			#endif
-            {
+			{
 				vkDestroySwapchainKHR(allocationhandler->device, swapChain, nullptr);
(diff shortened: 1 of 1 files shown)


turanszkij (approved):

turanszkij:
There is a way to test whether the swapchain is destroyed, by calling `swapChain.IsValid()`, but it will only be destroyed together with the Maincomponent usually when the program exits.

portaloffreedom:
I'm going to check if the `swapchain.IsValid()` approach works first. I would like to keep hacks at a minimum if possible.

turanszkij:
I am looking at swapchain destructor, and I think it might be a problem of destroying swapchain objects while GPU might still use them. As a quick workaround, you could add this to the beginning of swapchain_Vulkan's destructor:
```cpp
vkDeviceWaitIdle(allocationhandler->device);
```

turanszkij:
You can leave it like this, I have a better destruction in the hdr branch so the DeviceWaitIdle() will be replaced

PR #354, opened 2021-11-09
Linux GameMode Support

megumumpkin:
GameMode sets up performance optimizations on the OS to make sure the engine can utilize the hardware better on Linux.

### DIFF
--- a/WickedEngine/Utility/gamemode_client.h
+++ b/WickedEngine/Utility/gamemode_client.h
@@ -1,0 +1,365 @@
+/*
+
+Copyright (c) 2017-2019, Feral Interactive
+All rights reserved.
+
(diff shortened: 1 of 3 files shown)


turanszkij (approved):

portaloffreedom:
It bothers me that we are not calling `gamemode_request_end();` at the end of the program...

turanszkij:
> It bothers me that we are not calling `gamemode_request_end();` at the end of the program...

All this could be the linux application's responsibility in my opinion, not really needed to do this as part of engine's initialization, after all it's very linux specific. And then the application could call `gamemode_request_end();` before exiting without any problem.

turanszkij:
This can be part of the application, probably doesn't need to be engine feature.

PR #363, opened 2021-11-25
Added Focus events on SDL editor

portaloffreedom:


turanszkij (dismissed):

turanszkij (approved):
Looking at the large filesize of the Editor and Tests executables, it means that the shaders were compiled successfully and were embedded into the application.

PR #367, opened 2021-11-29
version 0.60

turanszkij:
The namespaces will be refactored into nested namespaces with the `namespace wi` being the root namespace for the engine. The existing namespaces starting with `wi` will be refactored to be in wi:: followed by their remaining name, in lowercase. For example:
`wiGraphics` -> `wi::graphics`

Todos:
- to keep backwards compatibility with user code, provide namespace aliases for the old namespaces, for example: `namespace wiGraphics = wi::graphics;`
- update the documentation to reflect changes 
- cleanup


turanszkij:
In the meantime, added some other improvements unrelated to namespace refactor:
- resource manager will return `Resource` instead of `shared_ptr<Resource>` objects
- shader optimizations with MAD instruction
- implemented alpha to coverage with alpha tested materials when MSAA is enabled
- alpha testing fix with transparent shadow maps
- TLAS and scene buffers will be recreated less frequently when things get added/removed from the scene

PR #368, opened 2021-12-05
general improvements

turanszkij:
The performance of wi::vector has been improved compared to std::vector (both in Debug and Release builds).
The "Container perf" test case in the Tests app shows the difference with a simplified test.


turanszkij:
compiler crash what??

turanszkij:
std::vector will still be used by default.

PR #371, opened 2021-12-07
Add SetDepthBounds.

amerkoleci:
- Add BindDepthBounds, this binds to OMSetDepthBounds in D3D12 and vkCmdSetDepthBounds with VK_DYNAMIC_STATE_DEPTH_BOUNDS on Pipeline.
- Add DEPTH_BOUNDS_TEST capability
- Add depth_bounds_test_enable to DepthStencilState

### DIFF
--- a/WickedEngine/wiApplication.cpp
+++ b/WickedEngine/wiApplication.cpp
@@ -72,7 +72,6 @@ namespace wi
 
 	void Application::Run()
 	{
-
 		if (!initialized)
 		{
 			// Initialize in a lazy way, so the user application doesn't have to call this explicitly
--- a/WickedEngine/wiGraphics.h
+++ b/WickedEngine/wiGraphics.h
@@ -354,6 +354,7 @@ namespace wi::graphics
 		RAYTRACING = 1 << 9,
 		PREDICATION = 1 << 10,
 		SAMPLER_MINMAX = 1 << 11,
+		DEPTH_BOUNDS_TEST = 1 << 12,
 	};
 
 	enum class ResourceState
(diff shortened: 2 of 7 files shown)


turanszkij (changes requested):
- SetDepthBounds function name should be BindDepthBounds for consistency with other names.
- Depth bounds testing must be first enabled in the depth stencil state in dx12: https://docs.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_depth_stencil_desc1

Also, out of curiosity, might I ask what do you want to use it for? What should be the test case for it?

amerkoleci:
> * SetDepthBounds function name should be BindDepthBounds for consistency with other names.
> * Depth bounds testing must be first enabled in the depth stencil state in dx12: https://docs.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_depth_stencil_desc1
> 
> Also, out of curiosity, might I ask what do you want to use it for? What should be the test case for it?

I updated as you suggested, about usage?
I know depth bounds can be used during shadow rendering to set the depth range of light, feel free to reject it if you don't want to add it :)

turanszkij:
Also, this is optional feature in dx12, so it needs some checking: https://docs.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-id3d12graphicscommandlist1-omsetdepthbounds

```
OMSetDepthBounds is an optional feature. Use the CheckFeatureSupport method to determine whether or not this feature is supported by the user-mode driver. Support for this feature is reported through the D3D12_FEATURE_D3D12_OPTIONS1 structure.
```

turanszkij (approved):

turanszkij:
Thanks!

PR #372, opened 2021-12-10
DX12 custom root signatures

turanszkij:
This implements support for custom root signature support for DX12 graphics device.
- custom root signature can be defined inside the shader
- if no custom root signature is defined, a default one will be used
- removed dependency on dxcompiler.dll
- removed usage of dxcompiler shader reflection
- added reflection based on root signature deserializer
- vulkan can not use custom root signatures for now, it will still use spirv reflection

TODO:
- optimize new resource binding code
- reduce size of default root signature
- add custom root signatures for most important shaders
- raytracing library can still not handle custom root signature, implement it


turanszkij:
todo: the ray traced reflection is crashing gpu for unknown reasons

Issue #383, opened 2022-01-04
SPH viscosity smoothing kernel correctness


yhamdoud:
Hi,

I think that the viscosity smoothing kernel is erroneously applied instead of the Laplacian of this smoothing kernel in the SPH fluid simulation. I'm far from being an expert on fluid dynamics, so it would be nice if someone could confirm my findings.

I came across [a blog post on WickedEngine's SPH implementation](https://wickedengine.net/2018/05/21/scalabe-gpu-fluid-simulation/) when researching the topic. Something caught my eye in the section covering the viscosity force term computation. In the pseudocode, the smoothing kernel W is calculated and applied as follows:
```glsl
const float W = -(r3 / (2 * h3)) + (r2 / h2) + (h / (2 * r)) - 1;
particleA.force2 += (particleB.mass / particleA.mass) * (1.0f / particleB.density) * (particleB.velocity - particleA.velocity) * W * rNorm;
```
However, if we look at [the article that the post references](http://rlguy.com/sphfluidsim/), it calls for applying the Laplacian of the given smoothing kernel. I had a look at the accompanying source code of the article and indeed, [it seems to calculate and use the constant part of the Laplacian of W.](https://github.com/rlguy/SPHFluidSim/blob/07548003daf6ebcc3020f6fed37e30981d9f7e81/src/sphfluidsimulation.cpp#L71).

This is also in line with one of the papers I looked at, [Particle-Based Fluid Simulation for Interactive Applications](https://matthias-research.github.io/pages/publications/sca03.pdf):

![image](https://user-images.githubusercontent.com/18217298/147994174-806f8218-7407-4e19-92a6-470a3e12de99.png)

The relevant code doesn't seem to be changed since the blog was post was authored:
https://github.com/turanszkij/WickedEngine/blob/dfcdbdd40c5ff8d5425063258dbef740150cbd7e/WickedEngine/shaders/emittedparticle_sphforceCS.hlsl#L119-L120
I've changed it and loaded the emitter_fluid_50k scene and the simulation seems to be a bit more stable now. This is after tweaking the emitter settings a bit, viscosity in particular. I'm not sure if there is a better test scene, since I haven't actively used WickedEngine before.

Thanks for your time, I hope I'm not completely of the mark here!

turanszkij:
That part is computed here: 
[...]

PR #384, opened 2022-01-04
Use correct smoothing kernel for viscosity force calculation

yhamdoud:
#383

### DIFF
--- a/WickedEngine/shaders/ShaderInterop_EmittedParticle.h
+++ b/WickedEngine/shaders/ShaderInterop_EmittedParticle.h
@@ -75,19 +75,22 @@ CBUFFER(EmittedParticleCB, CBSLOT_OTHER_EMITTEDPARTICLE)
 	float		xSPH_spiky_constant;	// precomputed Spiky kernel function constant term
 	float		xSPH_visc_constant;	    // precomputed viscosity kernel function constant term
 	float		xSPH_K;					// pressure constant
-	float		xSPH_p0;				// reference density
 
 	float		xSPH_e;					// viscosity constant
+	float		xSPH_p0;				// reference density
 	uint		xEmitterOptions;
 	float		xEmitterFixedTimestep;	// we can force a fixed timestep (>0) onto the simulation to avoid blowing up
-	float		xEmitterRestitution;
 
 	float3		xParticleGravity;
-	float		xParticleDrag;
+	float		xEmitterRestitution;
 
 	float3		xParticleVelocity;
-	float		xParticleRandomColorFactor;
+	float		xParticleDrag;
 
+	float		xParticleRandomColorFactor;
+	int padding0;
+	int padding1;
+	int padding2;
 };
 
 static const uint THREADCOUNT_EMIT = 256;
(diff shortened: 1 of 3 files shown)


WickedEngine/shaders/ShaderInterop_EmittedParticle.h
```diff
@@ -73,6 +73,7 @@ CBUFFER(EmittedParticleCB, CBSLOT_OTHER_EMITTEDPARTICLE)
 
 	float		xSPH_poly6_constant;	// precomputed Poly6 kernel constant term
 	float		xSPH_spiky_constant;	// precomputed Spiky kernel function constant term
+	float		xSPH_visc_constant;	    // precomputed viscosity kernel function constant term
```
turanszkij, line 76: This could be a problem putting this value here. The members of constant buffer structures must be following the constant buffer packing rules, that's why the members are grouped into 16-byte regions (for example 4 floats following each other). The constant buffers must be padded explicitly, some info is in the documentation: https://github.com/turanszkij/WickedEngine/blob/master/Content/Documentation/WickedEngine-Documentation.md#shaders

turanszkij (changes requested):
There is one problem with constant buffer alignment, but I'll also need to review the shader logic changes later.

WickedEngine/shaders/ShaderInterop_EmittedParticle.h
yhamdoud: It should be packed correctly now.

WickedEngine/shaders/emittedparticle_sphforceCS.hlsl
```diff
@@ -115,8 +115,7 @@ void main( uint3 DTid : SV_DispatchThreadID, uint groupIndex : SV_GroupIndex, ui
 
 								f_a += mass * ((pressureA + pressureB) / (2 * densityA * densityB)) * W * rNorm;
 
-								float r3 = r2 * r;
-								W = -(r3 / (2 * h3)) + (r2 / h2) + (h / (2 * r)) - 1; // laplacian smoothing function
```
turanszkij, line 119: Sorry, I've been very busy. But I don't understand, why is the laplacian smoothing function completely removed now from the equations?

turanszkij: I think the problem is that I don't understand what the triangle means in this: 
image

yhamdoud: The triangle (nabla) squared is the Laplace operator. When you apply it to the viscosity kernel above you end up with the expression I posted in the issue earlier. This shows it more explicitly:

image

So the Laplacian of the smoothing function is applied, but I removed the comment. Sorry for the confusion.

turanszkij: Thanks. Since it's more closely matches the original article's implementation, this will be merged. I will probably modify the current fluid simulation sample's viscosity to better match the original look. And probably better default values for viscosity should be used too.

turanszkij (approved):

turanszkij:
Thank you for the contribution!

PR #387, opened 2022-01-17
Improvements

amerkoleci:
ThirdParty: 
- Update vulkan headers to 1.2.203
- Update volk 
- Update VMA and integration with recent changes.
- Update D3D12MA.


WickedEngine/wiGraphicsDevice_Vulkan.cpp
```diff
@@ -2767,14 +2767,45 @@ using namespace vulkan_internal;
...
+		vma_vulkan_func.vkDestroyImage = vkDestroyImage;
+		vma_vulkan_func.vkCmdCopyBuffer = vkCmdCopyBuffer;
+
 		VmaAllocatorCreateInfo allocatorInfo = {};
 		allocatorInfo.physicalDevice = physicalDevice;
 		allocatorInfo.device = device;
 		allocatorInfo.instance = instance;
+
+		// Core in 1.1
+		allocatorInfo.flags = VMA_ALLOCATOR_CREATE_KHR_DEDICATED_ALLOCATION_BIT | VMA_ALLOCATOR_CREATE_KHR_BIND_MEMORY2_BIT;
```
turanszkij, line 2797: Why do we need these flags now?

WickedEngine/wiGraphicsDevice_DX12.cpp
```diff
@@ -2319,6 +2358,12 @@ using namespace dx12_internal;
 					//D3D12_MESSAGE_ID_UNMAP_INVALID_NULLRANGE,
 					//D3D12_MESSAGE_ID_EXECUTECOMMANDLISTS_WRONGSWAPCHAINBUFFERREFERENCE,
 					D3D12_MESSAGE_ID_SETPRIVATEDATA_CHANGINGPARAMS,
+					D3D12_MESSAGE_ID_RESOURCE_BARRIER_MISMATCHING_COMMAND_LIST_TYPE,
```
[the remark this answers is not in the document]
amerkoleci: There was bug, reported in d3d12 discord with Windows 11 reporting validation layer issue.

WickedEngine/wiGraphicsDevice_Vulkan.cpp
amerkoleci: VMA can make use of dedicated allocation and bind memory extensions which are core in 1.1, if you want we can disable them.

turanszkij: It seems that there is equivalent functionality in core, but these are using KHR in their names. Does VMA use the core or KHR extensions? Because if it uses extension, I think it needs to be enabled separately.

turanszkij (approved):

WickedEngine/wiGraphicsDevice_Vulkan.cpp
amerkoleci: Not sure, I've baaed the implementation on vulkan samples: https://github.com/KhronosGroup/Vulkan-Samples/blob/master/framework/core/device.cpp#L226 feel free to change it

PR #388, opened 2022-01-22
Surfel GI updates

turanszkij:
- performance improvement
- moment integration improvement
- fixes for other raytracing shaders


turanszkij:
- shadow cascade depth clamping
- hairparticles will render after solid meshes

PR #394, opened 2022-02-21
SDL SetPointer support

flipdp:
Added SDL SetPointer support and changed mouse delta calculation


turanszkij (approved):

turanszkij:
Nice one, thanks! I think you will be able to merge it when you are ready and the build passed.

PR #395, opened 2022-02-25
Traced Reflection Improvements

Kliaxe:
- SSR hierarchy based tracing
- SSR tile classified rendering between HiZ and Marching
- Improved resolve pass
- Improved temporal pass
- Replaced median pass with bilateral blur pass
- Traced reflection now has parity.

Thanks for the patience. Finally got the time to finish things up! 
Changes has been made for both raytracing and SSR, so they share the same spatial reconstruction and denoising, so I wonder if we can merge the two render function together (wiRenderer). I could have done that, but for simplicity I solely wanted to focus on the rendering techniques. 
An observation is that surface.Load is very expensive to calculate in full-res which other passes need, so that's why ssr_surfaceCS comes into play to share it's outputs.
Also the blur pass could probably be optionable. I was thinking that some might skip the blur for performance or visual preference. What do you think?

If you have any questions, or have any suggestions please let me know!
Once again, thanks for letting me contribute!


WickedEngine/wiRenderer.cpp
```diff
@@ -9414,383 +9443,951 @@ void Postprocess_RTReflection(
...
+		device->EventBegin("SSR Depth hierarchy pass", cmd);
+		device->BindComputeShader(&shaders[CSTYPE_POSTPROCESS_SSR_DEPTHHIERARCHY], cmd);
+
+		TextureDesc hierarchyDesc = res.texture_depth_hierarchy.GetDesc();
 
-		device->BindResource(&output, 0, cmd);
-		device->BindResource(&res.texture_temporal[temporal_history], 1, cmd);
-		device->BindResource(&res.rayLengths, 3, cmd);
+		{
+			device->BindResource(&depthBuffer_Main, 0, cmd);
```
turanszkij, line 10044: The depthBuffer_Main no longer needs to be bound, it is available always globally in shaders as `texture_depth` (in globals.hlsli)

turanszkij (changes requested):
I have one comment so far, not that important. Other than that the raytraced reflections don't work with vulkan and getting validation errors from it too. This needs to be fixed before merging, but I can help with that too. 
I merged this into a branch here on the main engine and opened a new PR: https://github.com/turanszkij/WickedEngine/pull/397

turanszkij:
> I wonder if we can merge the two render function together (wiRenderer).

This can wait, but it makes sense, maybe SSR can fall back on raytraced reflection when it misses.

> An observation is that surface.Load is very expensive to calculate in full-res which other passes need, so that's why ssr_surfaceCS comes into play to share it's outputs.

Yes, I plan to improve the surface.Load by for example reducing amount of vertex buffers and adding mip mapping. But sharing the results with multiple passes is also fine.

> Also the blur pass could probably be optionable. I was thinking that some might skip the blur for performance or visual preference.

I'm ok with having this option.

Thanks!

WickedEngine/wiRenderer.cpp
turanszkij: (The idea is that commonly used textures and resources are now available with bindless system, and accessible from `GetCamera()`, or `GetScene()`, or `GetFrame()` in the shaders. For example the `texture_depth` is accessible from camera, but you can just use `texture_depth` macro too)

turanszkij:
I also see that `output_surface_environment` is used in the shaders, but never bound from c++ side. Also, the environment reflection I don't think needs to be sampled by SSR, because it will be sampled by object shaders to fall back on SSR.

WickedEngine/wiRenderer.cpp
```diff
@@ -9414,383 +9443,951 @@ void Postprocess_RTReflection(
...
+		}
+
+		// Vertical:
+		{
+			postprocess.params0.x = 0;
+			postprocess.params0.y = 1;
+			device->PushConstants(&postprocess, sizeof(postprocess), cmd);
+
+			const GPUResource* resarray[] = {
+				&output,
```
turanszkij, line 9752: `output` texture is bound here as Read Only resource, and bound as UAV (Read + Write) resource at the same time, this will not work correctly.

turanszkij (changes requested):

turanszkij:
There is a lot of great stuff here, but will need some corrections. We can continue also in the new branch here: https://github.com/turanszkij/WickedEngine/pull/397 or also your own branch if you prefer.

WickedEngine/wiRenderer.cpp
Kliaxe: You are right, I should've done that from the start. Just made a commit with your suggestion. I thought at the time of getting the rawest depth, but as you mentioned they're the same.

WickedEngine/wiRenderer.cpp
Kliaxe: Good eye, that totally glossed over my head! Uploaded the fix

Kliaxe:
>  the raytraced reflections don't work with vulkan and getting validation errors from it too, but I can help with that too.

Interesting. Wish I could figure this problem out now, but I'm not near any raytracing machine atm. Much appreciated if you could  help with this!

> I also see that output_surface_environment is used in the shaders, but never bound from c++ side

This was solely for the raytraced reflections since, it requires a global reflection fallback on early exit. SSR does not use that and does indeed fallback to screen info. RT Reflections and SSR use the same surface pass, but we could probably add a branch on the output if necessary

turanszkij:
The vulkan issue appeared to be missing a PushConstant() call after shader change. Submitted PR into your repo for it. With the fix, everything appears to be working.

turanszkij:
There was also an issue with the shared normals being stored in an unsigned format in SSR, I changed the format to `R8G8B8A8_SNORM` and it seems to be all right. This way the shader doesn't need to convert back and forth between [0,1] and [-1,1] ranges manually.

Kliaxe:
Thanks for the help! Your change has been merged.

> An other note, that SSR can be run on async compute under shadow map rendering, but raytraced reflections must be after shadow map rendering

That makes sense. We could then perhaps make a shared denoising function to avoid code duplication for the stuff in the reflection that has no difference.

turanszkij:
> Thanks for the help! Your changes has been merged.
> 
> > An other note, that SSR can be run on async compute under shadow map rendering, but raytraced reflections must be after shadow map rendering
> 
> That makes sense. We could then perhaps make a shared denoising function to avoid code duplication for the stuff in the reflection that has no difference.

This can be done later if needed, thanks for the contributions, I'll merge this!

turanszkij (approved):

turanszkij:
I removed the surface pass from SSR. The visibilityResolve pass which is separate from SSR will now create SSR surface params if they are needed. This is now not the responsibility of SSR and RT reflections.

PR #402, opened 2022-03-13
Linux package files

MolassesLover:
## About
A pull request full of changes relating to GNU/Linux package creation and distribution.

Screenshot from 2022-03-13 15-15-10

## Changes
- Wrote an **early** installer script.
- Added a `.desktop` file as seen in the screenshot.
- Wrote a launcher script, fixing a launch bug.

---

> Wallpaper from D&D 5e's Player Handbook, by Wizards of the Coast.
> The icon theme is [Tela Circle](https://github.com/vinceliuice/Tela-circle-icon-theme.git) by me, @vinceliuice, and other contributors.


Content/Distribution/Linux/wicked-engine.desktop
```diff
@@ -0,0 +1,16 @@
+[Desktop Entry]
+Name=Wicked Engine
+Name[ja]=ウィケッドエンジン
+GenericName=Game engine
+GenericName[ja]=ゲームエンジン
+Comment=A 3D game engine focusing on modern rendering.
```
turanszkij, line 6: As I mentioned, it should be `Comment=3D engine focusing on modern rendering.`

turanszkij (changes requested):
The `Editor/Linux` folder for these seems more appropriate than `Content/Distribution/Linux`

Editor/Linux/Install.py
```diff
@@ -0,0 +1,75 @@
...
+os.system('clear')
+
+answer_installModuleDependencies = input("This installer has some Python dependencies, would you like to install them? [Yes/No]" + '\n')
+
+if answer_installModuleDependencies == "Yes":
+    print("Installing required modules.")
+    os.system('pip install pillow ttkbootstrap distro')
+elif answer_installModuleDependencies == "No": 
+    print("Skipping module dependencies!" + '\n' + "The installer might not run.")
+elif answer_installModuleDependencies == "Yes":
```
portaloffreedom, line 14: this code will never execute? Or am I wrong?

turanszkij: Looks like it

MolassesLover: What do you mean?

turanszkij: That the first line of the branch `if answer_installModuleDependencies == "Yes":` is the same as this elif line: `elif answer_installModuleDependencies == "Yes":` So the elif line would never run. You don't need to close the PR because of this, unless you don't want to fix this.

Editor/Linux/Install.py
```diff
@@ -1,32 +1,61 @@
...
 if answer_installModuleDependencies == "Yes":
     print("Installing required modules.")
     os.system('pip install pillow ttkbootstrap distro')
 elif answer_installModuleDependencies == "No": 
-    print("Skipping module dependencies!" + '\n' + "The installer might not run.")
-elif answer_installModuleDependencies == "Yes":
+    print("Skipping module dependencies!")
+    print("The installer might not run.")
+    time.sleep(1)
+elif answer_installModuleDependencies == "yes":
```
portaloffreedom, line 17: there is a better way to do this, that removes the code duplication:
```if answer_installModuleDependencies.lower() == "yes":
...
elif answer_installModuleDependencies.lower() == "no":
...
```

MolassesLover: Oh, thanks! I didn't know about `.lower`, I'll make sure to apply that.

MolassesLover:
I shifted a lot of the responsibilities to distributors and package maintainers in these changes, I think it ended up being for the better. I hope it can get merged, but again, I'm willing to make more alterations.

turanszkij (approved):

PR #403, opened 2022-03-14
(Linux) File dialog glob pattern fix

megumumpkin:
This one is for Flatpak applications, since the Flatpak uses portals for dialogs the requirements is a lot more stricter for file dialog list, seems like it does not tolerate empty string after the list (the space at the end of the list is detected as a file with an empty glob would also be accepted, which in this case is illegal)


turanszkij (approved):

turanszkij:
Nice find, thanks!

PR #404, opened 2022-03-15
[Linux] Install engine static library system wise

portaloffreedom:
This only installs the engine, not the editor or the tests (yet)


turanszkij (dismissed):

turanszkij:
Is this complete? It can be merged if so.

portaloffreedom:
Actually, it could get complicated, like setting the search path for the shaders, so maybe it's better to make it into a separate PR

turanszkij (approved):

PR #406, opened 2022-03-24
SDL Input Refactor

megumumpkin:
So here's a code optimization and slimming for the SDL input, changes are:
- Input are polled once and then the input data can be used everywhere. (SDL_Event events are collected in each SDL_PollInput call, and then the collected events can be read, after done the event collection are flushed away (vector to be reset to zero)).
- Keyboard input are also revamped, scancode and keycode conversion code for engine keyboard input are also made smaller and optimized, all engine text character input are now handled by SDL_TextInput which removes away my code for character conversion.
- Added SDL gamepad Vibration and LED support.
- Added a preliminary compatibility with touch input such as single touch and multigesture detection (everytime you create a touch event using a touch device it will send something to the backlog), don't know if it works or not.
- WIth the changes here also improvement in input timings, which the previous version somehow failed to capture input when using SDL_PollEvent or the like to read the input and had to be reread again.

While it seems like lots of code removed, there are nothing removed in the actual implemented functionalities before this input refactor is done, even there are now complete functionality for text input and gamepad, also preliminary support for touch.

Merge-ready.


MolassesLover:
Awesome work!

turanszkij (changes requested):
Sniper SDK should be a different build from the Ubuntu build because it installs more things, and it should be called maybe "Steam Deck" build because that would indicate its purpose better. Also, there are multiple of these build scripts: 
- one for pull request (build-pr.yml)
- one for nightly build (build-nightly.yml)
- one for master commits (build.yml)

Ideally each of these build scripts would build for every platform. Their difference is what's described under the `on:` keyword.

turanszkij (approved):

PR #409, opened 2022-04-01
Plain text to Markdown changes

MolassesLover:
## About
I started replacing a few plain text files, swapping them out with Markdown files. I don't think this is ready to be merged, since I will be making a few changes to other files. I just want some feedback, really. These changes should make things a bit easier to read. 

## Changes
- The `LICENSE.txt` file now uses the [unofficial](https://github.com/IQAndreas/markdown-licenses) but **valid** MIT [`LICENSE.md`](https://github.com/MolassesLover/WickedEngine/blob/markdown_changes/LICENSE.md) file.
- The `credits.txt` file has been swapped out with the [`CREDITS.md`](https://github.com/MolassesLover/WickedEngine/blob/markdown_changes/CREDITS.md) file.


turanszkij:
I think I'd rather keep them as txt, they are easier to open/edit (txt is much more common file type) and don't have much visual content.

PR #410, opened 2022-04-02
Tests - Update Gamepad Testing

megumumpkin:
So the gamepad section of the controllers has now been changed to this, which will help for devs to test controller inputs and to understand more about how gamepad input works within WickedEngine.

contra


turanszkij (approved):
It's perfect, works with Xbox 360 and PS5 controllers on windows. The LED only supports PS4 controller on windows, so sadly I couldn't test it now, but looking at the code it looks good.

PR #412, opened 2022-04-09
Wrote the `FUNDING.yml` file for GitHub

MolassesLover:
## About

This just adds some sponsorship buttons to the GitHub page.

In order for it to work you might need to enable the sponsorship button in the [settings](https://github.com/turanszkij/WickedEngine/settings)!

## Preview

Screenshot from 2022-04-09 20-05-33
Screenshot from 2022-04-09 20-06-16

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,1 @@
+patreon: wickedengine


turanszkij:
I don't want this right now.

MolassesLover:
> I don't want this right now.

Ah, alright, just figured it would be helpful.

PR #431, opened 2022-05-05
Visibility Buffer shading in CS

turanszkij:
Currently the visibility buffer is only used for post processes, but the main scene shading uses forward rendering. Here the forward rendering for OPAQUE pass will be changed to shade in compute shaders. This can:
- reduce CPU time, because scene objects don't need to be drawn again
- vertex shader doesn't run for again for lighting pass
- shading in compute could be more efficient than pixel shader (decoupled from triangle density)
- multisampling + reduced resolution could be used for prepass, will save on prepass GPU time
- won't be compatible with MSAA or VRS

TODO:
- better derivatives for aniso mip mapping
- decals: derivatives
- wind for static object is problematic: generates wrong barycentrics (also problem  for raytracing)
- material binning: support multiple materials
- impostors


turanszkij:
This is not on par with forward rendering yet, but merging this for the other improvements, and making the compute shading optional

PR #433, opened 2022-05-08
Minor cleanup

jdswebb:
Removed some redundant namespaces and removed unused window param from GraphicsDevice_Vulkan constructor


turanszkij:
wait, we do need the window in the constructor, because the Linux stuff must use it. I don't like it either

jdswebb:
Oh oops, nevermind this one then :)

jdswebb:
Seems the window parameter is getting removed from `SDL_Vulkan_GetInstanceExtensions` so it can go eventually. Maybe creating a dummy SDL window inside the constructor is also an option, it seems the window is not actually important to the function.

PR #435, opened 2022-05-08
QOI decoding Implementation

Sororfortuna:
So, since the engine is focused on contemporary rendering techniques and performance, I thought why not integrate QOI?
www.qoiformat.org/
It seems to be 3x to 4x quicker at decoding picture files compared to stb_image; this might be a big benefit for extremely high resolution imagery (up to 20,000*20,000 max resolution)


WickedEngine/wiResourceManager.cpp
```diff
@@ -5,6 +5,8 @@
 #include "wiUnorderedMap.h"
 
 #include "Utility/stb_image.h"
+#define QOI_IMPLEMENTATION // TODO: Move this to wherever turanszkij did the STB_IMAGE_IMPLEMENTATION, yes im looking at you o-o
```
turanszkij, line 8: it's in Utility/utility_common.cpp

WickedEngine/wiResourceManager.cpp
```diff
@@ -513,6 +516,90 @@ namespace wi
 					else assert(0); // failed to load DDS
 
 				}
+				else if (!ext.compare("QOI"))
```
turanszkij, line 519: if possible, it would be better to share most of the code of the stb_image branch, except the way you get the pointer to the pixels. I mean by combining the two branches into one.

turanszkij (changes requested):

WickedEngine/wiResourceManager.cpp
Sororfortuna: true, should I just compare it inside the last else then with ext.compare() then?

Sororfortuna: figured it out, am pushing requested changes in a min

turanszkij (approved):

WickedEngine/Utility/utility_common.cpp
```diff
@@ -15,7 +15,8 @@
 #define TINYDDSLOADER_IMPLEMENTATION
 #include "tinyddsloader.h"
 
-
+#define QOI_IMPLEMENTATION
+#include "Utility/qoi.h"
```
turanszkij, line 19: This is why the build is failing on Linux, the utility_common.cpp is already in the Utility folder, so the Utility folder doesn't need to be added to the include path. Please fix it.

turanszkij (changes requested):

WickedEngine/Utility/utility_common.cpp
Sororfortuna: added it to cmakelists and fixed that typo

PR #437, opened 2022-05-09
Fixed missing header in cmake

portaloffreedom:


turanszkij (approved):
I thought you only needed to include cpp files in the cmake

portaloffreedom:
The header files in cmake are only necessary for the install process (`sudo make install`)

PR #438, opened 2022-05-10
Visbuffer Optimizations

turanszkij:
Splitting visibility buffer shaders to reduce register usage. 
- resolve (async)
- binning (async)
- velocity (async)
- surface sampling (async)
- sky (async)
- lighting (gfx)

Register usage is improving, but the multiple passes outweigh the gains currently.


turanszkij:
- Successfully improved GPU utilization in lighting shader. 
- Surface, sky and decal rendering also runs now in async compute.
- Velocity is separate pass, improves GPU utilization in resolve, and only runs as optional.
- Surface shader occupancy is still bad, and wind computation is still bad per pixel, but now it is only done in surface shader.
- Surface shader can now write out proper top layer of clearcoated material, this improves screen space and raytraced reflections.
- MSAA not implemented yet for visibility compute shading
- Tessellation is still only half way supported in visibility compute shading (tessellation data not available in memory for pixel interpolation, but depth is written in prepass, this causes mismatch)

PR #442, opened 2022-05-12
Changed the Wicked Engine launcher shebang

MolassesLover:
The Wicked Engine launcher script on GNU/Linux has been altered to use `sh` rather than `bash`, making it *almost* universal.


portaloffreedom (approved):

portaloffreedom (changes requested):
Would want also the `wicked-engine.template.sh` file to have the same change

MolassesLover:
> Would want also the `wicked-engine.template.sh` file to have the same change

There you go, changed it to use `sh`!

portaloffreedom (approved):

PR #446, opened 2022-05-15
Fixes the frame delay issue with rigid body visuals

eddieataberk:
Sent before and after videos to Discord.


WickedEngine/wiScene.cpp
```diff
@@ -1577,6 +1577,7 @@ namespace wi::scene
 		}
 
 		wi::jobsystem::context ctx;
+		wi::physics::RunPhysicsUpdateSystem(ctx, *this, dt);
```
turanszkij, line 1580: Please double check that this is necessary. I heard about motion state could cause lag, but this line could have larger implications (animation system to physics update for example).

PR #462, opened 2022-06-13
A few more Visual Studio Code/Codium changes

MolassesLover:
- Unused files such as images should be hidden in the explorer.
- The CMake build tasks are a bit cleaner now and use `$(nproc)` rather than `nproc`.


turanszkij (approved):

.vscode/settings.json
```diff
@@ -0,0 +1,18 @@
+{
+    "files.exclude": {
```
turanszkij, line 2: On second thought, I would prefer if the files were not hidden. They can be made hidden on your side only if you want. The user configuration files shouldn't be in the source control imo.

turanszkij (changes requested):

.vscode/settings.json
MolassesLover: Should I add the file to the `.gitignore`?

turanszkij: You can add it

turanszkij (approved):

PR #464, opened 2022-06-14
Added multiple jobs to the GitHub CI workflows 

MolassesLover:
## Changes

- Added `-j $(nproc)` to the end of all `make` command calls.

## Why?

The `$(nproc)` variable contains the number of processing units on your system.
```zsh
➜  ~ echo $(nproc)
12
```
In my case, the `make -j$(nproc)` command would create 12 jobs, drastically speeding up the build time. Though, I have absolutely no clue how many units GitHub CI has so this might be useless... (❛‿❛✿̶̥̥)


turanszkij (approved):
The linux build got faster with this

PR #468, opened 2022-06-22
fix - endless loop if some lights do not have shadows.

plemsoft:


WickedEngine/wiRenderer.cpp
```diff
@@ -3117,6 +3117,10 @@ void UpdatePerFrameData(
 					iterative_scaling *= 0.5f;
 				}
 			}
+			else
+			{
+				iterative_scaling = 0.0; //PE: fix - endless loop if some lights do not have shadows.
```
turanszkij, line 3122: Thanks, I think using simply `break;` would be better here

turanszkij (changes requested):

turanszkij (approved):

turanszkij:
Thanks for the fix

Issue #450, opened 2022-05-28
Bugs?


cailuming:
![image](https://user-images.githubusercontent.com/19427259/170810203-59a48706-497a-4cca-9f63-191049ec8c9f.png)
black blocks when I turn on the DOF

tictech1970:
The same bug occurs to me with DX12 but not with Vulkan.

cailuming:
another one 
![image](https://user-images.githubusercontent.com/19427259/170816796-37741995-538e-4e28-a896-39c12d5e24d1.png)
when I select any one of the items, it will crash.Test scene file is a little bit large , sorry for not being able to offer.

cailuming:
@tictech1970 I have not tested it on Vulkan yet

turanszkij:
Will take a look at the issues when I have some time.

cailuming:
Thanks，Maybe I should make it reproducible

turanszkij:
By the way, what type of GPU do you have? Nvidia? AMD? I have tried with Nvidia. The problem could also be not only the depth of field, but maybe how it interacts with an other effect. A RenderDoc capture of the issue could maybe help too.

cailuming:
My GPU is  Nvidia GF 1060 , and I will collect the RenderDoc info when I reproduce this bug in these days.

cailuming:
I failed to reproduce the black block but another new one arise. 
Firstly run the windowEditor.then just click the terrain button, and next turn on DDGI or surfel GI or SSR ,  crash will appear.

_Program: ...ne\CustomBuild\Editor\Debug\WickedEngineEditor.exe
F:\WickedEngine\WickedEngine\wiGraphicsDevice_DX12.cpp
Line: 3060_

And I found some errors always complain whatever I do.
onecore\com\combase\dcomrem\giptbl.cxx(1786)\combase.dll!00007FFAC05150C7: (caller: 00007FFAA1CD9759) ReturnHr(86) tid(4204) 80070057 参数错误。
onecore\com\combase\dcomrem\giptbl.cxx(1786)\combase.dll!00007FFAC05150C7: (caller: 00007FFAA1CD9759) ReturnHr(87) tid(4204) 80070057 参数错误。
onecore\com\combase\dcomrem\giptbl.cxx(1786)\combase.dll!00007FFAC05150C7: (caller: 00007FFAA1CD9759) ReturnHr(88) tid(4204) 80070057 参数错误。
[...]

PR #480, opened 2022-06-30
DX12 SV_PrimitiveID workaround

plemsoft:
Was not able to find the real problem with the black dots issue, but this only call the workaround 1 time at startup and only DX12.


WickedEngine/wiRenderer.cpp
```diff
@@ -6030,7 +6060,11 @@ void DrawDebugWorld(
 
 	if (GetRaytraceDebugBVHVisualizerEnabled())
 	{
-		RayTraceSceneBVH(scene, cmd);
+		//PE: Check if debug BVH is possible. or it will crash.
```
turanszkij, line 6063: Path tracing also needs the debug bvh visualizer

WickedEngine/wiGraphicsDevice.h
```diff
@@ -44,6 +44,12 @@ namespace wi::graphics
 
 		QUEUE_COUNT,
 	};
+	enum BACKEND_TYPE
```
turanszkij, line 47: I would prefer to use dynamic_cast instead and not have this enum.

WickedEngine/wiRenderer.cpp
```diff
@@ -2385,6 +2385,36 @@ ForwardEntityMaskCB ForwardEntityCullingCPU(const Visibility& vis, const AABB& b
...
+			static RenderPass renderpass_clear;
+			if (!renderpass_clear.IsValid())
+			{
+				RenderPassDesc renderpassdesc;
+				renderpassdesc.flags = RenderPassDesc::Flags::EMPTY;
+				device->CreateRenderPass(&renderpassdesc, &renderpass_clear);
+			}
+			device->RenderPassBegin(&renderpass_clear, cmd);
+			device->BindPipelineState(pso, cmd);
+			device->DrawIndexedInstanced(0, 0, 0, 0, 0, cmd); //PE: Just need predraw(cmd);
```
turanszkij, line 2411: Do you know which part of the predraw() is needed? Does it need the PSO, or maybe just a root signature or root parameter?

WickedEngine/wiApplication.cpp
```diff
@@ -101,6 +101,15 @@ namespace wi
 			return;
 		}
 
+		static bool startup_workaround = false;
```
turanszkij, line 104: without having the enum for device type, here you could check easily if device is dx12, for example by dynamic cast, or this bool could be set to true when the dx12 device gets created.

WickedEngine/wiRenderer.cpp
```diff
@@ -2385,6 +2385,36 @@ ForwardEntityMaskCB ForwardEntityCullingCPU(const Visibility& vis, const AABB& b
 	return cb;
 }
 
+void Workaround(const int bug , CommandList cmd)
+{
+	if (device->GetBackend() == BACKEND_DX12)
+	{
+		if (bug == 1)
+		{
+			//PE: Strange DX12 bug, we must change the pso/pipeline state, just one time.
```
turanszkij, line 2394: It would be good to have a `device->EventBegin()` and `device->EventEnd()` surrounding workaround gfx commands, so that workaround is obvious when gfx debugging.

turanszkij (changes requested):

WickedEngine/wiRenderer.cpp
plemsoft: Path tracing debug have its own call , so it works fine :)

WickedEngine/wiRenderer.cpp
plemsoft: Yes the pso is needed and it must be RENDERPASS_VOXELIZE , i tried a lot others they dont fix it, perhaps its the DSSTYPE_DEPTHDISABLED that do it ? but i tried other debug PSO with DSSTYPE_DEPTHDISABLED but they did not fix it, so it must be a combination in RENDERPASS_VOXELIZE ?. Anyway its needed.

WickedEngine/wiGraphicsDevice.h
plemsoft: Not sure , you would have to use the enum anyway , or what do you mean ?

WickedEngine/wiApplication.cpp
plemsoft: "without having the enum" not sure what you mean , let me know.

turanszkij: For example: 
```cpp
if(dynamic_cast<GraphicsDevice_DX12*>(device) != nullptr)
{
  workaround();
}
```

Or without dynamic cast, in wiapplication.cpp, when dx12 device gets created:
```
if(dx12)
{
  new GraphicsDevice_DX12(...)
  startup_workaround = false; // so the startup workaround will get executed
}
```

WickedEngine/wiGraphicsDevice.h
plemsoft: Ahh know what you mean now, fixed :)

WickedEngine/wiApplication.cpp
plemsoft: Fixed.

WickedEngine/wiApplication.cpp
```diff
@@ -101,6 +101,18 @@ namespace wi
...
+		if (!startup_workaround)
+		{
+			startup_workaround = true;
+			if (dynamic_cast<GraphicsDevice_DX12*>(graphicsDevice.get()))
```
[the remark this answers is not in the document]
plemsoft: Fixed :)

turanszkij (approved):

turanszkij:
Good job on finding a workaround for this, thanks. It still looks like bug in driver, more info here: https://github.com/turanszkij/WickedEngine/issues/450

PR #488, opened 2022-07-17
Some changes to ImGui docking sample

MolassesLover:
I was unable to build the example before this, and I don't think anyone else tried to on GNU/Linux. I don't think these changes are very intrusive, since I used `#ifdef _WIN32` to make the changes.

Something that does confuse me is why this line didn't work:
```c++
ImportModel_GLTF(wiscene_file_items[current_scene], scene);
```
Disabling it on Unix-based systems changed nothing as far as I'm aware. It would be nice if @plemsoft could take a look at it, since it completely broke compilation on GNU/Linux.

Aside from that, when compiling outside of the `WickedEngine/` directory, the example wouldn't create a window, only run the Lua interpreter...?

```
Lua 5.4.4  Copyright (C) 1994-2022 Lua.org, PUC-Rio
>
```

This wasn't the exact output but you get the idea, I hope.

Lastly, I hope merging from my `master` branch isn't an issue. I don't think I changed anything on there, so it should be fine.


turanszkij:
The ImportModel_GLTF doesn't work, because the ModelImporter_GLTF.cpp is not compiled on Linux (it is not in the Example_ImGui_Docking/CmakeLists.txt). This file is in Editor/ModelImporter_GLTF.cpp. Do you want to add it?

MolassesLover:
There you go, that should be ready to merge.

turanszkij (approved):

PR #489, opened 2022-07-18
Various updates

amerkoleci:
- Update DirectX headers to agility SDK 1.606.3
- Update D3D12MemAlloc to latest version.
- Update Vulkan headers to 1.3.221
- Update volk
- Update VMA


turanszkij:
Thanks, I will need some time to test these

amerkoleci:
> Thanks, I will need some time to test these

I already tested both D3D12 and Vulkan samples, editor etc

turanszkij:
DDGI and surfel GI are broken, so I will need to investigate them

turanszkij:
directional light in vulkan, in Sponza is broken, too

amerkoleci:
> directional light shadow maps in vulkan, in Sponza is broken, too

OMG, why that? DXC update?

turanszkij:
It seems both issues are caused by new DXC, because if I replace the dxcompiler.dll and dxil.dll in your branch with old ones, and rebuild shaders, problem goes away.

amerkoleci:
> It seems both issues are caused by new DXC, because if I replace the dxcompiler.dll and dxil.dll in your branch with old ones, and rebuild shaders, problem goes away. So could you add the old dxcompiler.dll and dxil.dll back?

OFC yes, strange tho, reverted back

amerkoleci:
Would be cool to understand what shader feature is causing recent DXC issue, so we can report to their github repo, as the version I've grabbed is marked as stable (https://github.com/microsoft/DirectXShaderCompiler/releases/tag/v1.7.2207)

turanszkij (approved):

turanszkij:
Thanks, it looks good now.

PR #491, opened 2022-07-19
CMake template project for Linux

MolassesLover:
Based on my Wicked Engine demo project, forked from Matteo's original demo.

No new dependencies or anything, this template uses the bare minimum.

If you find any references to my demo project, please let me know so I can remove them. I wrote this at midnight and I'm pretty tired, so I'm sure there's something I missed.


MolassesLover:
Should I add this to the root `CMakeLists.txt` or would that be too intrusive?
```cmake
option(WICKED_LINUX_TEMPLATE "Build WickedEngine Linux template" OFF)
if (WICKED_LINUX_TEMPLATE)
    add_subdirectory(Template_Linux)
endif()
```
Also, if yes, then should it default to **on** or **off**?

turanszkij:
The App.h and App.cpp is more like the code from Tests. The Template should be even more minimal, like on windows, it only displays a black screen and the Wicked Engine version text. It only has a single main.cpp file which creates the window and the engine's Application object. It should have no custom classes or logic (the Tests show that better already).

turanszkij:
> Should I add this to the root `CMakeLists.txt` or would that be too intrusive?
> 
> ```cmake
> option(WICKED_LINUX_TEMPLATE "Build WickedEngine Linux template" OFF)
> if (WICKED_LINUX_TEMPLATE)
>     add_subdirectory(Template_Linux)
> endif()
> ```
> 
> Also, if yes, then should it default to **on** or **off**?

I'd like to hear @portaloffreedom 's opinion on this.

.github/workflows/build-nightly.yml
```diff
@@ -81,10 +81,9 @@ jobs:
 
     - name: Build
       run: |
-        mkdir build
-        cd build
-        cmake .. -DCMAKE_BUILD_TYPE=Release
-        make -j $(nproc)
+        cmake -B build -DCMAKE_BUILD_TYPE=Release -D WICKED_LINUX_TEMPLATE=ON
```
turanszkij, line 84: Let's not change these commands, just specify that the template build is ON by default, like the other projects.

turanszkij (changes requested):

.github/workflows/build-nightly.yml
turanszkij: Yes, I mean that it's set to ON in the Cmake, and don't modify the github build scripts

MolassesLover: Why?

turanszkij: Because all the other projects work like that, and I prefer to not modify github builder if not necessary

MolassesLover: Don't worry, CMake's `-B` flag is just `mkdir` and `cmake --build` automatically finds what build system you're using to build. So if you build with **Unix Makefiles** it'll run `make`. If you use something like **Ninja** it will use that instead.

So this makes the CI more future proof, and can prevent errors.

If I change `CMakeSettings.txt` to build the template, that will definitely upset people since it's an unused build that increases global build time. For me it's not an issue, but someone with lower-end hardware would be affected and might not even know why.

What I did wasn't an actual change, just an abstraction for `mkdir` and `make`, which allows me to speed up build times for everyone and toggle settings, if we ever have to do that in the future. 

This won't affect the binaries or anything, I try my best to avoid intrusive changes to the CI.

turanszkij: You can use cmake however you want in your projects, but the minimal cmake steps that I used and trust are highlighted in the readme and have been used in the github builder with success. All the projects so far work the same way, so I would like this new project follow this too. 

"that will definitely upset people since it's an unused build that increases global build time" - but it can be disabled.

.github/workflows/build-nightly.yml
```diff
@@ -83,8 +83,8 @@ jobs:
       run: |
         mkdir build
         cd build
-        cmake .. -DCMAKE_BUILD_TYPE=Release
-        make -j $(nproc)
+        cmake .. -DCMAKE_BUILD_TYPE=Release -D WICKED_LINUX_TEMPLATE=ON
```
turanszkij, line 86: There is still the `-D WICKED_LINUX_TEMPLATE=ON`

CMakeLists.txt
```diff
@@ -38,3 +38,8 @@ if (WICKED_TESTS)
     add_subdirectory(Example_ImGui)
     add_subdirectory(Example_ImGui_Docking)
 endif()
+
+option(WICKED_LINUX_TEMPLATE "Build WickedEngine Linux template" OFF)
```
turanszkij, line 42: This should be ON as all the other samples

turanszkij (changes requested):

turanszkij (approved):

Issue #51, opened 2019-11-20
GUI improvements


turanszkij:
The existing GUI system should be improved for better Editor usability, and better features:

-Tree view: for displaying scene graph entities, like "outliner" in Blender
-Timeline: to create animations
-Window tabs, window docking: Reduce multiple windows cluttering up space in the Editor
-Controller support: The Editor should be operatable with a controller at least to some minimal extent
-Widget images and animations: GUI should be used in game as well, but this must be nicer than just the current colored boxes approach.

turanszkij:
Scroll bar should be abstracted so that it can enabled for other widgets too, not just combo box or tree list.

turanszkij:
Really need a scrollbar now for the windows or tabs...

MolassesLover:
Please, I want this so bad

Sororfortuna:
I think having a 0.0 to 1.0 rounding property for the UI elements would be a great addition; currently I'm relying on images for creating circles/rounded corners, which doesn't have a lot of consistency on different monitors and is probably not very resource efficient

turanszkij:
Reminder: add ResizeLayout() callback support for widgets.

turanszkij:
Textbox supports caret positioning: https://github.com/turanszkij/WickedEngine/commit/90fbbba9bada3f7c049c8d17b545a05c51ec7324

turanszkij:
Improvements will be made continually when needed, closing.

PR #495, opened 2022-07-31
GUI ResizeLayout

turanszkij:


turanszkij:
ResizeLayout is override, not callback #51

PR #499, opened 2022-08-01
Lua Live Update

megumumpkin:
Here's my design on Wicked Engine's Lua scripting live update system.
All scripts now support live reloading, added new mode for dofile to load scripts which are using the new script file path system. 
Variable persistence among reloads are explicitly declared by using D table.

### DIFF
--- a/Content/scripts/live_reload_demo.lua
+++ b/Content/scripts/live_reload_demo.lua
@@ -1,6 +1,7 @@
 -- Lua script live reload demo
 -- This is a script which showcases the new live-reload behavior and also the new path system for scripts
--- Live reload is done by loading the script by file, editing the script, and focus again on the editor window to see the changes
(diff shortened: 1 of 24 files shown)


turanszkij (changes requested):
The live_reload_demo.lua doesn't work in the current state for me. When I open with the Editor: 
image
When I open with dofile command with `true` second param: 
image

megumumpkin (commented):
Done

megumumpkin (commented):
Done

megumumpkin:
Can I add an option in the editor to run script between two path modes?

turanszkij (changes requested):
As discussed in private, at the current state it is too complicated to use, requires mental effort from developer to program scripts for correct live reload usage. The script path behaviour can be useful as an improvement that is separate from live reload.

PR #501, opened 2022-08-07
New Lua Scripting

megumumpkin:
Here's the new Lua scripting which will replace the old one.
New scripts now use SCRIPT_DIR() to get script's current path unlike before which is automatic.
Also new functionality for users to have more control over their script (Killing exactly 1 coroutine, or 1 instance, or instances related by 1 file), also for script reloads there are new functionality to detect if script has been run once before.

### DIFF
--- a/WickedEngine/wiLua.cpp
+++ b/WickedEngine/wiLua.cpp
@@ -42,7 +42,32 @@ namespace wi::lua
 		}
 	};
 	LuaInternal luainternal;
-	std::string script_path;
+
+	uint32_t Internal_GenScriptPID(){
+		static std::atomic<uint32_t> scriptpid_next{ 0 + 1 };
+		return scriptpid_next.fetch_add(1);
+	}
+
+	uint32_t Internal_EncapsulateScript(std::string& script, std::string filename = "", uint32_t PID = 0){
+		static const std::string persistent_inject = R"(
+			local runProcess = function(func) 
+				success, co = Internal_runProcess(SCRIPT_FILE, SCRIPT_PID, func)
+				return success, co
+			end
+		)";
+
+		if(PID == 0){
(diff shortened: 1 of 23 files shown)


turanszkij (changes requested):
- Sample scripts do not work: 
image

- I'd like if the `SCRIPT_DIR()` function was renamed to `script_dir()` instead, and same for `script_file()`, `script_pid()`. `Script_Initialized()` to `script_initialized()`

WickedEngine/wiLua.cpp
```diff
@@ -42,25 +42,64 @@ namespace wi::lua
...
+				_ENV.PROCESSES_DATA[SCRIPT_PID()]._INITIALIZED = _ENV.PROCESSES_DATA[SCRIPT_PID()]._INITIALIZED + 1
+			end
+		)";
+
+		if(PID == 0){
+			PID = Internal_GenScriptPID();
+		}
+
+		std::string dynamic_inject = "local function SCRIPT_FILE() return \""+filename+"\" end\n";
+		dynamic_inject += "local function SCRIPT_PID() return \""+std::to_string(PID)+"\" end\n";
```
turanszkij, line 70: Why after `return there is a `\` character? I think this causes error on Windows, I don't think this is needed.

turanszkij: Oh, it's for the `"`, I see

turanszkij:
Oh, I think the error is the path itself, because it contains `\` which lua recognizes as escape char..

turanszkij (approved):

turanszkij:
Done fixes and some changes, let me know if you find any issues or ok to merge.

megumumpkin:
Ok to merge, no issues

PR #504, opened 2022-08-10
Replace Arial with "Liberation Sans"

brakhane:
The Arial license does not allow redistribution, so this commit replaces
it with the metrically compatible "Liberation Sans" font.

An alternative would be to use eg. the modern Noto Sans font, which looks nicer IMO.
But that change would introduce incompatibility since the font has different metrics.

Let me know if you are willing to accept the different metrics and want me to use Noto Sans
(or any other font) instead.


turanszkij (approved):
I think it looks acceptable, and your implementation is good too. We can replace the default arial font with this.

PR #507, opened 2022-08-13
replaced default unordered_map to fix linux crash

SeleDreams:
I've set the default to 2 as well since ska::unordered_map seemed to have the most stable performances on linux while not crashing (like ska::flat_hash_map)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,7 +20,6 @@ elseif(UNIX)
     add_compile_definitions(SDL2=1)
     set(DXC_TARGET "dxc")
 endif()
-
 add_subdirectory(WickedEngine)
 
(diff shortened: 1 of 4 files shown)


turanszkij:
It is ok to add these unordered_map implementations, but please don't modify the default one in this header file. Instead, only modify it for the Linux build in the Cmake. The `ska::flat_hash_map` is faster than the `ska::unordered_map` that you enabled here.
perf test with `ska::flat_hash_map` (original):
image
perf test with `ska::unordered_map` (this PR):
image

turanszkij (changes requested):
[See my previous comment for requested changes]

turanszkij:
fixed in #518

PR #510, opened 2022-08-15
Add Raw Input and Virtual Input Mode for Keyboard

megumumpkin:
This is just a proposal, here's the PR just to make it compile with PR.
And it is an optional feature.

### DIFF
--- a/WickedEngine/wiSDLInput.cpp
+++ b/WickedEngine/wiSDLInput.cpp
@@ -1,4 +1,6 @@
 #include "wiSDLInput.h"
+#include "wiInput.h"
+#include <SDL_keyboard.h>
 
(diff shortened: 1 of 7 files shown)


WickedEngine/wiSDLInput.cpp
```diff
@@ -1,4 +1,6 @@
 #include "wiSDLInput.h"
+#include "wiInput.h"
+#include <SDL_keyboard.h>
```
portaloffreedom, line 3: you need to move this include inside the `#ifdef SDL2`

portaloffreedom: to fix the windows compilation

PR #511, opened 2022-08-15
Add ComponentLibrary and Component Interfaces

megumumpkin:
Here is the new addition to the ECS system, making it easier for users to serialize custom components by having it be done automatically.

This functionality is also used in the demo game, this makes it far easier to manage components in every aspect for me.
So users does not necessarily need its own scene entity processors for new types of component, especially when serializing into one scene (no need to keep two separate scenes).


WickedEngine/wiECS.h
```diff
@@ -346,6 +404,48 @@ namespace wi::ecs
 		// Disallow this to be copied by mistake
 		ComponentManager(const ComponentManager&) = delete;
 	};
+
+	// This is the class to store all ComponentManagers,
+	// this is useful for bulk operation of all attached components within an entity
+	class ComponentLibrary{
+	public:
+		uint64_t libraryVersion;
+		wi::vector<ComponentManager_Interface*> componentManagers;
```
turanszkij, line 413: Should be `wi::vector<std::unique_ptr<ComponentManager_Interface>> componentManagers;` instead, to avoid memory leak on destroy.

WickedEngine/wiECS.h
```diff
@@ -346,6 +404,48 @@ namespace wi::ecs
...
+		//Create a Component Library and set up its version
+		ComponentLibrary(uint64_t iLibraryVersion = 0){ libraryVersion = iLibraryVersion; }
+
+		// Set library version, making sure different changes 
+		// between your component library are able to be serialized properly
+		void SetLibraryVersion(uint64_t& iLibraryVersion){ libraryVersion = iLibraryVersion; }
+
+		// Create an instance of ComponentManager of a certain data type
+		// Once added, cannot be removed!
+		template<typename T> inline ComponentManager<T>& Add(){
```
turanszkij, line 424: Maybe it should be called `Register()` instead of Add, I think it would signal it better what is happening.

WickedEngine/wiECS.h
```diff
@@ -161,6 +191,34 @@ namespace wi::ecs
 			}
 		}
 
+		//Read/write one single component onto an archive, make sure entity are serialized first
+		inline void Component_Serialize(Entity entity, wi::Archive& archive, EntitySerializer& seri)
```
turanszkij, line 195: Instead of this, it would be better to have a function on the Library, called Entity_Serialize(). In that case, the `archive.IsReadMode()` doesn't need to be checked for every component, but instead:
```cpp
if(archive.IsReadMode())
{
  for(all component mangers)
  {
    read...
  }
}
else
{
  for(all component mangers)
  {
    write...
  }
}
```

turanszkij (changes requested):
I will also need more time to test it after changes are addressed..

WickedEngine/wiECS.h
megumumpkin: For this, the thing is that I can't get the specific type of the component since I've made the componentManagers into an interface.

megumumpkin: Can't directly access the component data so what I did here is to split Component_Serialize into two, and then serializing it like above:
```
if(archive.IsReadMode())
{
  for(all component mangers)
  {
    Component_Serialize_R...
  }
}
else
{
  for(all component mangers)
  {
    Component_Serialize_W...
  }
}
```

turanszkij: Oh, I didn't see that it can't be done. In this case I like the previous version better, sorry.

megumumpkin: @turanszkij also how about the archive version inclusion within the ComponentLibrary?

turanszkij: Looks good, but I'll want to test it

megumumpkin: Ok, done.

WickedEngine/wiECS.h
```diff
@@ -407,12 +407,11 @@ namespace wi::ecs
...
 	// This is the class to store all ComponentManagers,
 	// this is useful for bulk operation of all attached components within an entity
-	class ComponentLibrary{
+	class ComponentLibrary
+	{
 	public:
-		uint64_t libraryVersion;
+		uint64_t libraryVersion = 0;
 		wi::vector<std::unique_ptr<ComponentManager_Interface>> componentManagers;
-		wi::vector<uint64_t> componentVersions;
```
turanszkij: These were not used for anything, so I removed them. Am I missing something?

WickedEngine/wiECS.h
```diff
@@ -423,42 +422,48 @@ namespace wi::ecs
...
+			for(auto& componentManager : componentManagers)
+			{
 				if(archive.IsReadMode())
 				{
 					uint64_t archiveVersion;
 					uint64_t componentVersion;
 					archive >> archiveVersion;
 					archive >> componentVersion;
-					if((archive.GetVersion() >= archiveVersion) && (componentVersion >= libraryVersion)){
+					if((archive.GetVersion() >= archiveVersion) && (libraryVersion >= componentVersion))
```
turanszkij, line 465: I switched the order of checks between componentVersion and libraryVersion. Otherwise base engine that is libraryVersion = 0 would try to load custom components that were created with a libraryVersion > 0.

turanszkij (dismissed):
I think now it is good, let me know if you find a problem with my latest changes

megumumpkin:
Done updating, changing the serialization again so now the components can be placed on an arbitrary position.

WickedEngine/wiECS.h
```diff
@@ -346,6 +406,103 @@ namespace wi::ecs
...
+
+		inline void Serialize(wi::Archive& archive, EntitySerializer& seri){
+			if(archive.IsReadMode()){
+				bool has_next = false;
+				do{
+					archive >> has_next;
+					if(has_next){
+						uint64_t component_type;
+						archive >> component_type;
+						wi::vector<uint8_t> datablock_raw;
```
turanszkij, line 454: This is not good, we are making extra copy of all component data

WickedEngine/wiECS.h
```diff
@@ -346,6 +406,103 @@ namespace wi::ecs
...
+	// this is useful for bulk operation of all attached components within an entity
+	class ComponentLibrary
+	{
+	public:
+		wi::unordered_map<uint32_t, std::unique_ptr<ComponentManager_Interface>> componentManagers;
+
+		// Create an instance of ComponentManager of a certain data type
+		// Once added, cannot be removed!
+		template<typename T> inline ComponentManager<T>& Register(std::string name){
+			uint64_t name_hashed = wi::helper::string_hash(name.c_str());
```
turanszkij, line 420: Thinking about the string hash, I think we might run into hash collision problem, so probably it's better to use strings. That way on hash collision strings can still be compared.

WickedEngine/wiScene.cpp
```diff
@@ -2069,67 +2045,17 @@ namespace wi::scene
...
-		aabb_decals.Merge(other.aabb_decals);
-		animations.Merge(other.animations);
-		animation_datas.Merge(other.animation_datas);
-		emitters.Merge(other.emitters);
-		hairs.Merge(other.hairs);
-		weathers.Merge(other.weathers);
-		sounds.Merge(other.sounds);
-		inverse_kinematics.Merge(other.inverse_kinematics);
-		springs.Merge(other.springs);
+		for(auto& componentManager_kval : componentLibrary.componentManagers){
```
turanszkij, line 2048: Probably good idea to make the library compatible with `for(auto& componenManager : library) {}`

WickedEngine/wiECS.h
```diff
@@ -346,6 +406,103 @@ namespace wi::ecs
...
+			componentManagers[name_hashed] = std::make_unique<ComponentManager<T>>();
+			return static_cast<ComponentManager<T>&>(*componentManagers[name_hashed]);
+		}
+
+
+		// Bulk operations
+		// Perform deep copy of all the contents of "other" into this
+		inline void Copy(const ComponentLibrary& other){
```
turanszkij, line 428: Can you make `{` start on new line because that's what most of the other engine code follows?

turanszkij (changes requested):

turanszkij (approved):

PR #518, opened 2022-08-20
Fixing broken hashmaps on linux

### LINKED ISSUE #414: [bug] Terrain generator drain all memory (Debug, Linux)
When i try create terrain in Editor ( Debug, Linux ), its drain all memory, than crash ( about 20Gb)
when it crash :  
изображение
when it close after crash :    
изображение

portaloffreedom:
Fixes #414

The issue is caused by Linux (and possibly other platforms) implementing the hash function as an identify function for data smaller than 64bit (including `int`).

Fixing only the TerrainGenerator is possible, but the issue could be generated again elsewhere and it would be quite difficult to identify and debug.

Therefore the proposed solution is to replace the `wi::unordered_map` on Linux with another implementation which is still quite fast and does account for the issue of the identify hash function.
The library is https://github.com/martinus/robin-hood-hashing (MIT License)

robin_hood.h:768
```
// return mixed of that, to be safe against identity hash
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,7 +40,7 @@ if (WICKED_TESTS)
     add_subdirectory(Tests)
 endif()
 
-if (WICKED_TESTS)
+if (WICKED_IMGUI_EXAMPLE)
     add_subdirectory(Example_ImGui)
     add_subdirectory(Example_ImGui_Docking)
 endif()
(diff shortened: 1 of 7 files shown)


turanszkij (dismissed):

turanszkij (approved):

turanszkij (approved):

portaloffreedom:
Thanks to @stengun for the help debugging this issue :D

PR #521, opened 2022-08-21
Script Component

turanszkij:
Allows binding `ScriptComponent` to entity in the scene. `ScriptComponent` will always execute itself as a single lua script on `Scene::Update()`. The `ScriptComponent`'s lua script is special and it will gain access to a `GetEntity()` lua function which it can use to reference additional components on itself. An example model was added to Content/models called script_component.wiscene that binds a script component to a cube mesh, and it rotates it a bit every frame.


megumumpkin (approved):
Good to merge!

PR #522, opened 2022-08-21
Compiler warnings fixes

portaloffreedom:
I've enabled the GCC compiler warnings to check for possible bugs. Found a couple for sure in the constructors, and a lot of minor changes.

Especially for the removal of the unused variables, I expect most changes are going to be reverted. I've removed them anyway because I'm challenging others to check if some dead code remained after some changes and actually could be removed, or if it's code that will be used later but for now it's completely useless.


portaloffreedom:
I've separated the commits in categories, to make it easier to analyze what is going on

Editor/meshoptimizer/clusterizer.cpp
```diff
@@ -1,10 +1,10 @@
 // This file is part of meshoptimizer library; see meshoptimizer.h for version/license details
 #include "meshoptimizer.h"
 
-#include <assert.h>
-#include <float.h>
-#include <math.h>
-#include <string.h>
+#include <cassert>
```
turanszkij, line 4: This is meshoptimizer library, no use to change these because changes will be lost when library is updated. Why is this a warning anyway?

Editor/xatlas.cpp
```diff
@@ -37,14 +37,14 @@ Copyright (c) 2012 Brandon Pelfrey
 #include <condition_variable>
 #include <mutex>
 #include <thread>
-#include <assert.h>
```
turanszkij, line 40: xatlas library...

WickedEngine/wiGraphicsDevice_Vulkan.cpp
```diff
@@ -5761,6 +5765,8 @@ using namespace vulkan_internal;
 				view_desc.format = VK_FORMAT_D32_SFLOAT_S8_UINT;
 				view_desc.subresourceRange.aspectMask = VK_IMAGE_ASPECT_DEPTH_BIT;
 				break;
+			default:
+				assert(0);
```
turanszkij, line 5769: why not break?

WickedEngine/wiGraphicsDevice_Vulkan.cpp
```diff
@@ -5887,6 +5893,8 @@ using namespace vulkan_internal;
 				view_desc.format = VK_FORMAT_D32_SFLOAT_S8_UINT;
 				view_desc.subresourceRange.aspectMask |= VK_IMAGE_ASPECT_STENCIL_BIT;
 				break;
+			default:
+				assert(0);
```
turanszkij, line 5897: why not break?

WickedEngine/wiGraphicsDevice_Vulkan.cpp
```diff
@@ -2004,6 +1996,20 @@ using namespace vulkan_internal;
 					}
 					break;
 
+					case VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER:
```
turanszkij, line 1999: why not default?

WickedEngine/wiRenderPath3D.cpp
```diff
@@ -117,7 +117,7 @@ void RenderPath3D::ResizeBuffers()
 		device->CreateTexture(&desc, nullptr, &rtSceneCopy_tmp);
 		device->SetName(&rtSceneCopy_tmp, "rtSceneCopy_tmp");
 
-		for (uint32_t i = 0; i < rtSceneCopy.GetDesc().mip_levels; ++i)
```
turanszkij, line 120: mip_levels is uint32_t I don't understand this change

WickedEngine/wiRenderPath3D.cpp
```diff
@@ -271,7 +271,7 @@ void RenderPath3D::ResizeBuffers()
 
 		for (uint32_t i = 0; i < desc.mip_levels; ++i)
 		{
-			int subresource_index;
+			uint32_t subresource_index;
```
turanszkij, line 274: subresource index return is int, why was this changed?

WickedEngine/wiSpriteAnim_BindLua.cpp
```diff
@@ -342,24 +342,23 @@ namespace wi::lua
...
 	Luna<DrawRectAnim_BindLua>::PropertyType DrawRectAnim_BindLua::properties[] = {
-		{ NULL, NULL }
+		{ nullptr, nullptr }
 	};
 
-	DrawRectAnim_BindLua::DrawRectAnim_BindLua(const wi::Sprite::Anim::DrawRectAnim& data) :anim(anim)
-	{
-	}
+	DrawRectAnim_BindLua::DrawRectAnim_BindLua(const wi::Sprite::Anim::DrawRectAnim& data)
+		:anim(data)
```
turanszkij, line 352: This one is a nice find

turanszkij (changes requested):
I think a lot of these are not really helpful or correct.

WickedEngine/wiSpriteAnim_BindLua.cpp
turanszkij: This one specifically I will fix separately.

WickedEngine/wiPrimitive.h
```diff
@@ -123,7 +123,12 @@ namespace wi::primitive
 		float TMax = std::numeric_limits<float>::max();
 		XMFLOAT3 direction_inverse;
 
-		Ray(const XMFLOAT3& newOrigin = XMFLOAT3(0, 0, 0), const XMFLOAT3& newDirection = XMFLOAT3(0, 0, 1), float newTMin = 0, float newTMax = std::numeric_limits<float>::max()) : Ray(XMLoadFloat3(&newOrigin), XMLoadFloat3(&newDirection), TMin, TMax) {}
```
turanszkij, line 126: Good stuff, uninitialized Tmin, Tmax

Editor/meshoptimizer/clusterizer.cpp
portaloffreedom: This was more the IDE (clion) suggestion, saying that using the c headers directly in C++ is deprecated we should use <clibrary> instead. Probably doesn't make a difference at all, since all headers are most likely guarded for c++ inclusion. Didn't know this was a library, I'll remove header changes from all libraries.

WickedEngine/wiGraphicsDevice_Vulkan.cpp
portaloffreedom: because the assert will just exit the program anyway, right?
oh! But only in debug mode, otherwise they are compiled out. It doesn't matter because it's the last statement of the switch, but for style consistency I could add it back.

WickedEngine/wiRenderPath3D.cpp
portaloffreedom: same as above, but in this case I made a mistake. The subresource_index has to be signed, because `CreateSubresource` can return a negative number.

WickedEngine/wiSpriteAnim_BindLua.cpp
portaloffreedom: I'm thinking of enabling this particular warning on Linux as an error. It seems pretty useful and we automatically test it in the CI.

WickedEngine/wiGraphicsDevice_Vulkan.cpp
turanszkij: yes, always have break; in switches pls (unless it always returns)
The assert can be ignored even in debug mode on Windows.

WickedEngine/wiSpriteAnim_BindLua.cpp
turanszkij: ok, if this can be enabled that's useful

WickedEngine/wiRenderPath3D.cpp
turanszkij: it would be best to change the assert to not warn instead by casting

WickedEngine/wiGraphicsDevice_Vulkan.cpp
portaloffreedom: from the preview in the discussion tab I totally mistook the code with another component. default is definitely better.

portaloffreedom:
And I think the only part left are implicit type conversion and comparison + some switch cases not exausted. But none of them seem very important, so I will just closed this PR

PR #528, opened 2022-08-23
Compiler warnings fixes (part 2)

portaloffreedom:
Please double check the new `assert(0)` I've added, if they are correctly placed

- added extra useful warnings
- added a virtual destructor when necessary
- cleaned up and added default to different switch blocks (should have zero impact on performance)
- cleaned and made faster `ComboBox::RemoveItem(int index)`

part 2 of #522

### DIFF
--- a/WickedEngine/CMakeLists.txt
+++ b/WickedEngine/CMakeLists.txt
@@ -249,6 +249,8 @@ else ()
 			-Wwrite-strings
 			-Winit-self
 			-Wreturn-type
+			-Wreorder
+			-Werror=delete-non-virtual-dtor
 			-Werror)
 
 	target_link_libraries(${TARGET_NAME} PRIVATE dl)
--- a/WickedEngine/wiArchive.cpp
+++ b/WickedEngine/wiArchive.cpp
@@ -15,7 +15,8 @@ namespace wi
 	{
 		CreateEmpty();
 	}
-	Archive::Archive(const std::string& fileName, bool readMode) : fileName(fileName), readMode(readMode)
+	Archive::Archive(const std::string& fileName, bool readMode)
+		: readMode(readMode), fileName(fileName)
 	{
 		if (!fileName.empty())
 		{
--- a/WickedEngine/wiGUI.cpp
+++ b/WickedEngine/wiGUI.cpp
@@ -2388,16 +2388,11 @@ namespace wi::gui
 	}
 	void ComboBox::RemoveItem(int index)
 	{
-		wi::vector<Item> newItems(0);
-		newItems.reserve(items.size());
-		for (size_t i = 0; i < items.size(); ++i)
-		{
-			if (i != index)
-			{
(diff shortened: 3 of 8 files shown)


portaloffreedom:
A few notes on the two extra warnings this PR is adding:

`-Wreorder` just to be sure that the constructor initializer list elements are in the right order. This does not affect code execution at all, just readability (always put first the one variable that is initialized first)

```
In file included from /home/matteo/projects/games/WickedEngine/WickedEngine/wiArchive.cpp:1:
/home/matteo/projects/games/WickedEngine/WickedEngine/wiArchive.h: In constructor ‘wi::Archive::Archive(const std::string&, bool)’:
/home/matteo/projects/games/WickedEngine/WickedEngine/wiArchive.h:23:29: error: ‘wi::Archive::fileName’ will be initialized after [-Werror=reorder]
   23 |                 std::string fileName; // save to this file on closing if not empty
      |                             ^~~~~~~~
/home/matteo/projects/games/WickedEngine/WickedEngine/wiArchive.h:18:22: error:   ‘bool wi::Archive::readMode’ [-Werror=reorder]
   18 |                 bool readMode = false; // archive can be either read or write mode, but not both
      |                      ^~~~~~~~
/home/matteo/projects/games/WickedEngine/WickedEngine/wiArchive.cpp:18:9: error:   when initialized here [-Werror=reorder]
   18 |         Archive::Archive(const std::string& fileName, bool readMode) : fileName(fileName), readMode(readMode)
      |         ^~~~~~~
```

`-Werror=delete-non-virtual-dtor` This checks if an object is deleted as a father class but could have been a child class. If the deconstructor is not virtual, the wrong deconstructor is called. 

```
In file included from /home/matteo/projects/games/WickedEngine/WickedEngine/wiRenderPath_BindLua.h:3,
                 from /home/matteo/projects/games/WickedEngine/WickedEngine/wiRenderPath_BindLua.cpp:1:
/home/matteo/projects/games/WickedEngine/WickedEngine/wiLuna.h: In instantiation of ‘static int Luna<T>::gc_obj(lua_State*) [with T = wi::lua::RenderPath_BindLua; lua_State = lua_State]’:
/home/matteo/projects/games/WickedEngine/WickedEngine/wiLuna.h:92:3:   required from ‘static void Luna<T>::Register(lua_State*, const char*) [with T = wi::lua::RenderPath_BindLua; lua_State = lua_State]’
/home/matteo/projects/games/WickedEngine/WickedEngine/wiRenderPath_BindLua.cpp:44:38:   required from here
/home/matteo/projects/games/WickedEngine/WickedEngine/wiLuna.h:262:25: error: deleting object of polymorphic class type ‘wi::lua::RenderPath_BindLua’ which has non-virtual destructor might cause undefined behavior [-Werror=delete-non-virtual-dtor]
  262 |                         delete(*obj);
      |                         ^~~~~~~~~~~~
```

turanszkij (approved):

PR #534, opened 2022-08-26
Extend Lua Scripting

megumumpkin:
Added new property implementation for Luna classes.
Added option to add custom parameters to pass for script launch.
Implement more bindings for `wi::scene::Scene`


Content/scripts/weather_animation.lua
```diff
@@ -0,0 +1,45 @@
+runProcess(function()
```
turanszkij, line 1: Sample scripts should follow how the other scripts work (comments, posting start and end messages on backlog)

WickedEngine/wiScene_BindLua.cpp
```diff
@@ -4,6 +4,13 @@
 #include "wiEmittedParticle.h"
 #include "wiTexture_BindLua.h"
 #include "wiPrimitive_BindLua.h"
+#include <LUA/lua.h>
+#include <Utility/DirectXMath.h>
```
turanszkij, line 8: You must not include this, it is causing compile error on windows. On windows, we use a different version of math library. You can include wiMath.h to use math library.

turanszkij (changes requested):
All new lua binding functions must be added to the documentation (no need to comment them id they are trivial). Otherwise, we will forget what was not added to documentation. It is important , because we don't have code completion/intellisense for lua.
+ the other comments I made.

Content/scripts/weather_animation.lua
megumumpkin: For the documentation of the binding functions, where should I put this?

turanszkij (approved):

turanszkij (changes requested):
Documentation?

megumumpkin:
> Documentation?

Done.

turanszkij (approved):

PR #537, opened 2022-08-28
Lua Extend Refactor

megumumpkin:
- Cleans up code repetition for newly added bindings
- Add new LuaProperty and PropertyFunction() for easily creating bindings for properties
- Refactor & Improved bindings for WeatherComponent
- Exposed all properties for MaterialComponent
- Add Setters for TransformComponent

### DIFF
--- a/Content/Documentation/ScriptingAPI-Documentation.md
+++ b/Content/Documentation/ScriptingAPI-Documentation.md
@@ -576,9 +576,9 @@ An integer mask that can be used to group entities together for certain operatio
 
 #### TransformComponent
 Describes an orientation in 3D space.
-- Position : Vector XYZ  -- query the position in world space
-- Rotation : Vector Quaternion  -- query the rotation as a quaternion in world space
-- Scale : Vector XYZ  -- query the scaling in world space
+- Translation_local : Vector XYZ  -- query the position in world space
+- Rotation_local : Vector Quaternion  -- query the rotation as a quaternion in world space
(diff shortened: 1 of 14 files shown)


Content/Documentation/ScriptingAPI-Documentation.md
```diff
@@ -530,6 +576,12 @@ An integer mask that can be used to group entities together for certain operatio
 
 #### TransformComponent
 Describes an orientation in 3D space.
+- Position : Vector XYZ  -- query the position in world space
```
turanszkij, line 579: The problem here is that GetPosition() returns position in world space (depending on parenting or animation system, or some other system modified its world matrix explicitly), but the TransformComponent::translation_local, etc are in local space. So these properties should be Translation_local, Rotation_local and Scale_local instead.

turanszkij (changes requested):

WickedEngine/wiLua.h
```diff
@@ -2,8 +2,6 @@
 #include "CommonInclude.h"
 #include "wiMath.h"
 
-#include <string>
```
turanszkij, line 5: Windows build is failing because of this removal

Content/Documentation/ScriptingAPI-Documentation.md
megumumpkin: So, what do we do with SetPosition and the like?

turanszkij (approved):

PR #541, opened 2022-08-29
Volumetric clouds improvements

Kliaxe:
Made improvements to volumetric cloud renderer:
 - Reprojection update
 - Cloud model changes
 - Weather map import
 - Local lights
 - Environment capture
 - Volumetric cloud shadow
 - Environment capture
 - Aerial perspective for clouds

I wanted to simply focus on the rendering, so I didn't add adjustable values for eg. shadows, aerial perspective. We can always add those if needed.
Anything is subject to change.


Kliaxe:
Btw I would recommend that we regenerate weather for scenes, since I changed a few default values

turanszkij:
It's great, but there is this problem with MSAA probe (that you already mentioned): 
image
If I set it to non-MSAA, then again back to MSAA, it gets better (I think depth values from non-MSAA version will be used)
I will take a look.

turanszkij:
> Btw I would recommend that we regenerate weather for scenes, since I changed a few default values

Do you have more info about what should be regenerated?

turanszkij (changes requested):
At first look, it seems only MSAA probe needs to be fixed. Sponza has MSAA probe by default and it messes up the lighting when probe shows clouds through geometry.

turanszkij:
MSAA probes don't write the envmapDepth texture (they are using envrenderingDepthBuffer_MSAA instead). I will try to add support for auto depth resolve, I think vulkan and dx12 both support it now.

Kliaxe:
> Do you have more info about what should be regenerated?
It's just the VolumetricCloudParameters inside scenes that have the old weather parameters. Do you have an easy way to reset those values, or would we have to manually go through every wiscene that have weather?

Kliaxe:
oops didn't mean to close xD

turanszkij:
I don't have an easy way. I didn't see any scenes that look bad with the new settings, though haven't checked all of them.

turanszkij:
Also, my MSAA resolve updates are here, I think I'll merge it from that PR if it's ok, so don't have to PR back and forth: https://github.com/turanszkij/WickedEngine/pull/544
Couldn't get depth resolve to work in DX12, so I changed sponza envprobe to be non-msaa. Vulkan works fine with MSAA probe now.

turanszkij:
Btw, clouds comparison, old one:
image

New one:
image

New one looks a bit more "solid", kind of liked the look of the old more "fluffy" clouds.

Kliaxe:
I changed some performance and modelling settings that make the clouds look nicer imo and perform better. We could also just wait if we decide to make more changes.

Yes let's do that.

I tried to make another attempt for terrain gen. It should resemble the old one a bit better.
This approach should also give the clouds more detail. Ultimately I could instead go back to the drawing board and design new default clouds, where we'd use reference photos and eg. modify the weather map.

Current:
before

New (WIP):
after

turanszkij:
I think I would like to merge it now, and it can be later improved. On the #544 branch I changed the implementation from depth resolve to reading the multisampled depth directly instead, so dx122 and vulkan also work now with MSAA probe.

PR #562, opened 2022-09-09
Change RenderPassAttachment texture member to be owning

jdswebb:
- Change RenderPassAttachment texture member to be owning (`const Texture*` -> `Texture`). Due to on-demand PSO creation, the description on the textures can potentially be read at any point during the RenderPass's lifetime. The previous use of a `Texture*` meant that a _particular reference_ of a texture had to be kept live.
    - There are only around 150 RenderPassAttachment objects live at any moment in Wicked Engine, and they are created up-front. So the additional cost of ref counting and description duplication should not be an issue.
- Changed the RenderPassAttachment builder functions to not accept null textures. The graphics device does not actually handle them being null and would crash.

Testing:

Loaded up a terrain scene in the editor with d3d12 and vk.

### DIFF
--- a/WickedEngine/wiGraphics.h
+++ b/WickedEngine/wiGraphics.h
@@ -625,143 +625,6 @@ namespace wi::graphics
 		}
 	};
 
-	struct RenderPassAttachment
-	{
-		enum class Type
-		{
-			RENDERTARGET,
-			DEPTH_STENCIL,
-			RESOLVE, // resolve render target (color)
-			RESOLVE_DEPTH,
-			SHADING_RATE_SOURCE
-		} type = Type::RENDERTARGET;
-		enum class LoadOp
-		{
-			LOAD,
-			CLEAR,
-			DONTCARE,
-		} loadop = LoadOp::LOAD;
-		const Texture* texture = nullptr;
-		int subresource = -1;
-		enum class StoreOp
-		{
-			STORE,
-			DONTCARE,
-		} storeop = StoreOp::STORE;
-		ResourceState initial_layout = ResourceState::UNDEFINED;	// layout before the render pass
-		ResourceState subpass_layout = ResourceState::UNDEFINED;	// layout within the render pass
-		ResourceState final_layout = ResourceState::UNDEFINED;		// layout after the render pass
-		enum class DepthResolveMode
-		{
-			Min,
-			Max,
-		} depth_resolve_mode = DepthResolveMode::Min;
-
-		static RenderPassAttachment RenderTarget(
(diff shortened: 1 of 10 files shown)


WickedEngine/wiGraphics.h
```diff
@@ -625,6 +625,114 @@ namespace wi::graphics
...
+		TextureDesc	desc;
+
+		// These are only valid if the texture was created with CPU access (USAGE::UPLOAD or USAGE::READBACK)
+		const SubresourceData* mapped_subresources = nullptr;	// an array of subresource mappings in the following memory layout: slice0|mip0, slice0|mip1, slice0|mip2, ... sliceN|mipN
+		size_t mapped_subresource_count = 0;					// the array size of mapped_subresources (number of slices * number of miplevels)
+
+		constexpr const TextureDesc& GetDesc() const { return desc; }
+	};
+
 	struct RenderPassAttachment
```
jdswebb, line 736: Btw - this moved down because now it must be defined after `Texture`.

turanszkij (changes requested):
I checked the build, but black screen flickering is happening at some points. For example creating a new terrain causes random flickering. Looks like only in dx12,

jdswebb:
Fixed that check, I can't seem to reproduce any d3d12 terrain flicker so far.

turanszkij:
The whole screen is flashing to black for me if I create a new terrain, but not the terrain itself. Also some flashes to black when opening fighting game script for example, but then it only happens for one frame (probably when it switches the renderpath). Also when exiting the script with ESC for one frame..

jdswebb:
With debug device I get an error, so hopefully the same cause even though I've not been able to see it - will have a look.

jdswebb:
It's because of this dummy texture:

```
		internal_state->dummyTexture.desc.format = desc->format;
		internal_state->dummyTexture.desc.width = desc->width;
		internal_state->dummyTexture.desc.height = desc->height;
		internal_state->renderpass = {};
		wi::helper::hash_combine(internal_state->renderpass.hash, desc->format);
		internal_state->renderpass.desc.attachments.push_back(RenderPassAttachment::RenderTarget(internal_state->dummyTexture));
```

I could stop errors by not checking texture.IsValid() in `pso_validate`, because in this case the texture is not valid but its description is.

turanszkij:
removing that IsValid check seems to work fine for me. Could you make the change on your branch?

jdswebb:
No errors on d3d12 or vk for me now and the new assert is never hit.

turanszkij (approved):

PR #573, opened 2022-09-27
helper: added loadTextureFromMemory and convertTextureData

LeeBamberTGC:
Started a fresh fork for GameGuru MAX with the latest version of the Wicked Engine.  I hope it is okay to submit careful pull requests when we add some that is generic enough to benefit many :)

### DIFF
--- a/WickedEngine/wiHelper.cpp
+++ b/WickedEngine/wiHelper.cpp
@@ -112,6 +112,197 @@ namespace wi::helper
 		}
 	}
 
+	bool loadTextureFromMemory(wi::vector<uint8_t>& texturedata, const wi::graphics::Texture& texture)
+	{
+		using namespace wi::graphics;
+
+		GraphicsDevice* device = wi::graphics::GetDevice();
+
+		TextureDesc desc = texture.GetDesc();
+
+		uint32_t data_count = desc.width * desc.height;
+		uint32_t data_stride = wi::graphics::GetFormatStride(desc.format);
+		uint32_t data_size = data_count * data_stride;
+
+		Texture stagingTex;
+		TextureDesc staging_desc = desc;
+		staging_desc.usage = Usage::UPLOAD;
+		staging_desc.layout = ResourceState::COPY_SRC;
+		staging_desc.bind_flags = BindFlag::NONE;
+		staging_desc.misc_flags = ResourceMiscFlag::NONE;
+		staging_desc.mip_levels = 1;
+		bool success = device->CreateTexture(&staging_desc, nullptr, &stagingTex);
+		assert(success);
+
+		if (stagingTex.mapped_data != nullptr )
+		{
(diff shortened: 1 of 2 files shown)


WickedEngine/wiHelper.cpp
```diff
@@ -112,6 +112,197 @@ namespace wi::helper
...
+				float r = ((pixel >> 16) & 255) / 255.0f;
+				float a = ((pixel >> 24) & 255) / 255.0f;
+
+				uint32_t rgba8 = 0;
+				rgba8 |= (uint32_t)(r * 255.0f) << 0;
+				rgba8 |= (uint32_t)(g * 255.0f) << 8;
+				rgba8 |= (uint32_t)(b * 255.0f) << 16;
+				rgba8 |= (uint32_t)(a * 255.0f) << 24;
+				data32[i] = rgba8;
+			}
```
turanszkij, line 301: `successfully_converted = true;` seems to be missing from here

WickedEngine/wiHelper.cpp
```diff
@@ -112,6 +112,197 @@ namespace wi::helper
...
+		TextureDesc staging_desc = desc;
+		staging_desc.usage = Usage::UPLOAD;
+		staging_desc.layout = ResourceState::COPY_SRC;
+		staging_desc.bind_flags = BindFlag::NONE;
+		staging_desc.misc_flags = ResourceMiscFlag::NONE;
+		staging_desc.mip_levels = 1;
+		bool success = device->CreateTexture(&staging_desc, nullptr, &stagingTex);
+		assert(success);
+
+		if (stagingTex.mapped_data != nullptr )
```
turanszkij, line 137: This texture copy is not very robust, it will fail for textures except simple 2D textures. The `saveTextureToMemory` function has an example of a correct texture data copy that will correctly handle mip levels, slices and 3D textures. The copy must take into account the row and slice pitch to work correctly. It also handles block compressed texture format correctly.

For simpler implementation, you can also use the `CreateTexture()` second argument initial data, and you don't need to invoke copies on command list. (if that fits your needs)

turanszkij (changes requested):
Thanks for the updates. I noticed two issues with the changes. I think the loadTextureFromMemory should be improved if we want this to be a good helper function in the engine.

WickedEngine/wiHelper.cpp
LeeBamberTGC: Thanks for the immediate reply :)  Happy to expand contributions as needed.  Merging and syncing with the master version makes a lot of sense for MAX.  Just going through a few lines to sync our code with your new naming system and files. Naturally, I found the WICKEDENGINE_BACKWARDS_COMPATIBILITY_0_59 too late, so proceeding to do a full conversion, is probably for the best.  Will commit more once I am compiling again.

PR #576, opened 2022-09-30
Add name for subset mesh

Cop46:
It's just for add a name parameter to the subset mesh. This is practical to identify surface mesh by name.


turanszkij:
I don't understand why this would be needed. You already get the subset itself from the mesh.

Cop46:
Yep,  but this is really practical to identify surface by name. The id of subset mesh change when import from obj and after serialize the subset index is not the same. With a subset name I can send my proper format like surface_toto_tata and split with "_" and pass many other informations by this way directly with the serialization.

Cop46:
Of course by this way this is possible to group subset with a string id like this totoGroup1_blabla...

turanszkij:
Identifications by strings is not a good solution usually. This should be inside your application if you really want to do this. I suggest you look into creating a custom component or something that doesn't add bloating to the core engine side.

PR #594, opened 2022-11-17
Add capability to export GLTF (proposal)

megumumpkin:
- Proper import with flipping the Z axis on all scene data, no more mesh face flipping during duplication
- Export scene to GLTF and GLB file format

Exporting to gltf can be done by appending the .gltf or .glb file extension while saving the scene (save as), glb file will embed all the resources while gltf won't.

Currently has issues with reimporting with crash.


Editor/ModelImporter_GLTF.cpp
```diff
@@ -3006,3 +3170,1238 @@ void Import_Mixamo_Bone(LoaderState& state, Entity armatureEntity, Entity boneEn
...
+	"jpeg",
+};
+
+inline std::string _ExportHelper_GetOriginalTexture(std::string texture_file)
+{
+	for(auto& ext : original_texture_extension_iterator)
+	{
+		std::string target_file = wi::helper::ReplaceExtension(texture_file, ext);
+		wi::backlog::post(target_file);
+		if(std::filesystem::exists(target_file))
```
turanszkij, line 3220: Should use `wi::helper::FileExists()` instead, because uwp platform can't use std::filesystem

turanszkij (commented):
This looks excellent. Only found a minor thing when looking over, with filesystem usage. I will need some additional time to test these changes on my end.

Editor/ModelImporter_GLTF.cpp
megumumpkin: About filesystem, is it ok to extend the wicked's filesystem functions to cover all functionalities from std::filesystem?

turanszkij: No, we don't need to write a complete file system replacement. It can be done on a case-by case basis when you need a feature that must be supported by all platforms.

turanszkij:
- It needed fix for the humanoid lookAt direction (it is only imported from VRM and Mixamo rigs).
- The exported file can be opened by Blender, but it fails with the engine and Windows model viewer as well. When debugging in the engine, the first issue when importing an exported gltf seems to be that the gltfModel doesn't contain any scenes.

turanszkij:
Gltf normal mapping test was not correct, so tangent.w needed to be multiplied by -1, it looks like now it's correct:
image

megumumpkin:
Are there any other stuff that needs to be added/removed?

turanszkij:
> Are there any other stuff that needs to be added/removed?

I need more time to test the importer changes. Also the exporting should be fixed, right now it generates a lot of separate files even for GLB which should be only 1 file, and they can't be re-imported by the engine, one of the exported files seems to always be zero-sized.

megumumpkin:
That could possibly be a texture file buffer, which I store them to a separate bin file.

turanszkij:
> Are you currently working on a commit right now?

No, currently not

megumumpkin:
- Added fixes to be able to export and then reimport the GLTF model
- Fix GLB export logic

turanszkij:
Nice, the exported models can now be reimported, but the animated model like CesiumMan will be broken when reimported, the DamagedHelmet will have some strange blue colors appear on it.

turanszkij (approved):

turanszkij:
> Nice, the exported models can now be reimported, but the animated model like CesiumMan will be broken when reimported, the DamagedHelmet will have some strange blue colors appear on it.

- Blue colors was due to the surface map was exported into the specular map slot.
- Some nodes were imported twice which broke the animated model re-import and caused duplicated instances if reimporting DamagedHelmet as well, it is now fixed
- Some other fixes were made too, I think nothing else remains

turanszkij (approved):
Should be merged when ready.

turanszkij (dismissed):

turanszkij (approved):

Issue #475, opened 2022-06-27
[Feature] Add AMD FidelityFX-FSR2


docstrang:
I think it's important to add FidelityFX Super Resolution 2.0.1

https://github.com/GPUOpen-Effects/FidelityFX-FSR2

Thanks

turanszkij:
Done

PR #597, opened 2022-11-25
FidelityFX FSR 2.1

turanszkij:
#475


turanszkij:
todo: motion blur and depth of field are broken when FSR 2 enabled, because they will then need to run at display resolution

PR #602, opened 2022-12-11
DXC update

turanszkij:


turanszkij:
Also fixed raytracing shader issues with new compiler

PR #603, opened 2022-12-11
Volumetric clouds update

Kliaxe:
- second cloud layer implementation
- cloud shadow fixes for flickering
- more precise depth buffers when intersecting against geometry
- disabled temporal AA and FSR2 jitter due to over blurred results (looks even better with TAA now)
- performance optimizations
- exposed shape parameters in weather window
- weather map changes

Anything is due for changes.
Should I also write documentation? Let me know if the cloud parameters etc. needs explanation.


Kliaxe:
I've made some performance changes. I can tell it should be much better already. I before completely forgot to check up with performance since I had 2~ ms which I for some reason thought wasn't different

turanszkij:
It didn't improve performance for me unfortunately. I also noticed this other issue, if I look with the camera up-down horizontally, this issue appears in the distance: 
image

Kliaxe:
I've reverted the cloud velocity stuff to what we've had before. After some consideration it was already good

The performance could also have impact from the system thickness that I set from 4000 to 8000, since rays traverse through more "stuff" even with skipping techniques. I'd like to experiment more with performance in these days and see where that takes me.

turanszkij:
It seems like it only did a small performance improvement for me: 
image

Kliaxe:
I've made a few more changes. More cleanups and optimizations is never bad, but I doubt it make a big impact, but you can try though.

Tbh this cloud ray marcher is getting really fast! Just to compare I'm getting around 2~ ms in Unreal with their cloud system (default setup) with one layer, and 1~ ms with 2 layers in Wicked on my laptop.
Of course more layers takes up more resources, there is doubt denying that. Take MS Flight Sim that has 3 layers as an example.

I have a suggestion. Should we have 2 layers enabled by default? That way users can see what the weather system is capable of. And then make some specific adjustments for the terrain gen, because the cloud system is made highly customizable in terms of modelling but also performance, and can be tweaked to be much faster than it currently is. What are your thoughts on this?

turanszkij:
It's a noticeable improvement for me, now it's about 3.1 ms when looking up like prev screenshot at same resolution (4k) and second layer disabled. For comparison, on the master branch they take 2.3 ms with similar camera angle.

If you can find nice default setting for the clouds that can improve the look of the terrain, that would be great. I'd like though if performance was not heavier than currently on the master branch if possible. I think enabling second layer of clouds is ok to be optional if that is too heavy.

Note that performance numbers are not very accurate for clouds as it is running on async compute same time with shadow map rendering, so it's likely not much sense to compare with unreal engine's method.

Also, about the weather panel controls (sliders and weather map buttons), could you rename them so that the text fits into the window? For example "first layer" could be just "1" and "second layer" simply "2". Sorry that the gui is a bit rigid in this regard.

Kliaxe:
I will gladly make something that fits the terrain nicely. It's definitely possible now to make something faster than the old. Also did you like the look of the clouds with the old? Because while I'm at it, I can use some reference photos from real life to make a specific look for the terrain.

You're right with the UI, it is getting out of hand. I also thought about make the text above the sliders, but that would break the "order rule" of the rest.

turanszkij:
I like the old clouds too, but I'd be curious if you tweak it to look different. 

For the gui, sure you can add more controls for the clouds if you want. The window is now scrollable so it is not a problem if there are many controls. You will need to tweak the height of the weather window eventually if elements start to go off of it (don't want a double scroll bar both on weather window and components window, only one on the components window).

If you want to have the text above the sliders, I think the gui element's `font.params.v_align` can be modified and the text will be aligned accordingly relative to the control. You can experiment with different text alignment if you'd like.

Kliaxe:
Sorry it took a while. Had many things to attend to. I exposed cloud shape parameters in the weather window, made more performance optimizations, changed the weather map, and made a new look for the terrain clouds. Let me know if anything needs  change. I'm curious what performance you now will get

turanszkij (approved):
This looks amazing, especially like the rain clouds effect. Performance is about the same as master for me when looking up, but it seems faster with a normal camera angle, goes even below 1 ms at 4K. Also cool that you added those extra params to the editor. It will be an honor to merge this. Ready? :)

Kliaxe:
Yeah let's merge!

turanszkij:
Done, thanks for your work!

PR #605, opened 2022-12-12
Update linux DXCompiler to commit e9137cd1d

portaloffreedom:
should be the same version as windows


turanszkij:
Thanks, the dxcompiler was updated to this commit instead: 2c3d965b2fc734c09e76631de65f8ffbe4e3e68b

PR #618, opened 2023-01-08
General installer overhaul

MolassesLover:
## About
Improvements to the installer, since I'm a lot better at programming now. I made things a lot cleaner, and I'm making things much easier for the user; someone who has never opened a terminal should be able to use this when I'm done.

## Changes
- Moved from [TtkBootstrap](https://ttkbootstrap.readthedocs.io/en/latest/) to vanilla TkInter.
- Created a dependency manager to install TkInter.
- Improved Arch, Fedora, and Ubuntu support.

### Bugs
- [x] **Fix** [`stderr`](https://github.com/MolassesLover/WickedEngine/blob/f22a733aeeaf5a4980b38f7624f722d2cbcb930b/Editor/Linux/Installer/Installer.py#L24) not printing.
### UX
- [ ] Move to `pkexec` from `sudo`.
- [x] Create a sliding, presentation style UI.
- [x] Install the entire editor to a specific directory.
- [x] Standardise `$PREFIX/share/wicked-engine` as the editor path on Unix.
- [ ] **Option** to pick between `$HOME`, `/usr/local/`, or any other prefix.
- [ ] **Option** to add the editor path to detected runcom files as a variable.
- [ ] **Option** to build from scratch.
- [ ] **Option** to install from a list of found package managers in `$PATH`.
### CI?
I feel like it might be a good idea to compile this installer into an executable file, so others who have never used Python or a terminal can use it as any other installer one might find themselves downloading. Of course, if you are not comfortable with such an idea, that is okay too!

This would involve changes to the continuous integration tasks, making them a little bit slower. Windows and Ubuntu each would build the executable. The tasks would create an `.exe`, generic Python executable, and an `.appimage` file.


MolassesLover:
Sorry for the lack of updates, I work on this in my private repository because I experiment a lot.

turanszkij:
When will this be done?

MolassesLover:
> When will this be done?

I have no idea. It'll stay as a draft in the meantime.

turanszkij:
I don't wanna keep this here forever.

PR #636, opened 2023-02-04
Realistic sky improvements

Kliaxe:
- Aerial Perspective (opaque and transparent)
 - Volumetric shadows for Atmosphere
 - Clouds receive shadows from objects
 - New adjustments in weather window
 - Better cloud upsampler

AP is the missing piece from Realistic Sky and it adds a whole new look to the game world. Together with shadowing casted from objects, it simulates how atmosphere should behave IRL (in theory).

As always, if anything isn't according to specifications I'll be happy to change accordingly! Namings, structure, syntax etc...
And if documentation is needed, I'll also be up for that.


Kliaxe:
I'm also curious to hear your opinion on what I've done with the way we switch to high quality. 
We switch to another pass since it has to be done in half-res and need variation in samples, and thus need temporal.
Due to this, we can't really use skyPS or envMap_skyPS, hence we disable color and composite the pass over.

WickedEngine/ArchiveVersionHistory.txt
```diff
@@ -1,5 +1,6 @@
 This file contains changelog of wi::Archive versions
 
+89: serialized WeatherComponent::atmosphereParameters: rayMarchMinMaxSPP, distanceSPPMaxInv and aerialPerspectiveScale
```
turanszkij, line 3: We don't need to increase archive version number for things in scene serialization any more. The component managers have their own version number now which is preferred usage now (where they are declared in the Scene, you can specify version number in ComponentLibrary::Register). But this will work too, just not the best place. If you use the version number of component, then instead `archive.GetVersion()` you will need to use `seri.GetVersion()`

WickedEngine/shaders/envMap_skyPS_dynamic.hlsl
```diff
@@ -6,8 +6,16 @@
...
 {
-	float3 normal = normalize(input.nor);
-	float4 color = float4(GetDynamicSkyColor(normal), 1);
-	color = clamp(color, 0, 65000);
-	return float4(color.rgb,1);
+	float4 color = 0;
+	
+	// When realistic sky 'High Quality' is enabled, let's disable skybox color since we render high quality sky in SkyAtmosphere
+	[branch]
+	if ((GetFrame().options & OPTION_BIT_REALISTIC_SKY_HIGH_QUALITY) == 0)
```
turanszkij, line 13: Wouldn't it be better to disable the whole pass instead?

WickedEngine/shaders/lightingHF.hlsli
```diff
@@ -42,79 +43,6 @@ inline void ApplyLighting(in Surface surface, in Lighting lighting, inout float4
 	color.rgb += surface.emissiveColor;
 }
 
-inline float3 sample_shadow(float2 uv, float cmp)
```
turanszkij, line 45: Is it possible to keep the shadow code here instead of moving to a new file? If yes, I would prefer to keep it, because some people already reference shadow code from this file and it will be just confusing if it's suddenly moved. Also, url links to it will be no longer good. If there is a hard dependency issue, then it can be moved to a new file, but if it's only a refactor, then I don't want it.

turanszkij (changes requested):

WickedEngine/shaders/envMap_skyPS_dynamic.hlsl
Kliaxe: Yeah my thoughts as well, I'll do a commit later

turanszkij: Something to keep in mind, the main color pass is not clearing render target, but it expects the sky drawing will overwrite any untouched pixels. If we don't draw the sky, we will need to clear the render target in that case to not have undefined results show up. This will need to be done in RenderPath3D.cpp in the render pass where DrawSky() is being called. But only clear if sky pixel shader is disabled, otherwise we can avoid the clearing.

WickedEngine/shaders/lightingHF.hlsli
Kliaxe: I tried to avoid creating a new file, but since we call shadow in skyAtmosphere.hlsli that also is included in lightingHF.hlsli as well as other files that needed transmittance from atmosphere. Do you have any suggestions what we can do?

turanszkij: Ok, if this is dependency problem then it is fine. It wasn't clear to me by looking at changelog.

WickedEngine/ArchiveVersionHistory.txt
Kliaxe: Oh I see. I wondered why Register with a version number, I'll make change

WickedEngine/shaders/envMap_skyPS_dynamic.hlsl
Kliaxe: But we really can't do that for visibility sky, since we allocate tiles and dispatch indirectly?

turanszkij: Visibility sky is not a problem, visibility rendering mode will overwrite all pixels guaranteed.

turanszkij: Or... maybe it will be a problem now that I think about it. For now, I think you can leave the visibility pipeline though.

turanszkij: probably just needs a ClearUav() in the visibility pipeline in this case instead of clear by render pass

WickedEngine/ArchiveVersionHistory.txt
Kliaxe: Would this need an entry in archive history?

turanszkij: And this doesn't need tracking in ArchiveVersion.txt. Since the changes won't be scattered across codebase and be local to a component serializing, it will be easy to find changes of component versions just in the code.

Kliaxe: Should be good now!

WickedEngine/shaders/envMap_skyPS_dynamic.hlsl
Kliaxe: Perhaps something like this? I didn't do the visibility pipeline, sounds like you had an idea

Kliaxe: I didn't combine static and gradient background into the atmosphere pass, since I thought that atmosphere should be kept by itself, what do you think?

WickedEngine/wiRenderPath3D.cpp
```diff
@@ -1079,7 +1079,7 @@ namespace wi
 				RenderPassImage rp[] = {
 					RenderPassImage::RenderTarget(
 						&rtReflection,
-						RenderPassImage::LoadOp::DONTCARE,
+						RenderPassImage::LoadOp::CLEAR,
```
turanszkij, line 1082: We only want to clear if sky is not rendered, otherwise it is more optimal to not clear.

WickedEngine/wiRenderPath3D.cpp
```diff
@@ -1280,7 +1280,7 @@ namespace wi
 			RenderPassImage rp[] = {
 				RenderPassImage::RenderTarget(
 					&rtMain_render,
-					RenderPassImage::LoadOp::DONTCARE
+					RenderPassImage::LoadOp::CLEAR
```
turanszkij, line 1283: We only want to clear if sky is not rendered, otherwise it is more optimal to not clear.

WickedEngine/wiRenderer.cpp
```diff
@@ -6998,7 +6998,7 @@ void RefreshEnvProbes(const Visibility& vis, CommandList cmd)
 				),
 				RenderPassImage::RenderTarget(
 					&vis.scene->envmapArray,
-					RenderPassImage::LoadOp::DONTCARE,
+					RenderPassImage::LoadOp::CLEAR,
```
turanszkij, line 7001: We only want to clear if sky is not rendered, otherwise it is more optimal to not clear.

WickedEngine/shaders/skyPS_dynamic.hlsl
```diff
@@ -3,20 +3,13 @@
...
+	float4 unprojected = mul(GetCamera().inverse_view_projection, float4(clipspace, 0.0f, 1.0f));
+	unprojected.xyz /= unprojected.w;
 
-		const float3 V = normalize(unprojected.xyz - GetCamera().position);
+	const float3 V = normalize(unprojected.xyz - GetCamera().position);
 		
-		color = float4(GetDynamicSkyColor(V), 1);
-		color = clamp(color, 0, 65000);
-	}
+	float4 color = float4(GetSimpleSkyColor(V), 1);
```
turanszkij, line 11: We can't draw the realistic sky in the pixel shader anymore?

turanszkij (changes requested):

WickedEngine/shaders/skyPS_dynamic.hlsl
Kliaxe: Maybe I misunderstood. So we keep pixel shader, but only disable when compute is required?

turanszkij: Yeah I thought keep the current rendering of realistic sky in pixel shader, but if you switch to high quality and compute shader is required, then disable the sky draw call.

Kliaxe: Gotcha, give me a minute

Kliaxe: I've pushed new changes

turanszkij:
1. There are some problems, like if high quality sky is not enabled, I get a very blueish result. Just create a base default terrain to see it. 

2. Also, the sky pipeline state is disabled when sky is high quality, but the draw call is still running. This will produce undefined results.

3. It is quite convoluted and I couldn't follow it, does the high quality sky require a compute pass? Or the aerial? Or both? Why is the sun drawn as a separate pass now?

4. Things like this: `clearColor = RenderPassImage::LoadOp::CLEAR` are not a good naming, because LoadOp is not specifying a clear color. 

5. There are a lot of things inside the ifs for example: 
```
if (scene->weather.IsRealisticSky() && scene->weather.IsRealisticSkyHighQuality() && !scene->weather.skyMap.IsValid())
```
For this, we should expose a function perhaps inside the weather, that will tell us whether we need to:
- DrawSky() by pixel shader
- draw sky in compute
- do we need to clear render passes by LoadOp::Clear

6. Can we no longer render volumetric clouds inside async compute? I felt that was a good gain. Is this because clouds can receive shadows from shadow maps now?

turanszkij:
Is it possible to only stick to rendering sky in the pixel shader? It would remove a lot of complications of clearing etc. I would prefer to render high quality sky in pixel shader too without temporal even if it has performance hit if possible. I expect that using high quality sky will only happen in special circumstances, usually will be disabled.

Kliaxe:
The idea behind the separate pass for high quality was because of performance, but if high quality is considered a special case then it should actually fine to render the sky solely in PS. This would also make the design much better. I agree that it's getting tedious with all the checking, and would be better if it was done in PS. If high quality is done in PS and current resolution, then TAA should be enough to handle the shifting samples as well. The temporal pass was there to accumulate the shifting samples, and half-res couldn't keep up.

All of this would also eliminate the need of a separate sun pass. This was due to the sky being rendered at half-res and sun/stars was too low quality for their purpose.

Do you have any idea how to access depth information in skyPS and envMap_skyPS? With this, everything should fall into place.

I moved the cloud pass down due to the required shadow maps. I also felt this was a trade-off, but was there any other way? Perhaps with previous shadow-map access.

turanszkij:
Sky is only rendered on pixels where the depth = 0 (where nothing else was rendered). The texture_depth is accessible from sky, but it will have 0 values for every sky pixel.

If you can make the HQ sky in pixel shader like others, I would prefer that.

Cloud shadows are also quite niche - I don't see regular users really make use of it so it's a bit of a shame if they will make the regular clouds slower. Unfortunately the async is especially designed to run alongside shadow rendering, and there is no readable shadow map at that point.

turanszkij:
Yeah, aerial as separate pass sounds good. Maybe there could be issues with transparents tho. Maybe the transparents can just apply their own aerial perspective in their pixel shader?

I think it's better to keep cloud receive shadow, after all it will be a cool technique. Maybe move them back to async compute if receive shadow is disabled..

Kliaxe:
Oh yes, that's how it currently works, transparents use AP from LUT since they're rendered separately from opaque, so same approach like fog.

Will you take a look at that once merged? I'm unsure at this point what would be the best way to do it

turanszkij:
Yes, I will handle the async part, you don't need to.

Kliaxe:
I've made the changes as discussed and should be good now. Shadows is not available for visibility pipeline sky, so I left it disabled for now.

I also made a custom upsample pass for clouds, since we must use the depth info from clouds. If the clouds was between camera and objects, you should see that the quality is greatly decreased, but should be fixed now.

I also fixed a performance issue with high quality. If full-res is enabled, the performance is not amazing, but not insane either imo. This is a nice option to have though. If high quality is enabled, TAA is also needed since we jitter the samples (you should see the obvious difference), and I wondered why the reprojection was not as clean like before as it appears sharp. Turns out there exist different TAA techniques/modes that can benefit object AA and other temporal needs, like the alpha blend for grass, and especially the high quality atmosphere. 
Have you considered improving the TAA? If you don't mind I'll be happy to do an attempt. We can always revert. I can make some modes available, and you can decide what looks the best for Wicked.

WickedEngine/shaders/shadowHF.hlsli
```diff
@@ -107,4 +107,29 @@ inline float shadow_2D_volumetricclouds(float3 P)
 	return 1.0;
 }
 
+// Sample light and furthest cascade for large mediums (volumetrics)
+// Used with SkyAtmosphere and Volumetric Clouds
+inline void furthest_cascade_volumetrics(inout ShaderEntity light, inout uint furthestCascade)
```
turanszkij, line 112: Can this entity index be computed on CPU instead? Seems like a waste to run this loop in a shader. Or did you consider using the `GetWeather().most_important_light_index` ? The most important light index is the first directional light in the light array.

turanszkij (changes requested):
I had a quick look, it looks great and no visual issues that I could find so far. I noticed the furthest_cascade_volumetrics function, which I made some comments for.

I had a crash with this in my demo game script with this branch, that's also something I will want to take a look at later.

WickedEngine/shaders/shadowHF.hlsli
Kliaxe: Oh yes that's way better, have no idea why I didn't notice. I pushed a fix. I should even have made it run on the CPU instead regardless, my bad

Kliaxe:
Weird. I couldn't reproduce with the game samples. I also tried vulkan and checked through the files, but nothing inherently seemed faulty. Are there any commits where this doesn't happen?

Btw is the fighting game sample suppose to this fast? Looks like it's sped up 4x or so

WickedEngine/shaders/shadowHF.hlsli
turanszkij: Yes, it is offset into the light array, so it needs light offset into entity array. Just like how you did it.

turanszkij:
> Weird. I couldn't reproduce with the game samples. I also tried vulkan and checked through the files, but nothing inherently seemed faulty. Are there any commits where this doesn't happen?
> 
> Btw is the fighting game sample supposed to be this fast? Looks like it's sped up 4x or so

Sorry, I wasn't clear the crash happens in a game sample that I have, not the built in scripts, I will take a look at it today.

The fighting game is running 4x faster? Maybe you increased the Speed slider to 4x multiplier? It runs normally for me in default settings.

turanszkij:
I made a PR into your branch with some fixes. I think with those, everything will be good to merge. The game crash was due to some incorrectly versioned asset I had, so it was not your fault.

Kliaxe:
I made a fresh install of code and content, and it's still sped up. Pure speculation and nothing else. Probably something I've missed.

Do you have any thoughts on the TAA stuff?

turanszkij:
Oh, it seems that it's tied to whatever FPS you got. For me with 60FPS vsync it runs normally, without Vsync it doesn't. But this don't need to be handled by this PR.

For the TAA, can that be a separate discussion? I'm interested in what improvements there can be, but I also want the current simple and fast TAA to remain.

turanszkij (approved):

Kliaxe:
Yeah let's not get out of topic. I just think we should merge and as you said, take the discussion separately.

PR #645, opened 2023-02-22
Fog improvements

turanszkij:


turanszkij:
I love this new fog, the only problem I found is that from atmospheric camera, the height fog doesn't work well, it shows a straight line instead the planet's curve. Not sure if that can be solved efficiently:
image

Kliaxe:
That's the nature of the algorithm. It's based solely on the camera position and world center to evaluate where the fog is based on start and end heights. Since the skybox has currently infinite distance this unfortunately makes a solid line in the horizon, and is not scalable for planets. Ray marching would be great for this, and allows the fog to be distributed nicely on the planet, but I like the O(1) time and simplicity with this current approach, which of course has it's downside of being linear. I have an idea to use a max fog distance or cutoff for sky.

turanszkij:
I was thinking something like computing sky intersection from the planet top/bottom sphere radius that's used in atmosphere computations and using that for fog. Not sure whether that will be a good way, but I want to give it a try before merging.

Kliaxe:
I also tried to do that, but I couldn't get good results. I'm curious to see what you would do.

turanszkij:
I messed a lot with it, and found a solution that's probably good enough for now. Not always perfect, and kind of a hack but better than a straight line from top view:
image

PR #665, opened 2023-04-14
Added various int type casting calls for macOS support

MolassesLover:
## About

These changes will allow for compilation on macOS using GCC 12. 

Whilst this is a band-aid solution, it should hopefully not cause any problems.

### DIFF
--- a/WickedEngine/wiJobSystem.cpp
+++ b/WickedEngine/wiJobSystem.cpp
@@ -190,6 +190,27 @@ namespace wi::jobsystem
 			if (ret != 0)
 				handle_error_en(ret, std::string(" pthread_setname_np[" + std::to_string(threadID) + ']').c_str());
 #undef handle_error_en
+// To do: Rewrite this code for macOS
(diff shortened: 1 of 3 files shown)


turanszkij:
I think you should configure your compiler instead to allow current syntax, because it is c++ compliant. And the job system is commented out. It doesn't look like we will support mac os currently anyway.

MolassesLover:
@portaloffreedom Do you know how to configure MinGW's GCC 12 for this?

MolassesLover:
> And the job system is commented out.

I duplicated it into another preprocessor, so it's only commented out for macOS.

portaloffreedom:
@MolassesLover I did compile it on MacOS some years ago, and it did compile fine without any modification (the thread code was not there yet on linux). The biggest issue I had that SDL2 surfaces were incompatible with fake vulkan, but actually metal.
I gave up pretty early though, because I didn't care enough XD

What is the error that you are getting, that needs explicit type conversion to be fixed?

PR #666, opened 2023-04-14
Added the Dracula theme

MolassesLover:


turanszkij:
There is no highlight when mouse hovers over gui elements like buttons and checkboxes. The font shadow looks bad with this color theme. I think the skeleton icon is not very fitting.

MolassesLover:
> There is no highlight when mouse hovers over gui elements like buttons and checkboxes. The font shadow looks bad with this color theme. I think the skeleton icon is not very fitting.

They haven't added the FontAwesome bat icons, so it was either the crossbones or a cross. Any other ideas for the icon?

turanszkij:
There will be a theme editor to allow user created themes instead of making individual themes in pull requests: #667

PR #669, opened 2023-04-17
Tweaked Vulkan transfer queue family selection

jeweg:
Changed Vulkan transfer queue family search to more to first look for a fully exclusive transfer queue family, even without video and optical flow caps. If that fails, it will fall back to the previous behavior. This change works around driver issues in some scenarios.


turanszkij:
Thank you, I confirm this fixes the virtual texture issue in Vulkan with latest Nvidia drivers.

PR #681, opened 2023-05-22
update Localization zh_CN

conradax:
I was shocked that the xml I posted in issue has been adopted as I said it's just for font testing.
This one is much better but still a WIP.

Currently translation work is stalled because:
* I'm actually not capable to do it cause it's only months since I began touching graphics and C++, I need time.
* There are some problems in current localization solution and widget design (see below). I see you are making changes to graphics stuff frequently, and I don't want to bother you, so I didn't open issues.
* I think Wicked Engine is not ready for beginner(more documents), if someone has ability to use this engine, then he(she) must can read English well.

----------------------------------------------------

### 1. widget with same name overrides
eg. in Editor.cpp, line 3439
``` C++
	editorscene->tabCloseButton.Create("X"); //line 3439
	editorscene->tabCloseButton.SetLocalizationEnabled(false);
	editorscene->tabCloseButton.SetTooltip("Close scene. This operation cannot be undone!"); // This string is not found in xml
```
the above one probably overridden by this: (found in xml)
``` xml
			<section name="X" hint="Button">
				<entry id="0" hint="text">X</entry>
				<entry id="1" hint="tooltip">Remove currently selected subset. LODs will be lost and need to be regenerated!</entry>
```
another example is `FSR1.0` and `FSR2.1`'s `Sharpness` slider' tooltips, they share the same string loaded from xml file.

### 2. The xml file is just a mess
Buttons in the same toolbar are separated randomly, finding them is a pain.

The xml file is generated dynamically, so I think it's better to make translations in engine itself.  
like: *in debug build, add right click event for widget, when right clicking it, pop a small window ask user to enter translation directly.*

### DIFF
--- a/Editor/languages/简体中文.xml
+++ b/Editor/languages/简体中文.xml
@@ -424,7 +424,7 @@ This can add small shadows details to shadow maps in screen space.</entry>
 				<entry id="1" hint="tooltip">The sharpening amount to apply for FSR 2.1 upscaling.</entry>
 			</section>
 			<section name="Occlusion Culling: " hint="CheckBox">
-				<entry id="0" hint="text">Occlusion Culling: </entry>
+				<entry id="0" hint="text">遮挡裁剪: </entry>
 				<entry id="1" hint="tooltip">Toggle occlusion culling. This can boost framerate if many objects are occluded in the scene.</entry>
 			</section>
 			<section name="MipLOD Bias: " hint="Slider">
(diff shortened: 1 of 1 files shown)


turanszkij:
"I was shocked that the xml I posted in issue has been adopted as I said it's just for font testing."

What? It was not adopted, I don't know what you mean.

turanszkij:
Your first code example shows that localization is disabled for that element, that's why you don't see it. 

"the above one probably overridden by this"
What do you mean probably? It is not overridden because they are in separate section hierarchies, is it?

PR #716, opened 2023-07-26
Ensure IK gets only updated after a timestep occured

### LINKED ISSUE #707: IK seems to be framerate dependent and bugs out on higher framerates
See the attached video. With 60 FPS, the IK work as expected, with 165 FPS, there seem to be some rounding errors or something, making the IK very erratic (it's much more noticable at 165 than in this 60 FPS video, it seems like every second frame has the leg stretched instead of bent)

https://github.com/turanszkij/WickedEngine/assets/541637/4ecf61f6-f7ec-47b3-8ee1-5a3ca9d7250a

brakhane:
Update_IK gets called for every frame, but ground intersections are only checked up to 120 times per second.

This can lead to the Y velocity being >0 despite standing on the ground, messing up IK which is only done when the velocity is 0.

Fixes #707

Another way would be to just remove the check for `velocity.GetY() == 0` or make it `math.abs(velocity.GetY()) < 1` or something, but that feels even more hacky to me.

### DIFF
--- a/Content/scripts/character_controller/character_controller.lua
+++ b/Content/scripts/character_controller/character_controller.lua
@@ -329,6 +329,7 @@ local function Character(model_name, start_position, face, controllable)
 		position = Vector(),
 		controllable = true,
 		fixed_update_remain = 0,
+		timestep_occured = false,
 		root_bone_offset = 0,
 		foot_placed_left = false,
 		foot_placed_right = false,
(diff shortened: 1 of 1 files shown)


turanszkij (approved):

turanszkij:
Thank you for working on the fix!

PR #718, opened 2023-07-30
vulkan fix: ensure separate device queues for each queue type

brakhane:
I'm currently working on removing Vulkan bugs and valildation warnings that cause the engine to fail on my Linux box. This is the first change in a series of upcoming PR. 

----------

Device queues aren't thread-safe, so each queue type has a mutex. However, if a device doesn't provide dedicated queues for each device type, we end up using the same queue family for multiple queue types. So we need to create multiple separate device queues from the same family.

This commits ensures that by counting the number of shared queue families to allocate and use the appropriate amount of queue indices.

### DIFF
--- a/WickedEngine/wiGraphicsDevice_Vulkan.cpp
+++ b/WickedEngine/wiGraphicsDevice_Vulkan.cpp
@@ -5,7 +5,6 @@
 #include "wiBacklog.h"
 #include "wiVersion.h"
 #include "wiTimer.h"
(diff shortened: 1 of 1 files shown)


brakhane:
Seems like queueCount doesn't work as I thought it would, weirdly enough, it worked on my machine despite the IGPU only having 1 queue in the family.

I'll open another PR that fixes the mutexes instead. That was my previous approach which I abandoned because it obviously causes more locking

PR #719, opened 2023-07-30
vulkan fix: Ensure properties2 matches the physical device

brakhane:
Ensure that if there is more than one physical device, but no preferred one, (eg. iGPU and llvmpipe), properties2 will have the data for the selected device, and not the last one checked.

### DIFF
--- a/WickedEngine/wiGraphicsDevice_Vulkan.cpp
+++ b/WickedEngine/wiGraphicsDevice_Vulkan.cpp
@@ -2505,10 +2505,11 @@ using namespace vulkan_internal;
 			wi::vector<const char*> enabled_deviceExtensions;
 
 			bool h264_decode_extension = false;
+			bool suitable = false;
 
-			for (const auto& dev : devices)
-			{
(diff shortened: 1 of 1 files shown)


brakhane:
BTW, this fixes graphical bugs on my Intel Xe GPUs because it has stricted alignment requirements than llvmpipe

brakhane:
Noticed *properties_chain wasn't set to nullptr at the end, causing warnings, so I just added it to this PR. Also rebased to the current master

brakhane:
I'm getting some errors that indicate this approach will not work, I'll reopen it once I have checked.

brakhane:
Seems like queueCount doesn't work as I thought it would, weirdly enough, it worked on my machine despite the IGPU only having 1 queue in the family.

I'll open another PR that fixes the mutexes instead. That was my previous approach which I abandoned because it obviously causes more locking

brakhane:
... I was on the wrong PR, reopening this one

turanszkij (approved):

PR #720, opened 2023-07-30
vulkan: fix locking for non-dedicated queues

brakhane:
This replaces #718. I had hoped we could avoid having more lock contention, but it seems like this is not possible.

--------

Some devices don't offer dedicated queues, just one "does everything" queue. In those cases, our logically separate queues are the same Vulkan queue, so we need to make sure that those queue share the same mutex.


turanszkij (approved):

turanszkij:
Thank you for working on the Vulkan fixes!

PR #722, opened 2023-08-02
vulkan fix: add SHADER_STORAGE_IMAGE_MULTISAMPLE capability and adjust envprobe MSAA

brakhane:
This one I'm unsure and not really happy about; also I'm not familiar with DX12, so I just made some assumptions.

It turns out ImageMSArray is an optional feature for SPIR-V, and dxcompiler uses this to represent Texture2DMSArray, which is currently only used in aerialPerspectiveCS_capture_MSAA and volumetricCloud_renderCS_capture_MSAA.
Simply requiring that feature for Vulkan feels overkill, so instead I disable the MSAA mode when that feature is not present. (In a rather crude and hacky way) 

Not having the shaderStorageImageMultisample feature has additional restrictions (see https://registry.khronos.org/vulkan/specs/1.3-khr-extensions/html/vkspec.html#features-shaderStorageImageMultisample), so I added an assert to check for that.
I haven't been able to find an instance where CreateTexture gets called with UNORDERED_ACCESS *and* samples > 1. I had assume that would be the case whenever MSAA is enabled, but that doesn't seem to be the case.

From what I can tell, DX12 requires Texture2DMSArray to always be available, but I haven't found info on whatever the equivalent to setting VK_IMAGE_USAGE_STORAGE_BIT in DX12 is, and whether it guarantees that the samples can be larger than 1; nevertheless, I decided to just always enable the feature for DX12.

For the EnvProbe, I decided to do the "supports MSAA" check in the IsMSAA function, so that the "real" enabled value still gets serialized, to avoid unintentionally changing scene data when a scene is edited with a vulkan device that is missing the feature.

Let me know if I should go a different route.

---------

For SPIR-V, dxcompiler translates Texture2DMSArray into ImageMSArray. The latter is an optional feature for Vulkan, so we have to check it, and disable features that uses it. Currently, this is only the MSAA mode for environment probes.

For DX12, Texture2DMSArray is always available, so we enable that feature unconditionally there.

Further info:
https://registry.khronos.org/vulkan/specs/1.3-khr-extensions/html/vkspec.html#features-shaderStorageImageMultisample


turanszkij:
storageImageMultisample is not supported by DirectX and we don't use it anywhere.

WickedEngine/wiScene.cpp
```diff
@@ -3983,6 +3985,8 @@ namespace wi::scene
 			constexpr uint32_t blocks = envmapRes / GetFormatBlockSize(format);
 			constexpr uint32_t mip_count = GetMipCount(blocks, blocks);
 
+			EnvironmentProbeComponent::supports_MSAA = device->CheckCapability(GraphicsDeviceCapability::SHADER_STORAGE_IMAGE_MULTISAMPLE);
```
turanszkij, line 3988: This is for sure not good here, it's modifying a global from scene.Update() which can be called from any thread.

brakhane:
I'll see if I can change the shader code to avoid this whole mess.

brakhane:
closing this for now, no point in leaving it in draft forever

PR #723, opened 2023-08-02
use canonical cache directory on Linux

brakhane:


turanszkij:
What's the problem with what we have currently?

brakhane:
No problem per se, just best practise. There's probably a slightly faster startup after a reboot, since it can use the cache (most Linux distros delete /tmp on boot). Though the lion's share of startup time improvement is done by Mesa's shader cache, which is already in the correct directory.

portaloffreedom (approved):
from the [XDG documentation](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html):

 > $XDG_CACHE_HOME defines the base directory relative to which user-specific non-essential data files should be stored. If $XDG_CACHE_HOME is either not set or empty, a default equal to $HOME/.cache should be used.
 
I just wonder, do we really need the extra API? we could just return this folder when asking for a temporary directory path to use. Depends a bit on what type of temporary files we usually have (anything apart from shadercache?)

portaloffreedom:
Actually, since 
> If $XDG_CACHE_HOME is either not set **or empty**

we should probably test for that case as well

brakhane:
Good point, added a check for empty env

turanszkij:
Thank you for the review Matteo!

PR #725, opened 2023-08-05
fix using wrong function when calculating stage mask

brakhane:
This should be one of the last fixes, there is another small issue I'm currently working on, and I still need to figure out how to handle #722, but other than that, it looks good validation error wise.

As mentioned in the commit message, there were two instances where stageMasks were ORed with accessMasks:

```
barrierdesc.srcStageMask |= _ParseResourceState(ResourceState::RAYTRACING_ACCELERATION_STRUCTURE); 
barrierdesc.srcStageMask |= _ParseResourceState(ResourceState::PREDICATION);
```

I'm not 100% sure if srcAccessMask was supposed to be modified here or srcStageMask. Since `_ParseResourceState` does not handle `RAYTRACING_ACCELERATION_STRUCTURE` at all, I decided that those calls should have been to `_ConvertPipelineStage` instead, but I'm not 100% about this.

---------

Unlike VK_PIPELINE_STAGE_2_ALL_COMMANDS_BIT, the ALL_GRAPHICS_BIT specifies *all* pipeline stages, even unsupported ones. Since specifying unsupported stages is not allowed, that bit is useless unless all stages are actually supported by the device.

Hence, we need to calculate our own "ALL_USED_GRAPHICS_BIT".

Furthermore, this commit fixes an issue where the stageMask was accidentially ORed with accessFlags.

### DIFF
--- a/WickedEngine/wiGraphicsDevice_Vulkan.cpp
+++ b/WickedEngine/wiGraphicsDevice_Vulkan.cpp
@@ -405,7 +405,8 @@ namespace vulkan_internal
 
 		if (has_flag(value, ResourceState::SHADER_RESOURCE) ||
 			has_flag(value, ResourceState::SHADER_RESOURCE_COMPUTE) ||
-			has_flag(value, ResourceState::UNORDERED_ACCESS))
+			has_flag(value, ResourceState::UNORDERED_ACCESS) ||
+			has_flag(value, ResourceState::CONSTANT_BUFFER))
 		{
 			flags |= VK_PIPELINE_STAGE_2_ALL_COMMANDS_BIT;
 		}
@@ -434,10 +435,6 @@ namespace vulkan_internal
 		if (has_flag(value, ResourceState::INDEX_BUFFER))
 		{
 			flags |= VK_PIPELINE_STAGE_2_INDEX_INPUT_BIT;
-		}
-		if (has_flag(value, ResourceState::CONSTANT_BUFFER))
-		{
-			flags |= VK_PIPELINE_STAGE_2_ALL_COMMANDS_BIT;
 		}
 		if (has_flag(value, ResourceState::INDIRECT_ARGUMENT))
 		{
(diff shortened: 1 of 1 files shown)


turanszkij:
Are you sure there is an issue with `VK_PIPELINE_STAGE_2_ALL_GRAPHICS_BIT` ? Because the value it is not a bit combination of other stages, so it would make sense this is only includes supported ones on driver side. Are you getting validation error about this, can you show it? I will ask around about this.

turanszkij:
I confirmed with other Vulkan experts that `VK_PIPELINE_STAGE_2_ALL_GRAPHICS_BIT` means only the stages that are supported by implementation, so this shouldn't be a problem.

brakhane:
Hmm, it might very well be a bug in the validation layer. It complains with 
 
> [Vulkan Error]: Validation Error: [ VUID-VkBufferMemoryBarrier2-dstStageMask-03934 ] Object 0: handle = 0x55556e987c50, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0xc9fbbbdc | vkCmdPipelineBarrier2(): pDependencyInfo.pBufferMemoryBarriers[0].dstStageMask includes VK_PIPELINE_STAGE_MESH_SHADER_BIT_EXT when the device does not have meshShader feature enabled. The Vulkan spec states: If the meshShader feature is not enabled, pname:dstStageMask must not contain VK_PIPELINE_STAGE_2_MESH_SHADER_BIT_EXT (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkBufferMemoryBarrier2-dstStageMask-03934)

That and the specification saying: 

> VK_PIPELINE_STAGE_2_ALL_GRAPHICS_BIT specifies the *execution of all graphics pipeline stages*, and *is equivalent to the logical OR* of: [...]

vs

> VK_PIPELINE_STAGE_2_ALL_COMMANDS_BIT specifies all operations performed by all commands *supported on the queue* it is used with  

made me think it's intentional. Especially because the validation layer source has functions to limit ALL_COMMANDS_BIT down to what is supported.

However, I just found this comment in the source:

```cpp
 // Expand all pipeline stage bits. If queue_flags and disabled_feature_mask is provided, the expansion of ALL_COMMANDS_BIT
// and ALL_GRAPHICS_BIT will be limited to what is supported.
VkPipelineStageFlags2KHR ExpandPipelineStages(VkPipelineStageFlags2KHR stage_mask, VkQueueFlags queue_flags = kAllQueueTypes, const VkPipelineStageFlags2KHR disabled_feature_mask = 0);
```

but it's not called. I'll open an issue in the validation layer to ask for clarification. Either the comment is wrong, or the code is ;)

-----

For now, I will change the PR to only fix the incorrect function call, which I assume is indeed a bug?

turanszkij (approved):

turanszkij:
It wouldn't surprise me if there is a bug in validation layers, this is a common occurrence. The other fix is good.

PR #732, opened 2023-08-17
Various vulkan fixes

brakhane:
This PR contains a few bugfixes and cleanups.

Each change is its separate commit:

### 1. Don't check for VK_KHR_get_physical_device_properties2

It's part of Vulkan 1.1 core. Not important or anything. It just gets rid of a "best practise" warning. :shrug: 

### 2. Fix creating and destroying swapchains

We can't reuse a semaphore if it is signalled. There is no easy way to wait on the swapchain semaphore without some ugly hacks, so instead we just create a new one. We also need make sure we only destroy old swapchains once they are (probably) not used anymore. The recent `VK_EXT_swapchain_maintenance1` extension would make all of this much easier, but as of now, it's not yet widely supported.

This fixes the engine locking up on my Linux box after resizing the window.

### 3. reduce amount of copy-paste in AllocationHandler::Update()

Just a code cleanup. Using a lambda allows us to remove the copy-pasted loop and increase legibility by reducing the code by almost two thirds with no performance impact: both GCC and MSVC generate almost the same code as in the old version, the differences being often due to slightly different register allocations.

### 4. fix: ensure render area is not greater than color attachments

`swapchain->desc.width`/`height` can differ from the `swapChainExtent`, for example if a `VK_KHR_SUBOPTIMAL` happened before we received the "window size changed" event, or if it is not between the minimal or maximal extents. Ensure we don't use a render area larger than the color attachments, as this violates `VUID-VkRenderingInfo-pNext-06079` and `06080`.

### DIFF
--- a/WickedEngine/wiGraphicsDevice_Vulkan.h
+++ b/WickedEngine/wiGraphicsDevice_Vulkan.h
@@ -595,346 +595,104 @@ namespace wi::graphics
 			// Deferred destroy of resources that the GPU is already finished with:
 			void Update(uint64_t FRAMECOUNT, uint32_t BUFFERCOUNT)
 			{
-				destroylocker.lock();
+				const auto destroy = [&](auto&& queue, auto&& handler) {
+					while (!queue.empty()) {
+						if (queue.front().second + BUFFERCOUNT < FRAMECOUNT)
+						{
+							auto item = queue.front();
+							queue.pop_front();
+							handler(item.first);
+						}
+						else
+						{
+							break;
+						}
+					}
+				};
+
 				framecount = FRAMECOUNT;
-				while (!destroyer_allocations.empty())
-				{
-					if (destroyer_allocations.front().second + BUFFERCOUNT < FRAMECOUNT)
-					{
-						auto item = destroyer_allocations.front();
-						destroyer_allocations.pop_front();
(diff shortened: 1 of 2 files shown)


WickedEngine/wiGraphicsDevice_Vulkan.h
```diff
@@ -595,346 +595,105 @@ namespace wi::graphics
 			// Deferred destroy of resources that the GPU is already finished with:
 			void Update(uint64_t FRAMECOUNT, uint32_t BUFFERCOUNT)
 			{
-				destroylocker.lock();
+				const uint64_t DEADLINE = FRAMECOUNT - BUFFERCOUNT;
```
turanszkij, line 598: If FRAMECOUNT is less than BUFFERCOUNT this would cause negative uint64_t which will be a huge number. Then the `queue.front().second < DEADLINE` always passes and resource could be deleted while in use. I think this is a bug opportunity.

turanszkij (changes requested):

WickedEngine/wiGraphicsDevice_Vulkan.h
brakhane: Oops, you're absolutely right. I shouldn't do last minute "optimizations" late at night.

turanszkij (approved):

PR #739, opened 2023-08-31
PE: BugFix - Some of my scenes started to crash

plemsoft:
Some of my scenes started to crash when I tried to load them.
Not sure how the text got so big , but note the "...\Users\turan\..." at the end of the text ?


turanszkij:
Huh, that's weird. I think it would be better to find the root cause instead. Maybe one of the terrain models has an absolute path inside of it?

turanszkij:
Thanks, protection against writing outside memory will be in this pr: https://github.com/turanszkij/WickedEngine/pull/741
I will try to figure out why absolute path was not removed while saving asset, and how to fix it

turanszkij:
I think something happened with the terrain props assets at the time they were saved. The KTX textures contain the absolute paths as you noted on the video. Upon testing, KTX convert will no longer save with absolute paths. I went ahead and fixed the terrain props asset with removing absolute path. I think this will be correct. Fix is in https://github.com/turanszkij/WickedEngine/pull/741
Thank you for investigating

turanszkij:
This is fixed now

PR #740, opened 2023-09-01
Also reset the mouse delta position each frame

dpalais:
Reset to zero the SDL mouse delta position data for each frame.

### DIFF
--- a/WickedEngine/wiSDLInput.cpp
+++ b/WickedEngine/wiSDLInput.cpp
@@ -45,6 +45,7 @@ namespace wi::input::sdlinput
     void Update()
     {
         mouse.delta_wheel = 0; // Do not accumulate mouse wheel motion delta
+        mouse.delta_position = XMFLOAT2(0, 0); // Do not accumulate mouse position delta
 
         for(auto& event : events){
             switch(event.type){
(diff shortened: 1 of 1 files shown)


turanszkij:
What do you think about this @megumumpkin ? Was there an issue with mouse delta with SDL?

dpalais:
Indeed, mouse delta position will always be non-null and repeated until the next mouse movement in a different direction. It's always cleared in WickedEngine for every frame but then it's copied from the current SDL state which is never nullified.

Jontte:
In addition to this change I recommend we also enable accumulation of motion deltas that are being processed during the same frame on lines 79: 
```
-                    mouse.delta_position.x = event.motion.xrel;
-                    mouse.delta_position.y = event.motion.yrel;
+                    mouse.delta_position.x += event.motion.xrel;
+                    mouse.delta_position.y += event.motion.yrel;
```
In SDL2 the relative mouse motion values are given relative to the last event, and without this change we miss many relative mouse motion events if they are handled during the same frame. In master, the mouse moves very slow for me and I think that's because we are missing a large portion of the relative mouse motion events. With this change the responsivity is much better.

dpalais:
Agreed. I have added the accumulation of motion delta to the PR.

turanszkij (approved):

PR #744, opened 2023-09-10
vulkan: workaround a crash during window resizing

Jontte:
Add a vkDeviceWaitIdle in the swapchain creation function

I experienced frequent crashes whenever I resized the app window in Linux. I'm new to Vulkan, but reading online I learned that one should wait for the Vulkan device to become idle before re-creating the swapchain. This change fixes the crashes for me, however I'm not sure if this is the best way to fix the issue given the larger architecture.

### DIFF
--- a/WickedEngine/wiGraphicsDevice_Vulkan.cpp
+++ b/WickedEngine/wiGraphicsDevice_Vulkan.cpp
@@ -1228,6 +1228,10 @@ namespace vulkan_internal
 		createInfo.clipped = VK_TRUE;
 		createInfo.oldSwapchain = internal_state->swapChain;
 
+		// An explicit wait is added as a workaround for crashes experienced on some systems when rapidly resizing the screen
+		res = vkDeviceWaitIdle(device);
+		assert(res == VK_SUCCESS);
+
 		res = vkCreateSwapchainKHR(device, &createInfo, nullptr, &internal_state->swapChain);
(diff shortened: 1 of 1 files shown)


turanszkij:
Where did you read that waiting idle is required when resizing swapchain?

Jontte:
This tutorial claims so:
https://vulkan-tutorial.com/Drawing_a_triangle/Swap_chain_recreation . How
much it applies to our case is unknown to me. All I know is that the patch
makes the engine usable for me.

pe 15. syysk. 2023 klo 13.22 Turánszki János ***@***.***>
kirjoitti:

turanszkij:
My understanding is that when using oldSwapchain of createinfo, we don't need to wait for idle. The vulkan tutorial that you linked doesn't use old swap chain, but recreates the swapchain from scratch.

Jontte:
What do you want to do with this patch? Ideally we'd find the root cause which my hunch says is a race condition somewhere in the Vulkan backend that this workaround manages to hide, but I don't have the resources to debug this further right now.

turanszkij:
I think this goes against vulkan spec, the oldSwapchain is specified in createswapchain exactly because it allows the old swapchain to retire in a delayed manner after it's no longer used. Also can't just use vkDeviceWaitIdle there without protecting each queue against multithread access, since this uses vkQueueWaitIdle internally for each queue.

PR #745, opened 2023-09-13
Fix for NaNs in normal for scene capsule collision.

MortimerSnerd:
This is for a problem I ran into with a character capsule, where I wasn't limiting the movement step size to be < capsule.radius.  Whenever there was an overlap >= the capsule radius, the reported normal would come back as a vector of NaN.   It comes down to how the normal is calculated in the case where the line segment that makes the "spine" of the capsule intersects one of the triangles.  In this special case, the reported depth is also potentially too low, which my commit also addresses.

I don't work often with collision code, and less with the XMVECTOR routines, so I'd be glad to hear if you have another way you want approach it, or if it's just a limitation that should be addressed by just keeping the step size < capsule.radius.

### DIFF
--- a/WickedEngine/wiScene.cpp
+++ b/WickedEngine/wiScene.cpp
@@ -5107,7 +5107,8 @@ namespace wi::scene
 					// Find the nearest feature on the triangle to the sphere.
 					XMVECTOR Dist = XMVector3Dot(XMVectorSubtract(Center, p0), N);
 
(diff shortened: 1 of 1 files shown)


turanszkij (approved):

turanszkij:
Thank you for the fix. You used XMVECTOR correctly I believe.

PR #767, opened 2023-10-22
Vulkan: removal of timeline semaphore

turanszkij:


portaloffreedom:
This fixes the GPU hang on my setup (Intel Arc 770 on Linux Mesa)

PR #771, opened 2023-10-29
Adding Wobble Animation for the scripting API

Ficelloo:
Hello, I think it would be cool to add this to the scripting api


turanszkij (changes requested):
Yes, it can be added, but since you are adding binding for the wobbleAnim.amount, you should name the lua function as `SetWobbleAnimAmount()` instead of `SetWobbleAnimStrength()`. You could also consider adding the binding for wobble anim speed while you're at it.

Ficelloo:
Should I make an exemple lua script for it ?

turanszkij:
> Should I make an exemple lua script for it ?

Thanks for making the changes. No need to make an example script for it, though if you want it is ok to create one.

turanszkij (approved):

Ficelloo:
Alright! then let's push the pr

PR #774, opened 2023-10-31
Added lua binding for FindAllEntities

ElCapor:
I Added a lua binding for the function Entity_FindAllEntities() , it simply returns a table in the form {entity1, entity2, ..} containing all the entities in the current scene.
I believe that it would benefit everyone :)

### DIFF
--- a/WickedEngine/wiScene_BindLua.cpp
+++ b/WickedEngine/wiScene_BindLua.cpp
@@ -709,11 +709,12 @@ int Scene_BindLua::FindAllEntities(lua_State* L)
 		
 	int idx = 1; // lua indexes start at 1
 
-	lua_newtable(L); // build the return table
+	lua_createtable(L, listOfAllEntities.size(), 0); // fixed size table
+	int entt_table = lua_gettop(L);
 	for (wi::ecs::Entity entity : listOfAllEntities)
 	{
 		wi::lua::SSetLongLong(L, entity);
-		lua_rawseti(L, -2, idx); // the table is now at -2 because we pushed the entity
+		lua_rawseti(L, entt_table, lua_Integer(idx));
 		++idx;
 	}
 	// our table should be already on the stack
--- a/WickedEngine/wiScene_BindLua.cpp
+++ b/WickedEngine/wiScene_BindLua.cpp
@@ -512,7 +512,7 @@ Luna<Scene_BindLua>::FunctionType Scene_BindLua::methods[] = {
 	lunamethod(Scene_BindLua, Merge),
 	lunamethod(Scene_BindLua, UpdateHierarchy),
 	lunamethod(Scene_BindLua, Intersects),
-	lunamethod(Scene_BindLua, Entity_FindAllEntities),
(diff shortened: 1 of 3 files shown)


ElCapor:
I could also add an example usage script for this , for example , for new users , how they could get all entities and get their names etc, it would have helped me starting out too i think

ElCapor:
I just rebased my branch so that it's up to date with master

WickedEngine/wiScene_BindLua.cpp
```diff
@@ -701,6 +702,35 @@ int Scene_BindLua::Merge(lua_State* L)
 	return 0;
 }
 
+int Scene_BindLua::Entity_FindAllEntities(lua_State* L)
```
turanszkij, line 705: The c++ name of this is just `FindAllEntities()` so it would be better if lua followed the same naming.

WickedEngine/wiScene_BindLua.cpp
```diff
@@ -701,6 +702,35 @@ int Scene_BindLua::Merge(lua_State* L)
...
+{
+
+	int argc = wi::lua::SGetArgCount(L);
+	if (argc > 0)
+	{
+		wi::lua::SError(L, "Scene::Entity_FindAllEntities() doesnt take any argument");
```
turanszkij, line 711: It's ok if you provide parameters to a function that doesn't take any, it shouldn't be an error. All lua bindings follow this convention currently.

WickedEngine/wiScene_BindLua.cpp
```diff
@@ -701,6 +702,35 @@ int Scene_BindLua::Merge(lua_State* L)
...
+		wi::unordered_set<wi::ecs::Entity> listOfAllEntities;
+		scene->FindAllEntities(listOfAllEntities);
+		
+		int idx = 1; // lua indexes start at 1
+
+		lua_newtable(L); // build the return table
+		for (wi::ecs::Entity entity : listOfAllEntities)
+		{
+			wi::lua::SSetLongLong(L, entity);
+			lua_rawseti(L, -2, idx); // the table is now at -2 because we pushed the entity
```
turanszkij, line 724: Why do you hardcode -2 here? Please see an example of populating a table in the function `Entity_GetNameArray()`, where idx of the table is queried by `lua_gettop`.

turanszkij (changes requested):

WickedEngine/wiScene_BindLua.cpp
ElCapor: it's okay then i will remove that part , but is there still a way to provide some sort of a warning ? for example wi::lua::SWarning("Using too much arguments") for example ?

WickedEngine/wiScene_BindLua.cpp
ElCapor: I think i've done this

turanszkij (approved):

turanszkij:
Thanks, this is merged!

PR #779, opened 2023-11-21
vulkan: add asserts to check sparse is supported when used

### LINKED ISSUE #764: Terrain crash on Linux
Adding the terrain from the editor enters a crash. Cause is Division by zero.
Crash is happening here:
https://github.com/turanszkij/WickedEngine/blob/master/WickedEngine/wiTerrain.cpp#L1210-L1211
```cpp
region.width = atlas.maps[map_type].texture.desc.width / atlas.maps[map_type].texture.sparse_properties->tile_width;
region.height = atlas.maps[map_type].texture.desc.height / atlas.maps[map_type].texture.sparse_properties->tile_height;
```

The debugger shows the sparse_properties to be zero:
```
(gdb) print sparse_properties
$1 = (const wi::graphics::SparseTextureProperties *) 0x5555572f85c8
(gdb) print *sparse_properties
$2 = {tile_width = 0, tile_height = 0, tile_depth = 0, total_tile_count = 16384, packed_mip_start = 0, packed_mip_count = 0, packed_mip_tile_offset = 0, packed_mip_tile_count = 0}
```

brakhane:
This doesn't solve #764, but should make it clearer what is going on when the terrain system is used without sparse buffer support.

### DIFF
--- a/WickedEngine/wiGraphicsDevice_Vulkan.cpp
+++ b/WickedEngine/wiGraphicsDevice_Vulkan.cpp
@@ -3816,6 +3816,7 @@ using namespace vulkan_internal;
 		}
 		else if (has_flag(desc->misc_flags, ResourceMiscFlag::SPARSE))
 		{
+			assert(CheckCapability(GraphicsDeviceCapability::SPARSE_BUFFER));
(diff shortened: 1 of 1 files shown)


turanszkij (approved):

turanszkij:
Looks good, thanks

PR #785, opened 2023-12-04
Improvements + shared texture support

amerkoleci:
- Update vulkan headers, volk and vma
- Update DirectX headers and D3D12.
- Add shared extensions and texture creation under Vulkan.
- Add shared texture handle creation under D3D12.


WickedEngine/wiGraphicsDevice_Vulkan.cpp
```diff
@@ -2504,6 +2506,15 @@ using namespace vulkan_internal;
 
 			const wi::vector<const char*> required_deviceExtensions = {
 				VK_KHR_SWAPCHAIN_EXTENSION_NAME,
+#if defined(VK_USE_PLATFORM_WIN32_KHR)
```
brakhane, line 2509: Do all current vulkan implementations support these? If not, I think it would be better to make those optional and use them when they are available.

amerkoleci: They should all be supported, although semaphore and fence can be ignored, as I don't actually use them

brakhane: So, according to a very short check on gpuinfo.org, most GPUs should support these extensions on Linux, however, eg. the VK_KHR_EXTERNAL_SEMAPHORE_WIN32 only 19.93% cards in the database support it, most of them seem to be llvmpipe or integrated or old cards, but I also found one entry for the [Radeon RX 6900 XT](https://vulkan.gpuinfo.org/listreports.php?devicename=AMD+Radeon+RX+6900+XT+%28RADV+NAVI21%29&platform=windows), which is a relatively new card; though a different entry for that card does support it. Weird.

brakhane: > They should all be supported, although semaphore and fence can be ignored, as I don't actually use them

I think if they aren't currently used, they shouldn't be in the requirements, we can always add them later if need be.

amerkoleci: > VK_KHR_EXTERNAL_SEMAPHORE_WIN32

Make sure to check under Windows, not all platforms :)

turanszkij:
I will test this in January.

turanszkij:
I'm merging this on second thought, and will test xbox dx12 later

Issue #788, opened 2023-12-23
Issues using WickedEngine as submodule


mlavik1:
Hi!
First of all, thanks for making this *amazing* project. It has been perfect for my needs so far :)

I'm trying to add WickedEngine as a git submodule, and using `add_subdirectory` in CMake to add it as a subproject that I can link directly to.
This is my project structure:
- MyProject
   - WickedEngine

However, I'm running into some issues due to the use of `CMAKE_SOURCE_DIR` in several CMakeLists.txt files.
Example log from my project:
```
 Error copying file (if different) from "/home/matias/Projects/citygml-viewer/WickedEngine/libdxcompiler.so" to "/home/matias/Projects/citygml-viewer/build/3rdparty/WickedEngine/WickedEngine".
```

`CMAKE_SOURCE_DIR` returns the path to the top level of the source tree, which in my case is the root directory of my project. I believe it might be better to avoid using `CMAKE_SOURCE_DIR` at all, since it adds some limitations.

I was able to solve this issue by setting a new `WICKED_ROOT_DIR` variable in the root CMakeLists.txt and using that instead of `CMAKE_SOURCE_DIR`. Another alternative would be to use relative paths of `CMAKE_CURRENT_SOURCE_DIR`.

I'll make a PR, and then you can decide if you want this change or not :)

By the way: To use use WickedEngine as a subproject I also had to copy the "shaders" directory from "WickedEngine/shaders" into "MyProject/WickedEngine/shaders". I guess making that customizable could also be something worth considering, though it's not that much of an issue.

turanszkij:
Hi, the shaders folder can be overridden if you wat to avoid copying. In your application you can call `wi::renderer::SetShaderSourcePath()` (location where .hlsl shader souce code files are located) and `wi::rederer::SetShaderPath()` (location where compiled shader binary .cso files are located after they are compiled).

I'm not sure about the Cmake stuff as I am personally not using it, but maybe @portaloffreedom can say something about it as it's his creation.

mlavik1:
Thanks for the info! I will try that 😁

And ok, I guess they can have a look at my auggested fix as well then: #789
But no hurry 🙂

Happy holidays!

mlavik1:
Thanks for merging! I'll close this issue then.

brakhane:
@mlavik1 A tip for the future: if you add "fixes #788" or "closes #788" in the commit message, it will automatically close the issue once it gets merge

mlavik1:
[...]

PR #789, opened 2023-12-23
Replaced CMAKE_SOURCE_DIR, so WickedEngine can be used as submodule

mlavik1:
Suggested fix for #788 

**Test case:**
1. Create a new CMake-based project
2. Add WickedEngine as a subdirectory
3. Add `add_subdirectory(WickedEngine)` to your CMakeLists.txt, and also link to WickedEngine (`target_link_libraries(MyProject WickedEngine)`)
4. Build your project.

Of course, feel free to reject if you can think of a better fix, or if you don't feel this is a necessary change. Though, I can imagine there are others out there who also wish to use WickedEngine like this :)


portaloffreedom:
I like the idea! I need first to test if this breaks the cmake install flow, otherwise it's good to go for me!

CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 cmake_minimum_required(VERSION 3.8)
 
+set(WICKED_ROOT_DIR ${CMAKE_CURRENT_SOURCE_DIR} CACHE INTERNAL "")
```
portaloffreedom, line 3: I don't think this needs to be a CACHE entry, as we don't want to expose this variable to the user configuring the project

portaloffreedom (changes requested):

mlavik1:
> I like the idea! I need first to test if this breaks the cmake install flow, otherwise it's good to go for me!

Ok, cool! I'll test that as well 😃 I suppose that shouldn't be affected, but it definitely needs testing😅

CMakeLists.txt
mlavik1: You're right. i originally did that to make it acceasable from parent directories, but realise now that's silly, since the path should be known anyway. I'll fix that today or tomorrow 😃

mlavik1:
Hi @portaloffreedom, I updated my PR. `WICKED_ROOT_DIR` is no longer a cached variable.
Feel free to let me know if there's anything else.
Have a great week! :)

turanszkij (approved):

PR #798, opened 2024-02-09
- Root Motion WIP

Almahmudrony:
- Root Motion For WI
Its not the best solution(Kinda Hacky) but it will work for general purpose. 
If someone is very good at math please contact me(almahmudrony@gmail.com)

- Rules to follow for perfect Root Motion Implementation
Root bone must be point upward(Like a default bone in Blender).
Root boot should touch the ground always.
When animating Character Root bone ROTATION can only be animated in UP axis(Ex. In Blender its Z Axis rotation).
When animating Character Root bone POSITION can be animated for XYZ, No problem.

- Application in Engine.
Enable RootMotion in AnimationComponent property window.
Set the RootBone(name should be same as you name in your desired 3D modelling software)
Attach a script component to AnimationComponent.
Do not rotate character(in which you will apply root deltas from LUA) other than Y Axis, 
[Example: Lua Code]

```
	local Tx,Ty,Tz = animation_comp.GetRootTranslation()
	character_transform.Translate(Vector(Tx,Ty,Tz))
	
	
	local Rx,Ry,Rz,Rw = animation_comp.GetRootRotation()
	character_transform.RotateQuaternion(Vector(Rx,Ry,Rz,Rw))
```
[ExampleProject: https://drive.google.com/file/d/14qRkTtcSYv6pTYCeCNbO9aEdA0X06wHc/view?usp=sharing ]

[Beast Character Copyright - Frictional games(Amnesia The Bunker.)]  
https://frictionalgames.com/general-faq/#mod-monetisation

### DIFF
--- a/WickedEngine/wiMath_BindLua.h
+++ b/WickedEngine/wiMath_BindLua.h
@@ -48,18 +48,19 @@ namespace wi::lua
 		int Add(lua_State* L);
 		int Subtract(lua_State* L);
 		int Lerp(lua_State* L);
-		int Rotate(lua_State* L);
 
 
-		int QuaternionInverse(lua_State* L);
 		int QuaternionNormalize(lua_State* L);
 		int QuaternionMultiply(lua_State* L);
 		int QuaternionFromRollPitchYaw(lua_State* L);
 		int QuaternionToRollPitchYaw(lua_State* L);
-		int QuaternionSlerp(lua_State* L);
 		int Slerp(lua_State* L);
 
 		int GetAngle(lua_State* L);
+
+		int QuaternionInverse(lua_State* L);
+		int Vector3Rotate(lua_State* L);
+		int Vector3InverseRotate(lua_State* L);
 
 		static void Bind();
 	};
@@ -108,6 +109,11 @@ namespace wi::lua
 		int Add(lua_State* L);
 		int Transpose(lua_State* L);
 		int Inverse(lua_State* L);
+
+		int GetForward(lua_State* L);
+		int GetUp(lua_State* L);
+		int GetRight(lua_State* L);
+		int GetRotateAroundLocalAxis(lua_State* L);
 
 		static void Bind();
 	};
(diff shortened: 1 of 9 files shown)


WickedEngine/wiScene_Serializers.cpp
```diff
@@ -1163,6 +1163,9 @@ namespace wi::scene
 				}
 			}
 		}
+
+		// Root Bone Name
+		SerializeEntity(archive, rootMotionBone, seri);
```
turanszkij, line 1168: If you serialize a new property, you need to create a new version for that component. It is done in wiScene.h, where the component type is registered on the scene, the last argument can be given a version number, that must be always increased by one when serializing something new, to keep backward compatibility.

Then in the serialization code, you check if the version number is high enough before serializing the property:
```
if(seri.GetVersion() < 2) // requires version 2 of AnimationComponent
{
  // Root Bone Name
  SerializeEntity(archive, rootMotionBone, seri);
}
```

WickedEngine/wiScene_BindLua.cpp
```diff
@@ -3814,6 +3826,36 @@ int AnimationComponent_BindLua::SetEnd(lua_State* L)
...
+
+int AnimationComponent_BindLua::RootMotionOff(lua_State* L)
+{
+	component->RootMotionOff();
+	return 0;
+}
+
+int AnimationComponent_BindLua::GetRootTranslation(lua_State* L)
+{
+	wi::lua::SSetFloat3(L, component->rootTranslationOffset);
```
turanszkij, line 3849: These should use Vector lua binding instead of pushing a number of floats. 
```
Luna<Vector_BindLua>::push(L, component->rootTranslationOffset);
```

WickedEngine/wiMath_BindLua.h
```diff
@@ -48,19 +48,20 @@ namespace wi::lua
 		int Add(lua_State* L);
 		int Subtract(lua_State* L);
 		int Lerp(lua_State* L);
-		int Rotate(lua_State* L);
```
turanszkij, line 51: Your merge was incorrect, it has overwritten some of the recent changes and some of my recently added lua bindings are deleted

turanszkij (changes requested):

PR #799, opened 2024-02-10
- Root Motion WIP [RESOLVED]

Almahmudrony:
- Root Motion For WI
Its not the best solution(Kinda Hacky) but it will work for general purpose. 
If someone is very good at math please contact me(almahmudrony@gmail.com)

- Rules to follow for perfect Root Motion Implementation
Root bone must be point upward(Like a default bone in Blender).
Root boot should touch the ground always.
When animating Character Root bone ROTATION can only be animated in UP axis(Ex. In Blender its Z Axis rotation).
When animating Character Root bone POSITION can be animated for XYZ, No problem.

- Application in Engine.
Enable RootMotion in AnimationComponent property window.
Set the RootBone(name should be same as you name in your desired 3D modelling software)
Attach a script component to AnimationComponent.
Do not rotate character(in which you will apply root deltas from LUA) other than Y Axis, 
[Example: Lua Code]

```
	local root_translation = animation_comp.GetRootTranslation()
	character_transform.Translate(root_translation)
	
	
	local root_rotation = animation_comp.GetRootRotation()
	character_transform.RotateQuaternion(root_rotation)
```
[ExampleProject: https://drive.google.com/file/d/12RkdRpw61AEEGzKy9VhtLlFU0EWWwVcR/view?usp=sharing ]

[Beast Character Copyright - Frictional games(Amnesia The Bunker.)]  
https://frictionalgames.com/general-faq/#mod-monetisation

### DIFF
--- a/Content/Documentation/ScriptingAPI-Documentation.md
+++ b/Content/Documentation/ScriptingAPI-Documentation.md
@@ -577,6 +577,7 @@ A four component floating point vector. Provides efficient calculations with SIM
 - Dot(Vector v1,v2) : Vector result
 - Cross(Vector v1,v2) : Vector result
 - Lerp(Vector v1,v2, float t) : Vector result
+- Rotate(Vector v1,quaternion) : Vector result -- rotates the first argument 3D vector with the second argument quaternion
 - QuaternionInverse(Vector quaternion) : Vector resultQuaternion
 - QuaternionMultiply(Vector quaternion1,quaternion2) : Vector resultQuaternion
 - QuaternionFromRollPitchYaw(Vector rotXYZ) : Vector resultQuaternion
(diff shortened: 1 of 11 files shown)


brakhane:
JFYI, you don't need to open a new PR for changes, you can just push to the same branch you made your PR from and the original one would get updated automatically.

Almahmudrony:
Thank you. For letting me know. I have learned it now.

turanszkij (approved):
I think this is good now, only minor issue I saw is that you copy the argument vector in 
```
static bool EntityExistsInList(Entity entity, std::vector<Entity> entityList)
```
instead of passing by reference, and anim window is doesn't fit all elements without a scrollbar, but I'll fix these in the next commit.
Thanks for your work.

Almahmudrony:
Thank you.

PR #806, opened 2024-02-27
work around MSVC compiler bug

brakhane:
Current MSVC version fails with an internal compiler error, moving the declaration of `shaderentity` outside of the loop helps preventing the crash.


turanszkij (approved):
Thank you for taking care of this!

PR #810, opened 2024-03-10
terrain props editor

Wc4ever:
Here is my first iteration of props editor.
Basiclly I tried to do `vector<Props>` inspector without any data structure changes.

It is works, but I don't like overall workflow with terrain props right now.

Main problem is how to change prop object. First thing it is not obvious that then you select object it is serialized and than original object is not engaged anymore. If you change original, nothing happend but it seems like should. So if you changed original object you should reselect it, feels not good. Second you need to select object component. Terrain generates alot of object components and the combobox is huge.

Let's discuss how to improve terrain props editing.


turanszkij:
Yeah I'm not sure, you're right selecting things in combo box is problematic when there are a lot of items to select from, but this method is also used at other places, so it is fine to me as long as there is nothing better. I think your implementation is pretty good and better than what we currently had. 

Changing the main object shouldn't change the other prop objects, but changing the mesh or material should take effect in those, because those are reused. This works with your changes too, so I don't think there is a problem.

I'll test it a bit more but I think it will be ok to merge in your changes. Let me know if you are finished with it.

Wc4ever:
Ok I see. I think I'm finished with that approach. If you have any suggestions on implementation details let me know.

turanszkij (approved):

turanszkij:
Thanks for your work!

PR #813, opened 2024-03-13
Window resize fix.

Wc4ever:
Noticed a bug then hdr is on. It happend when I tried to resize window.

Seems like `rendertarget` used for final composition never got resized since app initialization.

image


turanszkij (approved):

turanszkij:
This was a good finding, thanks for your work!

PR #827, opened 2024-04-17
Fix Intersect behavior to prevent indexing object vector past its capacity

rjpilgrim:
I ran into a bug when trying to remove entities from the scene which would cause the size of the aabb_objectsvector to be greater than the objects.components vector.

Adding this check prevents that from occurring.

### DIFF
--- a/WickedEngine/wiScene.cpp
+++ b/WickedEngine/wiScene.cpp
@@ -5065,7 +5065,8 @@ namespace wi::scene
 				const AABB& aabb = aabb_objects[objectIndex];
 				if (!ray.intersects(aabb) || (layerMask & aabb.layerMask) == 0)
 					continue;
-
+				if (objectIndex >= objects.GetCount())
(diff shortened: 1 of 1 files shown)


WickedEngine/wiScene.cpp
```diff
@@ -5065,7 +5065,8 @@ namespace wi::scene
 				const AABB& aabb = aabb_objects[objectIndex];
 				if (!ray.intersects(aabb) || (layerMask & aabb.layerMask) == 0)
 					continue;
-
+				if (objectIndex >= objects.GetCount())
+					continue;
```
turanszkij, line 5069: This is not very good, limit the loop count instead, or use break, also not needed to be after the intersect check.

turanszkij (changes requested):

rjpilgrim:
Closing because I am opening another pull request with this change included.

PR #828, opened 2024-04-18
Improve entity remove

rjpilgrim:
I was seeing massive drop in framerate when removing entities during gameplay. Obviously, it's possible to simply transform entities instead of removing them, but this is more difficult and potentially a memory bloat when spawning a lot of projectiles. 

I have gotten rid of the recursive call from Entity_Remove, instead traversing through the hierarchy and populating a vector of entities to remove. I also put the removal on a separate thread. Potentially a spin lock might be needed, but in my testing I did not see any issues with the current implementation.

In addition, I have included the check on the objects index in the Intersects function. This time I put it in the condition of the loop.

### DIFF
--- a/WickedEngine/wiECS.h
+++ b/WickedEngine/wiECS.h
@@ -112,7 +112,9 @@ namespace wi::ecs
 		virtual void Serialize(wi::Archive& archive, EntitySerializer& seri) = 0;
 		virtual void Component_Serialize(Entity entity, wi::Archive& archive, EntitySerializer& seri) = 0;
 		virtual void Remove(Entity entity) = 0;
+		virtual void Remove(wi::vector<Entity>& killEntities) = 0;
 		virtual void Remove_KeepSorted(Entity entity) = 0;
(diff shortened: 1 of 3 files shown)


turanszkij:
I don't like this, you didn't even wait for the job to finish, this might work in some cases, but it will fail if you access anything while removal job is still going. Why did you implement Remove functions again instead of calling the existing ones for each element? I think if you're not satisfied with the existing entity remove function, this threading can be implemented in your project instead.

rjpilgrim:
To address your points:

1. I don’t wait for the job because the point is to improve performance by allowing it to happen on a different thread.

2. I created a new remove overload because it is actually faster to nest the loops in the order that I have them. As an entity is discovered and deleted, the time spent iterating through subsequent component types is decreased.

3. Using the job system at a higher level is a good idea, but is currently not possible through the lua binding. I am calling the Entity_Remove from a lua script.

turanszkij:
It would be better to add a lua binding for async removal that can also be waited.

turanszkij:
Async removal added for lua in this PR: https://github.com/turanszkij/WickedEngine/pull/842

PR #833, opened 2024-04-30
Make dofile return results

EmptySetRBLX:
Currently dofile returns a PID value, but this seems to be a input argument into the function anyways.

Most Lua implementations for dofile return the results of the dofile execution. Only easyish way currently to get the results from executing another Lua script in a Lua script is something like

`
print(loadfile("../Content/scripts/exampleScript.lua")())
`

or a require call if you don't mind the caching, although I don't think you get some of the globals set in dofile if you do any of these alternatives

### DIFF
--- a/WickedEngine/wiLua.cpp
+++ b/WickedEngine/wiLua.cpp
@@ -107,12 +107,17 @@ namespace wi::lua
 				std::string command = std::string(filedata.begin(), filedata.end());
 				PID = AttachScriptParameters(command, filename, PID, customparameters_prepend, customparameters_append);
 
+				lua_settop(L, 0);
+
 				int status = luaL_loadstring(L, command.c_str());
 				if (status == 0)
 				{
 					status = lua_pcall(L, 0, LUA_MULTRET, 0);
-					auto return_PID = std::to_string(PID);
-					SSetString(L, return_PID);
+				}
+
+				if (status == 0)
+				{
+					return lua_gettop(L);
 				}
 				else
 				{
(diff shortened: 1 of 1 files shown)


turanszkij:
This would break anyone's script that relies on current behaviour of dofile. If you don't give it PID argument, it will generate one internally.

EmptySetRBLX:
@turanszkij 
Only 0 is ever generated as the PID in my testing unless explicitly passed

image

Should the auto generated PID be incrementing? Would adding the PID as the first output variable be acceptable to get this merged? That would handle current use cases if someone is relying on this PID return.

turanszkij:
@megumumpkin Do you have any thoughts about this? Asking because you created the script tracking PID for lua, looks like it always sets the PID for dofile to zero when you call it from Lua (in Internal_DoFile()), but it generates unique PID if you call it from C++ (in RunFile()). I personally never used the return value of dofile.

turanszkij:
@EmptySetRBLX You could add your dofile status as the second return value since lua supports multiple returns.

EmptySetRBLX:
RunFile does not pass an explicit pid, causing the value to default to the GeneratePID() function. I have changed the Lua side to replicate that behavior. If an explicit PID is not passed, it will now call the generation function. 

The PID is the first return value of dofile(), any values returned by the file itself are returned as additional values in a multireturn.

turanszkij (approved):

turanszkij:
Thanks, I think this should be good now.

PR #836, opened 2024-05-03
Make the file easier to see in error logs

EmptySetRBLX:
Small tweak that injects the file path as a comment at the top of Lua executions so that tracing Lua errors is a bit easier.

image

### DIFF
--- a/WickedEngine/wiLua.cpp
+++ b/WickedEngine/wiLua.cpp
@@ -76,7 +76,8 @@ namespace wi::lua
 		std::string filepath = filename;
 		std::replace(filepath.begin(), filepath.end(), '\\', '/');
 
-		std::string dynamic_inject = "local function script_file() return \"" + filepath + "\" end;";
+		std::string dynamic_inject = "--" + filepath + "\n";
+		dynamic_inject += "local function script_file() return \"" + filepath + "\" end;";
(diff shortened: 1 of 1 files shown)


WickedEngine/wiLua.cpp
```diff
@@ -76,7 +76,8 @@ namespace wi::lua
 		std::string filepath = filename;
 		std::replace(filepath.begin(), filepath.end(), '\\', '/');
 
-		std::string dynamic_inject = "local function script_file() return \"" + filepath + "\" end;";
+		std::string dynamic_inject = "--" + filepath + "\n";
```
turanszkij, line 79: I think the `\n` linebreak will be an issue, because it will offset line indicators of error messages compared to the real file, so you should instead just use `;` at the end of these injected statements.

turanszkij (changes requested):

WickedEngine/wiLua.cpp
EmptySetRBLX: Ahh, I have adjusted this so that it no longer messes with line count.

Also cleaned up error handling a little bit.

turanszkij (approved):

turanszkij:
Nice, thanks!

Issue #455, opened 2022-06-05
NeedHelp


cailuming:
Hi there, if I have a brick, and export it to wiscene, then how can I use it like a prefab, I have tried like this:

- 1.Load the wiscene.
- 2.Create several objectComponent
- 3.Find the meshID by FindEntityByName.
- 4.Assign the meshID to each Object->meshID

Here is the sample code:
 ```
        Entity brick = wi::scene::GetScene().Entity_CreateObject("Brick _"+GUID);
        ObjectComponent	*objectComponent = wi::scene::GetScene().objects.GetComponent(brick );
	objectComponent->meshID = GetScene().Entity_FindByName("Castle_Brick_Mesh");
	transformComponent = GetScene().transforms.GetComponent(brick );
        transformComponent->Translate(XMFLOAT3(1,2,3));
```
As a result ,I couldn't get what I want

**Comment by turanszkij** [comment]
You could use `scene.Entity_Duplicate()` for example. But there are other options, too, for example in the editor's terrain generator, I am loading 3 scenes (tree, rock, bush), then reuse their meshes manually, you could check that out. Or an other option, when you loaded a scene from file into a `wi::Archive`, you can serialize that into a `Scene` and use `scene.Merge()` to move your loaded scene into the main scene.

**Comment by cailuming** [comment]
![image](https://user-images.githubusercontent.com/19427259/172175641-7908f160-47bf-4254-ac1e-4206604eccf2.png)
```
Entity tempEntity = GetScene().Entity_FindByName("FortressGate.glb");

gate = GetScene().Entity_Duplicate(tempEntity);
GetScene().names.GetComponent(gate)->name = "Castle_Gate_" + GUID;
 
objectComponent = GetScene().objects.GetComponent(gate);
transformComponent = GetScene().transforms.GetComponent(gate);
```
it seems that when I copy the "FortressGate" which is an root node for both roof and building more than three times, I will crash, but two times will not.
![image](https://user-images.githubusercontent.com/19427259/172178651-6d20b578-defa-41a7-9202-2e0b6eba4b47.png)
the crash shoot
![image](https://user-images.githubusercontent.com/19427259/172179071-f8618a1e-c75d-4490-848f-b0339ca47bb8.png)
it's suspected that when the wiscene file has more than one mesh as children , it will crash , but works well if the scene file has only one mesh.
Here are the test scene files
[single.zip](https://github.com/turanszkij/WickedEngine/files/8845084/single.zip)
[...]

PR #841, opened 2024-05-09
Add Instantiate() method to Scene

matpx:
This adds a helper function to the Scene which allows to quickly duplicate and merge one scene into another. This enables using scenes as prefabs. Also see https://github.com/turanszkij/WickedEngine/issues/455#issuecomment-1146807159.


turanszkij (approved):

turanszkij:
I think this will be useful, thanks!

PR #865, opened 2024-06-19
add physics integration tests

JohnGert:
Wicked simple in-editor integration tests. 
Content/scripts/runtime_test/physics_integration_test.lua is the test-harness. It just runs tests sequentially and waits for the previous test to finish before the next begins. Only the tests listed in there will be run. At the end of all the tests, an aggregate is reported to the backlog (e.g "Tests passed:  <passedTestCount>, failed: <failedTestCount>".

The two test-cases included are for physics and verify that gravity is working correctly (might be useful for validating other physics engines). 
You can run the test-cases individually from the editor as well. They are each designed to tear-down after they run.

### DIFF
--- a/Content/scripts/runtime_test/gravity01.lua
+++ b/Content/scripts/runtime_test/gravity01.lua
@@ -1,0 +1,75 @@
+-- Integration test for physics
+
+local scene = GetScene()
+
+local function Test()
+	local self = {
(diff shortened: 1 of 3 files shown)


turanszkij (approved):

turanszkij:
It's pretty cool, thanks.

PR #879, opened 2024-07-06
QOL: Scene Composition Picker to reflect PickResult visually.

ThatPipelineGuy:
Heya,

If you read my commit message you'll see that the current implementation works most of the time, but breaks down in larger dense wiScene's, where I believe some of the issues come from collections/folders that may be expanded to show visually.

I'd like to flesh this out so that any emitter/entity and it's sub object, if exposed to scene can be selected, and the guide in hierarchy will be chosen inside the ComponentsWindow ComboBox.

Not sure what more is missing here, I can tidy it up further, but also looking for feedback.

Cheers!
Dimitri


turanszkij (changes requested):
This is a good idea, but as you said doesn't work reliably yet, so if you can improve it, would be appreciated.

ThatPipelineGuy:
@turanszkij - Fixes have been implemnted. Child objects have not been validated.

I'm having issues debugging if Sponza scene is highlighting top-level nodes or child objects. Perhaps you might see something I'm missing here?

Looking forward to your feedback.

turanszkij:
I think there is still some issue, it doesn't work correctly for me, only in simple scenes. It's not only when there are child objects, seems to happen in any large scene with a lot of entities. I haven't been debugging, just trying the build.

ThatPipelineGuy:
> I think there is still some issue, it doesn't work correctly for me, only in simple scenes. It's not only when there are child objects, seems to happen in any large scene with a lot of entities. I haven't been debugging, just trying the build.

Correct. I think this issue is larger at hand stemming from PickResult not attributing the selection if the entity isnt a mesh object.

This is a guess, but I don't think it relates to this implementation, right now.

Going to be a deep dive to debug..

turanszkij:
I don't see a problem with selection on the master branch. It is sometimes easy to miss seeing what's selected if the list is too long, because scrolling the list can be too fast.

ThatPipelineGuy:
> I don't see a problem with selection on the master branch. It is sometimes easy to miss seeing what's selected if the list is too long, because scrolling the list can be too fast.

I see what you're talking about. Not sure how I didn't see that.

So, I did some debugging today and I'm not quite able to rely on offset to get a approximate range on the selected object in the tree index, but I can bring it into view. But, if we updated the ComponentsBox to have a higher length limit (essentially rearranging some the sidebar layout)

I can essentially do a quick and dirty fix to get this to work for the meantime, and maybe plan for another overhaul of the UI akin to something similar to Unreal with their slate window management?

turanszkij:
The Entities panel can be resized, I don't think rearranging the layout is necessary. I think fixing the offsetting is needed here, probably there is some issue with computing the offset and scrollbar values, it's a bit complicated.

PR #880, opened 2024-07-07
Bicubic filtering with Catmull-Rom filter

rs189:


turanszkij:
I have problems with this, and I think this is not the right solution for your issue that you reported on discord. I think this only solves the blurry image for you because this forces sampling mipmap 0 opposed to the regular sample we use in the image shader. So for your issue it would be better to use an image without mipmaps or better mipmaps (if you use DDS texture, you get to control what is contained in your mipmaps).

The problem with this approach is that it greatly increases register usage for the common image rendering shader (this shader is used for all gui and sprite rendering):

- before: 24 VGPR
- after: 56 VGPR

Increased VGPR usage can limit shader performance. An other issue is that the Bicubic filtering function forces sample mipmap level to 0, but the sprite rendering should use the correct mipmap level because sprites can be drawn in 3D space too.

PR #882, opened 2024-07-11
ensure GetPositionOnSurface doesn't return an uninitialized value for softbodies

brakhane:
Clang was complaining that P was not always initialized to a value, and it has a point. For softbodies, P was not calculated.

I assume that the same calculation needs to be done than in the other cases, but I'm not sure, so this PR might be complete garbage.


turanszkij:
Oh, thanks this is a good catch. But for the soft bodies, the `P` final position doesn't need to be transformed by the world matrix because the `p0, p1, p2` are already in world space (only for soft bodies). Fixed in separate commit.

PR #887, opened 2024-07-17
cmake: replace explicit list of files with globs

brakhane:
This PR is due to a recent discussion in Discord. I'm in two minds about this one. It prevents the problem of missing *.cpp files in the CMakeLists, and makes it a bit clearer which files are used for what (especially the Utility folder had a nice gotcha that wasn't obvious from the old CMakeLists), but it is also recommended against by the CMake team, however, I think the problem they mention is currently more theoretical.

Note that keeping CMakeLists.txt in the Template folder as-is is intentional: there's a edge case when a *.c file is listed in the root directory, so to prevent people from ever falling into that trap, and to follow CMake recommendation to not use globs, I've left it as it. If people want, they can always modify the cmakefile in their program based on the template. 

-------

CMake recommends against doing this because some build systems might not support CONFIGURE_DEPENDS. But currently, the ones we care about (Make, Ninja) do, and if there's ever a problem, we can just start listing them again.

Using globs has the advantage that the linux build will not break when a file is added in VS and somebody forgets to add the file to cmakelists as well. I've also found some instances were some *.h files were not listed, which means they weren't copied when installing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.8)
+cmake_minimum_required(VERSION 3.12)
 
 set(WICKED_ROOT_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 
--- a/Editor/CMakeLists.txt
+++ b/Editor/CMakeLists.txt
@@ -1,55 +1,9 @@
+cmake_minimum_required(VERSION 3.12)
 
-set (SOURCE_FILES
-	main_${PLATFORM}.cpp
-	#$<$<STREQUAL:${PLATFORM},Windows>:App_${PLATFORM}.cpp>
-	AnimationWindow.cpp
-	CameraWindow.cpp
-	CameraComponentWindow.cpp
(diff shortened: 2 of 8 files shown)


brakhane:
forgot something, brb

portaloffreedom:
We need to set `cmake_minimum_required(VERSION 3.12)` for this feature. Let's set it in both the root cmakelists and in the wickedengine subfolder cmakelists.

brakhane:
ok, found the problem. volk.c is not compiled in the old version.

brakhane:
With all those special cases made explicit, I now think this is preferable to the old way of doing things; thinks like volk.c being included by volk.h was pretty subtle before

turanszkij (dismissed):

turanszkij:
It looks ok to me, but from the conversation I'm not sure if it should be merged now, or is there anything else remaining?

portaloffreedom (approved):
I quite like it. Think this is a great improvement

turanszkij (approved):

PR #900, opened 2024-07-22
Improve AnimationComponent and AnimationWindow

AttackButton:
I made some improvements to the animation component, starting with the AnimationComponent::Stop() function, so that it updates immediately without waiting for AnimationComponent::Play() to be called again.


turanszkij (changes requested):
This pull request breaks things in existing animations. I haven't investigated what causes it, but the fighting game script's animations break. The FPS sample (patreon only) also breaks in a way, characters can blow up when they are shot dead. Existing animation logic should not change.

turanszkij (approved):

Issue #901, opened 2024-07-24
TextInputField Input Loss When Hovering Over Entities Tree


AttackButton:
Starting to type in a TextInputField and then hovering the mouse over the Entities Tree area (without clicking) causes the input to be canceled, resulting in the loss of the entered text.

I'm not sure if this issue occurs with other widgets.

![input_cancel](https://github.com/user-attachments/assets/40d3fd14-45cc-46c7-b229-2f41bc199131)

turanszkij:
Thanks for reporting, it should be fixed.

PR #907, opened 2024-08-02
Prevent delete key from removing an entity when TextInputField is active

AttackButton:
It is similar to https://github.com/turanszkij/WickedEngine/issues/901. Fixed an issue where pressing the delete key would remove an entity if the mouse was hovering over the entity tree list, instead of deleting a character in the TextInputField.


turanszkij (approved):

turanszkij:
Thanks

PR #929, opened 2024-08-23
linux: set nice level for low priority jobs

brakhane:
There's no easy way to set the niceness level of a thread from another one, so I had to move the call to the thread itself.

POSIX defines the nice level as process wide, while Linux implements it as thread-wide. This could change in the future, but there's no good other way to change the priority of "normal" tasks.

We could schedule the low prio threads as "idle" threads, but those will then have really low priority.

The windows version uses LOWEST, which is "2 points below normal", IDLE seems to be "15 points below normal". Since Linux nice levels go from -20 (highest) to 19, i chose 3 (2 / 15 \* 20), which means that those will be considered around half as important as other threads. That value can of course be changed to say 5, which would make it around 1/3 as important (the factor is around 1.25^nice)

### DIFF
--- a/WickedEngine/wiJobSystem.cpp
+++ b/WickedEngine/wiJobSystem.cpp
@@ -14,6 +14,7 @@
 
 #ifdef PLATFORM_LINUX
 #include <pthread.h>
+#include <sys/resource.h>
 #endif // PLATFORM_LINUX
 
 #ifdef PLATFORM_PS5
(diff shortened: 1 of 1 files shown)


brakhane:
We could experiment with setting the scheduling policy to SCHED_BATCH, which is similar to SCHED_OTHER (the default that is used now), but assumes that the thread is CPU-intensive (which is correct in our case), giving those threads a slight penalty (in addition to the low prio one) when scheduling, basically prioritizing the main (input/display) thread.

turanszkij (approved):

turanszkij:
Thanks!

PR #931, opened 2024-08-27
Add functions for openVR integration

Cop46:


turanszkij:
Please rename things like GetGraphicCommandQueue to GetGraphicsCommandQueue? Or just GetCommandQueue(QUEUE_TYPE queue) where you can give the type as parameter? Same for GetGraphicFamilyIndex, etc.

turanszkij (approved):

Issue #938, opened 2024-09-03
[EDITOR] When clicking on items it slowly rotates


CitroenGames:
https://github.com/user-attachments/assets/60442413-9b15-4e70-8a5a-52220a6f4757

it only happens on the highlighted object

CitroenGames:
to add more info to this:

The whole level is 1 big mesh and then i rotated it 90° because it wasnt rated correctly and that causes the issue.

i imported it into blender rotated it there exported it and now it works fine but still wanted to report it.

Cop46:
Could you please try this fix here and report if it works ? Thanks :)
https://github.com/turanszkij/WickedEngine/issues/759#issuecomment-2284866411

CitroenGames:
> Could you please try this fix here and report if it works ? Thanks :) [#759 (comment)](https://github.com/turanszkij/WickedEngine/issues/759#issuecomment-2284866411)

oh hey thanks for mentioning that without it i wouldnt have found it. 
sadly your explaination is hard for me to follow could you maybe make a fork of this project and make the change? 
that would help a lot thanks!

Cop46:
You can get the file on my fork here :
https://github.com/Cop46/WickedEngine/blob/master/Editor/Translator.cpp

CitroenGames:
> You can get the file on my fork here : https://github.com/Cop46/WickedEngine/blob/master/Editor/Translator.cpp just replace with your file and rebuild to try

yep that seems to have fixed it for me thanks!

Cop46:
Thanks for your reply I have made a pull request to fix this bug :)

Issue #786, opened 2023-12-08
scale and rotation drifting


Enhex:
tried importing a model and its scale and rotation constantly change:
https://github.com/turanszkij/WickedEngine/assets/7088062/e55876de-cfe5-47fc-9039-135d33042cb2
(the change pauses are when the editor window isn't focused)

using v0.71.354 Linux release.

klaussilveira:
Does this happen with any model? Any specific format?

Issue #759, opened 2023-10-02
Transform component scale decreasing in editor with certain rotations.


MortimerSnerd:
I'm on commit bb519474cad797af78f53bbee622520efbb725f7 , Windows.

In the editor, if you rotate an object around the y-axis by 90 degrees, the X and Z scale displayed in the transform component box will slowly start to decrease.  I'm guessing it's some sort of numerical error that's happening when the transform is decomposed and recomposed for the panel display.  But I didn't look, I just ran into this fooling around on my lunch break.

To reproduce:

1) start editor
2) create a new Cube
3) Open transform panel for cube, put 90.0 as the Y rotation
4) X and Z of scale should start dropping.

turanszkij:
Yes, I was aware of this, I also think this is some precision issue with matrix decompose. Haven't found a better way to do the editor transforms yet.

PR #939, opened 2024-09-04
Update Translator.cpp

Cop46:
Fix for bug #938,#759,#786

### DIFF
--- a/Editor/Translator.cpp
+++ b/Editor/Translator.cpp
@@ -503,6 +503,14 @@ void Translator::Update(const CameraComponent& camera, const XMFLOAT4& currentMo
 			dragging = true;
 		}
 
+		if (isScalator || isRotator || isTranslator)
+		{
+			if (dragging)
+			{
+				PostTranslate();
+			}
+		}
+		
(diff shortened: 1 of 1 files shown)


brakhane:
> Fix for bug #938,#759,#786

I think GitHub will automatically close the PR only if you use the right magic words. I'll try it here, but it might be that it only counts if you change your description.

Fixes #938, Fixes #759, Fixes #786

Cop46:
I have changed title and description, this is ok now ?

brakhane:
Title shouldn't matter, just description is enough. It worked, the issues in question now mention they will get closed by this issue. Thanks.

brakhane:
I just checked. Title actually gets ignored, only description or commit message are checked: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

turanszkij:
It's not really a fix, but pushing the problem a bit further away, as the issue still remains after moving the translator. Though in practice it solves the issue when the translator is not being grabbed, so it's probably worth to merge.

PR #960, opened 2024-10-08
Infinite reverse Z

amerkoleci:
Add new XMMatrixPerspectiveInfiniteReverseZFovLH and usage for perspective projection.


turanszkij (changes requested):
This doesn't work, completely breaks rendering and also picking.

PR #962, opened 2024-10-11
Transition to SDL3

portaloffreedom:
SDL3 is not release yet, but has hit ABI freeze. Until it's officially released this PR should not be merged.


portaloffreedom:
Currently SDL3 is being installed the hard way (`git clone; make; sudo make install`), we need to come up with a plan for this to change. Possible alternatives:
- vendor sdl inside the project statically
- vendor sdl inside the project, giving a dynamic library
- wait for sdl to be available in the next ubuntu LTS, which seems to be the most common denominator for old distros. (probably going to take 2 years or more)

metaleap:
> wait for sdl to be available in the next ubuntu LTS, which seems to be the most common denominator for old distros. (probably going to take 2 years or more)

Another option is to "wait until SDL's Wikipedia page lists a `3.x` as 'Stable Release' instead of the current `2.x`" :grin: even libsdl.org lists 2.x as stable and 3.x as preview..

Early-adopting the bleeding edge is fine for self-contained solo indie game projects etc, but a many-users engine project like Wicked with almost as many WIP projects can let the-hot-new-thing settle into stabilization and even some maturation first  =)

Vendoring: intuitively but without quite knowing explicitly why, I'd be wary about vendoring something like SDL the same way it's currently only done with LUA, Jolt, FAudio etc. It's such a common "system-wide dep".. and if one's Wicked Game C++ app ever imports anything else that also references SDL but the system one (or its own static or shared vendored one... :face_with_spiral_eyes: ), even if only to "optionally support" it (say some immediate-mode-GUI lib or other, who knows), at best you have the whole codebase loaded twice in memory and at worst maybe people get link-time conflicts / version differences, duplicate symbols or who knows what build-time or run-time confusions?

james2doyle:
[Unvanquished](https://unvanquished.net/) just posted about their usage of SDL3 and how they are currently working with it:

> Since SDL3 is a bit too new for some releases, we distribute a prebuilt SDL3 in our build dependencies for people wanting to build the game themselves (our script for rebuilding SDL3 oneself is also provided). People wanting to rely on the system SDL3 when packaging the game for a Linux distribution can use the new `-DUSE_EXTERNAL_DEPS_LIBS=OFF` CMake option to not use the prebuilt SDL3 we provide.

You can read more in the blog post on their site: https://unvanquished.net/unvanquished-0-55-5-polishing-it/

vrguru:
Any news on this now when SDL3 is officially released?

portaloffreedom:
The work was already been finished if I recall correctly, but SDL3 was not present on all major distributions (including steam Linux containers). The situation has much improved since then, so we can consider it.

vrguru:
What is the chance for getting this into head in a near future? Or do I have to make the port locally? I would really appreciate it as I am now moving over to Linux and need SDL3 for headless and UI integration, which seems to be hard with SDL2.

vrguru:
Since SDL3 has now had its official release, I wanted to add a data point in favor of picking this back up: I independently built and ran WickedEngine's core engine (WickedEngine_ext_shaders, with
WICKED_EDITOR/WICKED_TESTS/WICKED_IMGUI_EXAMPLE/WICKED_LINUX_TEMPLATE off) against the officially released SDL3 on Ubuntu 26.04 — 0 errors, 0 warnings — via a similar set of SDL2→SDL3 API changes (wiPlatform.h,
 wiInput.cpp/wiSDLInput, wiHelper.cpp, wiGraphicsDevice_Vulkan.cpp, wiAudio/FAudio's SDL3 backend). 
It ran correctly end-to-end (window, input, audio, Vulkan swapchain) in a downstream application embedding WickedEngine, so the port itself looks solid against the released library, not just the pre-release ABI this PR originally targeted.

Happy to help test or contribute if it'd move this forward — this has clearly got real demand now that SDL3 is out.
Actually we really need support for SDL3 due to headless rendering and other stuff we are doing. Ping @turanszkij

brakhane:
I'm planning to pick this up, but first I want to make some small other changes to the current SDL2 codebase, since they conflict a bit. But if you want, you can also prepare a SDL3 PR

turanszkij:
I don't know about the SDL things much so anyone can pick it up and add support.

portaloffreedom:
Hi! Sorry for being a bit unresponsive. Feel free to pick this up. You probably want to upgrade the build to use the latest steam build image that includes the new SDL3 library.

PR #973, opened 2024-11-01
flush debug output streams immediately

k4ss4n:
using std::flush ensures debug logging even in case of unexpected termination:

"This manipulator may be used to produce an incomplete line of output immediately, e.g. when displaying output from a long-running process, logging activity of multiple threads or logging activity of a program that may crash unexpectedly."

https://en.cppreference.com/w/cpp/io/manip/flush

### DIFF
--- a/WickedEngine/wiHelper.cpp
+++ b/WickedEngine/wiHelper.cpp
@@ -1388,12 +1388,15 @@ namespace wi::helper
 		default:
 		case DebugLevel::Normal:
 			std::cout << str;
+			std::flush(std::cout);
 			break;
 		case DebugLevel::Warning:
 			std::clog << str;
+			std::flush(std::clog);
 			break;
 		case DebugLevel::Error:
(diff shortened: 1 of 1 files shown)


turanszkij (approved):

WickedEngine/wiHelper.cpp
```diff
@@ -1388,12 +1388,15 @@ namespace wi::helper
...
 			std::cout << str;
+			std::flush(std::cout);
 			break;
 		case DebugLevel::Warning:
 			std::clog << str;
+			std::flush(std::clog);
 			break;
 		case DebugLevel::Error:
 			std::cerr << str;
+			std::flush(std::cerr);
```
brakhane, line 1399: Just a side note: unlike clog and cout, cerr is unbuffered and implicitly flushed. So strictly speaking, this is not necessary. But it doesn't hurt unless we literally want to get every last bit of performance.

PR #987, opened 2024-12-11
C++ 20 changes 

CitroenGames:
These changes wont effect C++ 17 but will aid in the conversion to C++ 20


turanszkij:
I think there was some problem with generic_string that's why it was changed to u8string, but can't remember what.

CitroenGames:
well you are converting it from u8 string to normal string so unless you change the return type that wont effect anything

turanszkij:
I downloaded the build, it crashes outright.

CitroenGames:
hmm weird lemme look on my side

CitroenGames:
ah yes you are right! the editor crashes the template seems to work as usual let look into why this is happening and come up with a solution

CitroenGames:
i have come up with a solution but i think at this point its just better to decline this PR or upgrade to c++ 20

CitroenGames:
yea nah this PR is getting stupid

PR #990, opened 2024-12-13
make the build compatible with cpp20

turanszkij:


brakhane:
JFYI, on Linux clang with warnings enabled more changes need to be done since implicit `this` capture in `[=]` is deprecated, and warnings are configured as errors, you can see the changes here https://github.com/CitroenGames/WickedEngine/pull/2

turanszkij:
uh, in that case I prefer those warnings to be disabled

PR #995, opened 2024-12-22
get rid of compiler warning

brakhane:
clang 14 is unhappy about inconsistent override (Window has it, ContentBrowserWindow doesn't)


turanszkij (approved):

turanszkij:
Thanks!

PR #997, opened 2024-12-28
Fix IMGUI Demo with HDR10

dwsJason:
When using HDR10, a normalization pass must run over the display buffer.  IMGUI leaves the scissor rectangle in a random state, preventing this from happening correctly.  Fix by restoring the Scissor Rect back to the size of the viewport.  After IMGUI is done generating command lists.


turanszkij (approved):

turanszkij:
Nice find, thanks for the fix!

PR #1006, opened 2025-01-08
use ubuntu 24.04 to compile to use modern gcc

brakhane:


brakhane:
While this fixes the Linux build (it currently crashes because GCC 11 doesn't like the filename with the japanese characters), ubuntu-latest will refer to 24.04 after the 17th anyway, so I don't think hardcoding the version is worth it.

PR #790, opened 2024-01-03
WIP === Blender exporter plugin

portaloffreedom:
pybind11 required

PR #1010, opened 2025-01-11
cmake: allow static PIC builds with new optional WICKED_PIC_CODE

metaleap:
Previously, position-independent code builds were only possible with `WICKED_DYNAMIC_LIBRARY`, but not with statically-linkable library `.a` builds.

This PR introduces the new `WICKED_PIC_CODE` cmake option to allow for the latter. It defaults to `WICKED_DYNAMIC_LIBRARY` which defaults to `OFF`.

Tested locally `Debug` and `RelWithDebInfo` builds with and without the new option: all build fine and for all, the Editor worked fine, too.


CMakeLists.txt
```diff
@@ -30,6 +31,7 @@ endif()
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
+set(CMAKE_POSITION_INDEPENDENT_CODE WICKED_PIC_CODE)
```
brakhane, line 34: Nitpicking, but PIC code is redundant, like LED diode, PIN number, or SSD drive. I'd prefer WICKED_PIC

metaleap: True enough! ATM machine etc :grin: took it over from [@portaloffreedom's draft PR](https://github.com/turanszkij/WickedEngine/pull/790/files) but sure, can change it to WICKED_PIC before merging.

metaleap: Done — 3 commits now unfortunately, somehow couldn't squash them into 1 locally (might be the `master` merge in between), but I think GH has an option for that here during PR Merging.

WickedEngine/Utility/FAudio/CMakeLists.txt
```diff
@@ -90,7 +90,7 @@ set_target_properties(FAudio PROPERTIES
 	OUTPUT_NAME "FAudio"
 	VERSION ${LIB_VERSION}
 	SOVERSION ${LIB_MAJOR_VERSION}
-	POSITION_INDEPENDENT_CODE ${WICKED_DYNAMIC_LIBRARY}
```
brakhane, line 93: Since this CMakeLists comes from FAudio, it will be overwritten. I think having this line here is an oversight, which we can fix now. Doing the same thing you did for Jolt will probably be the best solution

brakhane (changes requested):
see my comment in the FAudio CMakeLists, I missed that

WickedEngine/Utility/FAudio/CMakeLists.txt
metaleap: **Re Jolt:** look again, there's no changes to `Jolt/Jolt.cmake` — the Jolt-specific is in `WickedEngine/CMakeLists.txt`.

**Re FAudio:** this was an intentional oversight I should have elaborated on in PR text: check out the ca. dozen lines right above the diff line, lines 76 - 88 in this here changeset. It's all Wicked-specific that would be overwritten and would have to be factored out, but I'm not quite CMake-versed enough to confidently know whether they can just be pasted up into the top-level `CMakeLists.txt`, from the looks of it I kinda doubt it. Two courses of action:

- We make a TODO-type Issue to postpone this FAudio-only refactoring as a separate action item.
- We solve it now & here, in which case: what do you think @brakhane and/or @portaloffreedom as to what that kinda change should/would have to look like, such that there's no Wicked-specific stuff left in FAudio's `CMakeLists.txt`? (It's only all the stuff from L76 until the diff line that's pertinent in that file, already checked that.)

CMakeLists.txt
brakhane: What works best for me is: rebasing to master (not merging), and amending the original commit with the changes. You can then force push and GitHub will update the PR while keeping the old commit for reference.

brakhane: but I can squash this one, it's just an FYI for the future :)

WickedEngine/Utility/FAudio/CMakeLists.txt
brakhane: Oh yeah, I see. We should definitely try to fix that some time. But can't we just move this line out and have it be consistent with Jolt? Then it would be one thing less we'll have to fix.

I'm no CMake expert either, but I think setting the properties in the main cmake file just like you did with Jolt will have the same result as setting it here.

metaleap: Ack, done!

brakhane (approved):

PR #1012, opened 2025-01-11
use Zstd for wiHelper::Compress

brakhane:
ZStd has better compression and is faster as well:

Before (Deflate):

Total compressed: 48927172 -> 12048973 (24%)                                                                    
[Wicked Engine Offline Shader Compiler] ShaderDump written to wiShaderDump.h in 8.864 seconds                  

Now:

Total compressed: 48927172 -> 8954159 (18%)                                                                     
[Wicked Engine Offline Shader Compiler] ShaderDump written to wiShaderDump.h in 0.7831 seconds             

(It's also possible to use a higher compression level than 3, but 3 usually works well enough. Just for fun, I tested with the extreme setting 19, it reduced it to 16%, but took a whopping 21 seconds.)

### DIFF
--- a/WickedEngine/offlineshadercompiler.cpp
+++ b/WickedEngine/offlineshadercompiler.cpp
@@ -616,6 +616,8 @@ int main(int argc, char* argv[])
 	{
 		std::cout << "[Wicked Engine Offline Shader Compiler] Creating ShaderDump...\n";
 		timer.record();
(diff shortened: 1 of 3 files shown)


turanszkij (approved):

turanszkij:
That's great, thanks!

PR #1020, opened 2025-01-13
linux: add simple crash handler that dumps stacktrace

brakhane:
Since the Linux version seems much more unstable than the Windows one, having more information where the crash occured will help with figuring out where the problems are.

This commit adds a simple crash handler that dumps out the stacktrace, it's only supported on Linux.

### DIFF
--- a/Editor/CMakeLists.txt
+++ b/Editor/CMakeLists.txt
@@ -26,6 +26,12 @@ else ()
 	)
 	set(LIB_DXCOMPILER "libdxcompiler.so")
 
+	# needed for proper names in crash stacktrace
+	set_target_properties(WickedEngineEditor
+		PROPERTIES
+		ENABLE_EXPORTS ON
+	)
+
 endif ()
 
 # Copy content to build folder:
--- a/Editor/main_SDL2.cpp
+++ b/Editor/main_SDL2.cpp
@@ -6,6 +6,13 @@
 
 #include "icon.c"
 
+#ifdef __linux__
+#  include <execinfo.h>
+#  include <csignal>
+#  include <cstdio>
+#  include <unistd.h>
+#endif
+
 using namespace std;
 
 Editor editor;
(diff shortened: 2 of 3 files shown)


brakhane:
@turanszkij yes, it works in release mode as well.

Steam might be a problem because it usually doesn’t show the console. If that’s a concern I could also write it into a crash_log.txt file or something.

However, in the long term something like crashpad would probably be the best solution, as it could generate more data and would work cross platform.

turanszkij:
I was only concerned if it will even start up on non-development machine, if it does then it's ok by me. I think it would be ok if it logged to the backlog in that case so it gets logged with other things.

turanszkij:
I think logging into the backlog after the console is a good idea.

brakhane:
Added logging into backlog. I had to modify post() so that you can turn off DebugOut, otherwise the error would be shown twice.

turanszkij:
It's better to not modify engine's interface just for this. Why the backlog's DebugOut is not good to begin with since it will post on the console with std::cout/cerr?

brakhane:
It just looks ugly. Writing to the backlog works, the error message and stacktrace appear twice. Not printing out the stacktrace without going through backlog::post is risky, because it could fail.

I agree that "it looks ugly" is not a good reason to significantly modify an API. I'm not happy with it either.

Maybe we just accept that a crash looks ugly?

brakhane:
Alternatively, I could write to a crash_log.txt file. I can't write to log.txt because IIRC backlog complete rewrites the file every time

turanszkij:
I see, the DebugOut inside backlog::post is not so trivial because it needs the locking. Write a crash_log.txt then instead.

brakhane:
OK, I'll go with the crash_log.txt; but the DebugOut wouldn't be a problem because the crash handler only works for Linux, and there wi::DebugOut is just writing to stderr.

brakhane:
Pushed a new version that writes to a file.
I've reverted the post() changes, but kept the removal of zero initializing the `text` buffer in wilog_level/messagebox. Since sprintf will add a trailing zero, filling the entire buffer with zeroes beforehand is just wasting CPU time.

brakhane:
ah, we can just install upx via chocolatey, that will make things easier

turanszkij (approved):

PR #1022, opened 2025-01-15
Compress Editor with UPX

brakhane:


turanszkij:
Can we not install upx on the windows github builder instead? I found this, maybe it could work: https://github.com/marketplace/actions/installs-upx

I don't like that there is now an additional package to download, it's confusing. But I checked the compressed version, I didn't notice worse startup speed, and the size did get much smaller.

Also, it doesn't work on latest ubuntu builder, or was that just a test to set it to a fixed version?

brakhane:
The temporary package will disappear after 1 day, but I think I can also delete it immediately. I saw that package, but I usually try to avoid third-party actions, but I can reevaluate it.

The ubuntu-24.04 was just a temporary change to test the newest UPX version, (ubuntu-latest currently still uses 22.04, which will change in a few days)

brakhane:
ah, we can just install upx via chocolatey, that will make things easier

brakhane:
Sure, I'll update it. As mentioned on Discord, I will disable UPX for build-pr to keep the PR builds as fast as possible; I could compress them with level 1, which is instantaneous, but installing the windows version seems to take a few minutes

brakhane:
OK, I've added it to default and nightly, but removed it from PR as mentioned.

If you prefer to also have it on PR builds, let me know.

turanszkij:
Yes it should be on all builds.

brakhane:
done

brakhane:
Just for completeness sake, I want to mention that UPX has another compression mode using LZMA, which is significantly slower to decompress but gives much better compression.

For the editor, it introduces a slight delay of around 0.5s before the application starts, it feels barely noticable. And the compressed size is down to 6.8 MB instead of 8.4MB.

If you want to try it, change `--best` to `--best --lzma`

PR #1024, opened 2025-01-15
Mini docs fix: get entity via GetEntity

metaleap:
:scroll: :ring_buoy:  =)


turanszkij (approved):

turanszkij:
Nice find, thank you! 🫡

PR #1029, opened 2025-01-17
Update FAudio to 25.01

brakhane:


portaloffreedom (approved):

WickedEngine/Utility/FAudio/CMakeLists.txt
```diff
@@ -69,10 +70,10 @@ add_library(FAudio STATIC
...
 	src/FAudio_operationset.c
 	src/FAudio_platform_sdl2.c
+	src/FAudio_platform_sdl3.c
 	src/FAudio_platform_win32.c
 	# Optional source files
 	src/XNA_Song.c
-	src/FAudio_gstreamer.c
 )
 
 install(FILES ${PUBLIC_HEADERS}
```
portaloffreedom, line 79: there is no need for the `PUBLIC_HEADERS` variable, we can use the folder location like the original CMake

WickedEngine/Utility/FAudio/CMakeLists.txt
```diff
@@ -1,19 +1,20 @@
 # CMake Project for FAudio
 # Written by @NeroBurner
-cmake_minimum_required(VERSION 3.5)
+cmake_minimum_required(VERSION 3.5...3.10)
```
portaloffreedom, line 3: why the `...3.10` ?
I thought it came from upstream, but you added this yourself instead. What is the reason? (and maybe add a comment for it for future developers)

portaloffreedom (changes requested):

WickedEngine/Utility/FAudio/CMakeLists.txt
brakhane: It get rids of a CMake compatibility warning (supporting pre-3.10 behaviour is deprecated. This syntax says it's compatible with 3.10) I can't remember why we changed it to 3.5 in the first place, though; but since our main CMakeLists depends on 3.12 anyway, it's more of a technicality. Maybe we should just change it to 3.12 as well and be done with it?

WickedEngine/Utility/FAudio/CMakeLists.txt
brakhane: Original doesn't have the `install(FILES ${PUBLIC_HEADERS} DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/WickedEngine/FAudio/")` though. Do we actually need this?

portaloffreedom: oh, in the original file I saw this
https://github.com/FNA-XNA/FAudio/blob/master/CMakeLists.txt#L329
```
# Public Headers...
install(
	DIRECTORY include/
	DESTINATION ${FAudio_INSTALL_INCLUDEDIR}
)
```
Which is saying every file in the include directory is installed in the folder

But I realized that we did change the destination, so we still need at least that modification.

WickedEngine/Utility/FAudio/CMakeLists.txt
portaloffreedom: Both are fine, whatever you feel like it's better

WickedEngine/Utility/FAudio/CMakeLists.txt
brakhane: I think I'll just go with 3.12, as in all other CMakeLists

portaloffreedom (approved):

PR #1033, opened 2025-01-18
Docs: clarify how to achieve looping audio playback

metaleap:
[Background](https://wickedengine.net/forum/topic/how-to-make-the-soundinstance-in-samples-tests-loop/) — note, if this is an FAudio-only issue (don't have a Windows around to check), then that makes this PR moot & void of course, and instead a new Issue is in order, unless fixed by #1029   =)


turanszkij (approved):

turanszkij:
Yes, SetLooped should be called before sound instance is created.

PR #1037, opened 2025-01-19
Add to README.md missing Linux dep

metaleap:
I was wondering why on my system, `SDL_ShowSimpleMessageBox` (and thus `wi::helper::messageBox`) was a total no-op returning `-1`. This had me perplexed since SDL's docs on the function state:

> On X11, SDL rolls its own dialog box with X11 primitives instead of a formal toolkit like GTK+ or Qt.

Well, `SDL_GetError` informed me that "zenity failed to start" and a quick check affirmed that this wasn't already installed on my KDE system, zenity being a GNOME project (and apparently SDL2 won't try for `kdialog` if no `zenity` :grin: — either of which I wouldn't call "X11 primitives" but rather "external-program dependencies").

Well not all Linux users are on Ubuntu or are GNOME-using, so in the interest of completeness, this nano PR  =)

It's kind of an optional rather than hard "dependency" this way, given that things keep working without message boxes — OTOH usually one does not want to miss them when they have sth to say.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ If you have questions or stuck, please use the `windows` communication channel o
 <img align="right" src="https://github.com/turanszkij/wickedengine-gifs/raw/main/character_grass.gif" width="320px"/>
 
 #### Linux
-To build the engine for Linux, use Cmake. You can find a sample build script for Ubuntu [here](.github/workflows/build.yml) (in the linux section). On the Linux operating system, you will need to ensure some additional dependencies are installed, such as Cmake (3.7 or newer), g++ compiler (C++ 17 compliant version) and SDL2. For Ubuntu 20.04, you can use the following commands to install dependencies:
+To build the engine for Linux, use Cmake. You can find a sample build script for Ubuntu [here](.github/workflows/build.yml) (in the linux section). On the Linux operating system, you will need to ensure some additional dependencies are installed, such as Cmake (3.7 or newer), g++ compiler (C++ 17 compliant version), zenity and SDL2. For Ubuntu 20.04, you can use the following commands to install dependencies:
 
 <img align="right" src="https://github.com/turanszkij/wickedengine-gifs/raw/main/inverse_kinematics.gif" width="320px"/>
 
--- a/README.md
+++ b/README.md
@@ -49,9 +49,11 @@ If you have questions or stuck, please use the `windows` communication channel o
 <img align="right" src="https://github.com/turanszkij/wickedengine-gifs/raw/main/character_grass.gif" width="320px"/>
 
 #### Linux
-To build the engine for Linux, use Cmake. You can find a sample build script for Ubuntu [here](.github/workflows/build.yml) (in the linux section). On the Linux operating system, you will need to ensure some additional dependencies are installed, such as Cmake (3.7 or newer), g++ compiler (C++ 17 compliant version), zenity and SDL2. For Ubuntu 20.04, you can use the following commands to install dependencies:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -49,7 +49,7 @@ If you have questions or stuck, please use the `windows` communication channel o
 <img align="right" src="https://github.com/turanszkij/wickedengine-gifs/raw/main/character_grass.gif" width="320px"/>
 
 #### Linux
-To build the engine for Linux, use Cmake. You can find a sample build script for Ubuntu [here](.github/workflows/build.yml) (in the linux section). On the Linux operating system, you will need to ensure some additional dependencies are installed, such as Cmake (3.7 or newer), g++ compiler (C++ 17 compliant version) and SDL2. For Ubuntu 20.04, you can use the following commands to install dependencies:
+To build the engine for Linux, use Cmake. You can find a sample build script for Ubuntu [here](.github/workflows/build.yml) (in the linux section). On the Linux operating system, you will need to ensure some additional dependencies are installed, such as Cmake (3.7 or newer), g++ compiler (C++ 17 compliant version), zenity and SDL2. For Ubuntu 20.04, you can use the following commands to install dependencies:
```
brakhane, line 52: As you said, zenity isn't required. It only seems to be required on KDE, and even then it's not a hard dependency.

I would at least add a note on why zenity is required. It seemingly comes out of the blue. [Add "refuses to elaborate, leaves" meme here]

metaleap: OK tweaked. Even though it's not really _that_ "optional", since _every_ user realistically would like to _not_ silently miss errors or warnings that Wicked wants to tell them. And it's not missing or needed "only on KDE", rather it's just not pre-installed on non-GNOME Linuxes, of which there are _plenty_ out there and all with non-zero user bases. But whatev, I tweaked it, so there :grin:

brakhane: are you sure about the zenity requirement? It seems to me that it is required for dialog boxes, but not message boxes. And the messagebox I see is definitely neither GTK nor Zenity:

grafik

metaleap: > are you sure about the zenity requirement? It seems to me that it is 

My surety merely stems from what `SDL_GetError` reported when my `SDL_ShowSimpleMessageBox` call returned `-1`. I was surprised myself.. given what they stated in their docs. Sometimes docs are wrong, sometimes error messages are wrong..

> Edit: oh, the "implementation" for wayland is calling zenity. kek. That's pretty stupid, I agree

Even odder since I'm on X11 :grin:

turanszkij:
Is there no way to do message box for linux that doesn't require an other dependency?

metaleap:
> Is there no way to do message box for linux that doesn't require an other dependency?

Nah, the kernel has no such syscalls on offer (and hence neither does libc) because any & all windowing systems, window managers, display managers and desktop shells are not its business but decidedly userland software  =)

Reusing the SDL function is really the best choice here if we don't want to fiddle with somehow invoking "either X11 or Wayland" APIs directly, linking those libs & includes & all that junk, for a mere messagebox. (All it is, is SDL2 being not as refined in its programs-search as Wicked's own cross-platform-file-dialogs `.h` is.) And also, for our Linux users who are well used to the Linux openness=fragmentedness status quo, stating such "scattered extra packages you'll want" will not be a headscratcher for them or confusion or annoyance, but simply a welcome "completeness of system-requirements information".

metaleap:
> it should work on default linux and only error on some other systems (if I understand correctly)

There _is_ no one "default Linux"  :grin:  but yeah whatev  :man_dancing:

PR #1046, opened 2025-01-23
offlineshadercompiler: shaderdump no longer implies rebuild

brakhane:
When testing the zstd vs zlib compression the constant rebuilding took a lot of time, so I thought it might be a good idea to not require a full rebuild when using shaderdump.

A few short tests indicate that it does indeed work correctly, though there might be some weird edge case I didn't consider.

Hopefully, this will allow us in the future to speed up builds of PRs even more, but currently I can't think of a reliable way, as two separate PRs might modify different shaders.

This PR doesn't affect the github builds, as those are using "rebuild" explicitly, and also start with a clean slate every time.


turanszkij (approved):

turanszkij:
Should be fine with your solution I think.

PR #1050, opened 2025-01-30
build with unoptimized shaders for testing

brakhane:
Provide a Linux build where shaders have been compiled without optimization. This can be used by testers to rule out problems caused by those.

I don't think it makes sense to provide these for "official" releases, we can always tell people to download the latest nightly build if needed.

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -94,6 +94,16 @@ jobs:
 
 
   linux:
+    name: linux ${{ matrix.shaders.postfix }}
+    strategy:
+      matrix:
+        shaders:
+          - type: optimized
+            sc_args:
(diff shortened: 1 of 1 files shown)


turanszkij:
This should be PR only, then they can still download them.

brakhane:
I'm not sure I understand correctly. Do you want the changes to be in build-pr, so that every PR in the future will have two builds, or do you want to keep this specific PR unmerged and as a "testing ground" so to speak?

brakhane:
i've changed it so that it's not in build-pr

turanszkij:
I meant to just keep this pr open as a testing ground. Thanks.

brakhane:
OK, I've marked it as draft so as not to confuse people

PR #1060, opened 2025-03-11
wiScene: support address sanitizer

brakhane:
MSVC, GCC and Clang all support Google's Address Sanitizer which is a useful tool when debugging memory issues (out-of-bounds, missing free, etc.)

While trying to debug the issues I have with miniaudio in the FPS sample, I stumbled upon an ASAN error because the "deinterlaced" collider structure does not leave poisoned memory areas inbetween, so we have to mark the region excempt from checks.

There probably are more instances, but for now the FPS sample works; if I stumble upon more, I will add them.

When the code is not compiled with ASAN support, the macro compiles into an empty statement.


brakhane:
should XBOX/PS5 not provide the sanitizer header, I can add an `#if __has_include` and define the macro to empty if needed.

brakhane:
Also I just realized technically we might need to poison the memory region once we free the memory again, but for now I won't bother with it.

turanszkij (approved):

PR #1068, opened 2025-03-17
Change Jolt physics structure for ragdoll development

nattydread69:
I want to be able to make more complex ragdolls in the future, in particular add feet and hands.
I have split the ragdoll into a separate class so we can give it more functionality.


turanszkij:
I don't want to expose the physics internals like this. This can be discussed in an issue, or you can modify your own version of the engine.

nattydread69:
OK I will modify my own.

turanszkij:
Yes I can add feet for ragdolls in the future.

turanszkij:
The ragdoll feet have been added btw.

nattydread69:
Thanks I'll take a look :)

PR #1070, opened 2025-03-18
Jolt: update to v5.3.0

brakhane:
Only did a quick test on Linux, but it seems to work without any changes needed to wiPhysics.

Edit:whoops. forget to update the visual studio solution, will do that later

Edit2: fixed Windows build and confirmed it works with the FPS sample


brakhane:
One of the changes of 5.3.0 include a custom UnorderedSet and UnorderedMap which replace std::unordered_set/map, maybe those could be made another compile time option for wi::unordered_map

turanszkij (approved):

turanszkij:
Thankx.

PR #1103, opened 2025-05-11
linux: improve crash handling

brakhane:
add infos about device and driver version to the crash log, as well as the backlog; so we have this information if people make bug reports.

Also replace exit() with abort(), because calling exit could cause problems because of exit handlers (I had in fact seen a memory exhaustion on my machine, and others have reported hard crashes after a crarsh), so usind abort is much safer.

### DIFF
--- a/Editor/main_SDL2.cpp
+++ b/Editor/main_SDL2.cpp
@@ -15,7 +15,19 @@
 
 using namespace std;
 
-Editor editor;
+class EditorWithDevInfo : public Editor
+{
+public:
+	const char* GetAdapterName() const
+	{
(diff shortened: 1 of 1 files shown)


turanszkij (approved):

brakhane:
the abort() change doesn't work as intended, I'll have to fix that first

brakhane:
ok, now the handler will only trigger the first time. this fixes it.

brakhane:
ok, this was the last change. @turanszkij you can merge it if you're happy with it.

turanszkij:
Thanks for the improvements.

PR #1104, opened 2025-05-11
linux: don't use getText() in crash handler

brakhane:
getText() allocates memory which is a big no no in a signal handler. Instead, use an internal function that calls us for each entry so we can write it to the crash log.

This is ugly, but I can't think of a better solution atm.

### DIFF
--- a/Editor/main_SDL2.cpp
+++ b/Editor/main_SDL2.cpp
@@ -168,8 +168,10 @@ void crash_handler(int sig)
 		fflush(logfile);
 		backtrace_symbols_fd(btbuf, size, fileno(logfile));
 		fputs("\nBacklog:\n", logfile);
-		fflush(logfile);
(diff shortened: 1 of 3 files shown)


brakhane:
> Why not a function in the backlog called write_to_file(dest) that writes into a file with fputs? 

I thought about that, but it would have to be "unsafe" as well, since it cannot use the mutex, so it would have to be a "don't use this in normal code" function as well. 

Also, making the function write to a file would make it only usable for that, and not for crash handlers that eg. send the data via HTTP, so I thought a lambda would be more flexible. But I can rewrite it to use a file handle if you don't like lambdas here.



(Another thing I've been thinking about is just using a `friend` declaration, but it seems at least in GCC, you cannot have a friend declaration of a non existent function, so I went with the current solution)

turanszkij:
It's fine then.

turanszkij (approved):

PR #1105, opened 2025-05-13
Removed deprecared <codecvt> string conversion

### LINKED ISSUE #1100: -Wno-deprecated-declarations needed as a compiler option (due to std::wstring_convert deprecation)
@portaloffreedom 

Issue Reproduction:
Compile with GCC 15

portaloffreedom:
Removed deprecated <codecvt> header and replaced with equivalent C functions.

codecvt implementation in the CPP standard library was deemed not good enough and therefore it was removed, pending new proposals to make it good. https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0618r0.html

Fixes #1100

### DIFF
--- a/WickedEngine/wiHelper.cpp
+++ b/WickedEngine/wiHelper.cpp
@@ -17,7 +17,6 @@
 #include <iomanip>
 #include <fstream>
 #include <sstream>
-#include <codecvt> // string conversion
 #include <filesystem>
 #include <vector>
 #include <iostream>
@@ -28,6 +27,8 @@
 #include <Psapi.h> // GetProcessMemoryInfo
 #include <Commdlg.h> // openfile
 #include <WinBase.h>
+#else
+#include <cwchar> // string conversion
 #endif // _WIN32
 
 #ifdef PLATFORM_LINUX
@@ -1395,8 +1396,12 @@ namespace wi::helper
 			MultiByteToWideChar(CP_UTF8, 0, from.c_str(), -1, &to[0], num);
 		}
 #else
-		std::wstring_convert<std::codecvt_utf8<wchar_t>> cv;
-		to = cv.from_bytes(from);
+		std::mbstate_t state = std::mbstate_t {};
+		const char* from_c = from.c_str();
+		const std::size_t len = 1 + std::mbsrtowcs(nullptr, &from_c, 0, &state);
+		std::vector<wchar_t> to_c(len);
+		std::mbsrtowcs(to_c.data(), &from_c, to_c.size(), &state);
+		to = to_c.data();
 #endif // _WIN32
 	}
 
@@ -1410,8 +1415,12 @@ namespace wi::helper
 			WideCharToMultiByte(CP_UTF8, 0, from.c_str(), -1, &to[0], num, NULL, NULL);
 		}
 #else
-		std::wstring_convert<std::codecvt_utf8<wchar_t>> cv;
-		to = cv.to_bytes(from);
+		std::mbstate_t state = std::mbstate_t {};
+		const wchar_t* from_c = from.c_str();
+		const std::size_t len = 1 + std::wcsrtombs(nullptr, &from_c, 0, &state);
+		std::vector<char> to_c(len);
+		std::wcsrtombs(to_c.data(), &from_c, to_c.size(), &state);
+		to = to_c.data();
 #endif // _WIN32
 	}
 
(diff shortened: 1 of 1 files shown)


portaloffreedom:
Ops, made a mistake. Size_t cannot store negative numbers. will fix it as soon as I'm back at it.

WickedEngine/wiHelper.cpp
```diff
@@ -1410,8 +1415,12 @@ namespace wi::helper
 			WideCharToMultiByte(CP_UTF8, 0, from.c_str(), -1, &to[0], num, NULL, NULL);
 		}
 #else
-		std::wstring_convert<std::codecvt_utf8<wchar_t>> cv;
-		to = cv.to_bytes(from);
+		std::mbstate_t state = std::mbstate_t {};
+		const wchar_t* from_c = from.c_str();
+		const std::size_t len = 1 + std::wcsrtombs(nullptr, &from_c, 0, &state);
+		std::vector<char> to_c(len);
+		std::wcsrtombs(to_c.data(), &from_c, to_c.size(), &state);
```
brakhane, line 1422: as discussed on discord, this fails if locale is not an UTF-8 locale, so we might need to use `setlocale`

WickedEngine/wiHelper.cpp
```diff
@@ -1410,8 +1415,12 @@ namespace wi::helper
 			WideCharToMultiByte(CP_UTF8, 0, from.c_str(), -1, &to[0], num, NULL, NULL);
 		}
 #else
-		std::wstring_convert<std::codecvt_utf8<wchar_t>> cv;
-		to = cv.to_bytes(from);
+		std::mbstate_t state = std::mbstate_t {};
+		const wchar_t* from_c = from.c_str();
+		const std::size_t len = 1 + std::wcsrtombs(nullptr, &from_c, 0, &state);
+		std::vector<char> to_c(len);
```
turanszkij, line 1421: it can directly write into the `to.data()` instead of using an extra std::vector and copy.

portaloffreedom: but that it's a problem if the to buffer is not sized or intialized.

portaloffreedom:
Fixed all of the issues, should be good now. Maybe not speedy because of the setlocale, but better start from a stable point and optimize it later.
Do you think I should restore the locale after the function has been called? Or do we assume that every function that requires a locale will set it before?

WickedEngine/wiHelper.cpp
```diff
@@ -1440,8 +1440,8 @@ namespace wi::helper
 #else
 		std::mbstate_t state = std::mbstate_t {};
 		const std::size_t len = std::mbsrtowcs(to, &from, dest_size_in_characters, &state);
-		if (len > 0) return len + 1; // string length + null terminating character
-		return len;
+		if (len == static_cast<std::size_t>(-1)) return len + 1; // string length + null terminating character
```
brakhane, line 1443: shouldn't that be `>=`?

brakhane:
Personally, I don't like that we need to rely on locale. Just manually parsing/writing UTF-8 is maybe better?

But as a stop gap this should be good enough

WickedEngine/wiHelper.cpp
```diff
@@ -1461,8 +1461,8 @@ namespace wi::helper
 #else
 		std::mbstate_t state = std::mbstate_t {};
 		const std::size_t len = std::wcsrtombs(to, &from, dest_size_in_characters, &state);
-		if (len > 0) return len + 1; // string length + null terminating character
-		return len;
+		if (len == static_cast<std::size_t>(-1)) return len + 1; // string length + null terminating character
```
brakhane, line 1464: same

WickedEngine/wiHelper.cpp
```diff
@@ -1398,10 +1398,10 @@ namespace wi::helper
 #else
 		std::mbstate_t state = std::mbstate_t {};
```
brakhane, line 1399: could just be `std::mbstate_t state{};`, but this is ok as well

brakhane:
I haven't tested it, but I'm pretty sure the setlocale call can fail on systems that doesn't have the en_US.utf-8 locale installed. Eg. Debian asks what locales should be used, and a German system admin might choose to only install C and de_DE.UTF-8 for example

portaloffreedom:
> but I'm pretty sure the setlocale call can fail on systems that doesn't have the en_US.utf-8 locale installed.
This is terrible news... we might be forced to not use the C library then.... What a nightmare

WickedEngine/wiHelper.cpp
portaloffreedom: not >= won't work because it's going to do a comparison between size_t and size_t which are unsigned. Why is it returning a negative number inside an unsigned? I have no idea, but it is certainly a debatable choice.

PR #1110, opened 2025-05-26
update DXC to 2025-02, adjust license

brakhane:
With the Feb 2025 release, MS open sourced the DXIL signing code, and the closed sourced (lib)dxil are no longer required to create valid shaders.

Hence, Wicked is now 100% open source, so adjust the license accordingly, and remove dxil from third_party_software.txt.

~~Unfortunately, the new compiler still seems to fail to create SPIR-V for some shaders, so until that is fixed this cannot be merged.~~


brakhane:
Thanks. FWIW, I had to manually reduce the number of jobs to 16, otherwise I would run out of memory. The whole max-id thing is also weird, because I verified the shaders with spirv-val, and no shader uses up to 0x3fffff. But we need to increase the limit while compiling anyway. It feels like the input dxc feeds into the spir-v tools is just very unoptimized and the spirv optimizer has a hard time.

PR #1111, opened 2025-05-30
vulkan: Workaround for NVidia driver bug when compiled with new DXC

brakhane:
Please test if it works for you, and also if it fixes the graphics bugs; they don't seem to happen on my machine, so I can't test.


brakhane:
Btw, I haven’t checked if `+ 0` instead if `1 *` also works. It might, and could be slightly faster.

turanszkij (approved):

turanszkij:
Funnily, the new DXC for May 2025 just came out that doesn't have this bug, so the workaround will be removed in this pr: https://github.com/turanszkij/WickedEngine/pull/1112

PR #1137, opened 2025-06-22
Added hotkey for renaming the selected

### LINKED ISSUE #1079: Editor QOL: Object renaming key
This might be up for debate, but in the context of editing in general which has a heavier association for the F2 key:

* taking a screenshot
* renaming something

It's my belief more people using the editor have instinctively hit F2 with the intent to rename something within the Wicked editor, and have a need for renaming things far more often than taking screenshots. I think it would be a handy shortcut to allow a quick replacement (or creation) of the name data for the Name component.

AnkeDoesCoding:
I couldn't set F2 as the rename key so i made it R instead. This will close #1079

### DIFF
--- a/Editor/ComponentsWindow.cpp
+++ b/Editor/ComponentsWindow.cpp
@@ -590,6 +590,16 @@ void ComponentsWindow::Update(float dt)
 {
 	animWnd.Update();
 	weatherWnd.Update();
+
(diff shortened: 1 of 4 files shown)


turanszkij:
Thanks, looks good, I'll add some improvements for F2 and some other things.

PR #1146, opened 2025-07-01
zstd: add license text, update to 1.5.7

brakhane:
I noticed that the license text for zstd was missing, so I added it. While I was at it, I noticed that Wicked still uses 1.4.9, so I updated it to 1.5.7, which has [some improvements to compression speed and bugfixes](https://github.com/facebook/zstd/releases/tag/v1.5.7)


brakhane:
weird. compiles locally. I'll have a look

PR #1148, opened 2025-07-04
reduce code duplication

brakhane:
use lambdas to reduce the amount of copy pasted code. This makes it easier to see what happens in each handler.

### DIFF
--- a/Editor/RigidBodyWindow.cpp
+++ b/Editor/RigidBodyWindow.cpp
@@ -25,6 +25,27 @@ void RigidBodyWindow::Create(EditorComponent* _editor)
 		editor->componentsWnd.RefreshEntityTree();
 	});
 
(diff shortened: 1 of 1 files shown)


brakhane:
For easier review I recommend activating "hide whitespace"

brakhane:
I couldn't find a std::function declaration that would be compatible with the lambdas without causing invalid memory, so I went back to using `auto`

turanszkij (approved):

turanszkij:
> I couldn't find a std::function declaration that would be compatible with the lambdas without causing invalid memory, so I went back to using `auto`

I always use auto for lambdas too, it's fine.

brakhane:
> > I couldn't find a std::function declaration that would be compatible with the lambdas without causing invalid memory, so I went back to using `auto`
> 
> I always use auto for lambdas too, it's fine.

OnClick etc. use std::function, so I wanted to be consistent. I don't quite get why it works for OnClick, but not for forEachSelectedPhysicsComponent. #justcppthings I guess

Thanks for the merge.

PR #1149, opened 2025-07-05
compress embedded Liberation Sans

brakhane:
This saves 200KB in the binary and 600KB in the sources.

Just an spontaneous idea I had when stumbling upon the huge header file again. I wouldn't mind if it doesn't get accepted; reducing binary size is just a pet peeve of mine.


turanszkij:
This is good thanks.

PR #1152, opened 2025-07-07
fix missing call in forEachSelectedWithRefresh/fixXYZ

brakhane:


brakhane:
will be replaced with something better

PR #1153, opened 2025-07-07
replace default reference captures, fix missing calls

brakhane:
The default captures made it easy to accidentally reference values stored on the stack, causing invalid accesses.

Replace the `forEachXXX` functions and calls to only reference `this` when necessary and explicitly add exceptions. This actually found a few invalid accesses. (Clang does warn about those, but GCC doesn't. Not sure about MSVC)


Editor/MetadataWindow.cpp
```diff
@@ -172,7 +172,7 @@ void MetadataWindow::RefreshEntries()
 		Entry& entry = entries.emplace_back();
 		entry.name.Create("");
 		entry.name.SetText(name);
-		entry.name.OnInputAccepted(forEachSelectedWithRefresh(name, [&] (auto metadata, auto args) {
+		entry.name.OnInputAccepted(forEachSelectedWithRefresh(name, [&name] (auto metadata, auto args) {
```
ricejasonf, line 175: I see that `name` is a `std::string&` which is owned by a `std::vector<std::string>` in `MetadataComponent`. Do we know that this vector is not inserted to after capturing the reference here?

turanszkij:
The lambdas could be moved to class members or functions so that they don't need to be captured, and only capture `this`.

brakhane:
> The lambdas could be moved to class members or functions so that they don't need to be captured, and only capture `this`.

IMO, lambdas are better. They are only used in a single method (I think once or twice a similar loop was needed in another method, but that was an exception), and exposing some minor implementation detail, even as a private method, feels wrong, since the method would only ever be called by `Create`. Furthermore, the compilers do a better job at inlining lambdas, producing code that is close to if it were done with preprocessor macros. The only potential advantage I see is that we might be able to create a `forEach` method in a `Component` superclass and use some template magic to avoid redefining the loop in every component.

I don't like the verboseness of adding the function name in the capture either, but it's only in one place and a price worthy to pay, IMO.

If stuff like `[this, forEachSelectedWithRefresh]` is really a no go, we could just write `[=]` instead. It will still use a reference for `this`, and even if someone adds code to the lambdas that would result in some more variables to be copied, it shouldn't be a big deal since that part of the editor is not that time critical. The user doesn't care if a button click takes 5 ms or 50 ms to process.

So my suggestion is that we either leave the code as-is, or just use `[=]` which is shorthand for `[=, this]`

Editor/MetadataWindow.cpp
brakhane: I'm pretty sure the only way `metadata->bool_values.names` gets changed is by user action, and `RefreshEntries` only gets called after a user action; so unless the user is The Flash, it can be considered constant during the execution of this method.

turanszkij:
I think it's fine now, I'll leave it up to you to merge it when you want.

PR #1165, opened 2025-07-12
fix clang build on windows

brakhane:


brakhane:
will be replaced with a new PR that refactors cmake a bit

PR #1167, opened 2025-07-13
cmake/vs: add build flags for using AVX2 etc.

brakhane:
Previously, FMA3 and F16C intrinsics were implicitly enabled when the Windows SDK was found, but those instructions were only used for DirectX math, which clang doesn't like (specifically inlining functions using a disabled instruction set)

So now, we just define some compile time options to enable/disable AVX2, AVX, AVX512 etc. and make the compiler use the code for everything, not just DirectX math. By default, AVX2 is used.

Furthermore, the CMakeLists files were slightly refactored to be a bit less messy.

The Visual Studio solution was changed to use AVX2 for all generated code.


turanszkij (changes requested):
We didn't use AVX2, and it shouldn't be changed by default.

brakhane:
Hmm, ok, I'll just move the defines without ifdefs so that they are always defined, SDK or not. Apparently, they will use whatever is available.

I'll also undo the solution change.

FWIW, we were already using AVX2 on Linux due to Jolt's defaults, so I think keeping AVX2 on by default is ok there, people can always change the USE_AVX2 CMake setting.

I'll also add a workaround for a GCC bug that is triggered by this change in this PR, even though it's not strictly related. I will ping you once I' done

brakhane:
OK, VS solution changes are removed. Now the only thing that isn't related to CMake is that the the defines in wiMath are now also used when the SDK is not available, like on Linux.

brakhane:
I had to make some changes to silence some warnings (removing an "extern" and initializing some values), please have a look at ef4023d96863184badd5c1239736ada2a7d4abe2 if those are ok.

I didn't want to make a separate PR for this.

turanszkij:
> OK, VS solution changes are removed. Now the only thing that isn't related to CMake is that the the defines in wiMath are now also used when the SDK is not available, like on Linux.

This is great thanks, I tried to do it but failed.

Editor/ModelImporter_GLTF.cpp
```diff
@@ -2227,7 +2227,7 @@ void Import_Extension_VRM(LoaderState& state)
 				for (size_t boneGroup_index = 0; boneGroup_index < boneGroups.ArrayLen(); ++boneGroup_index)
 				{
 					const auto& boneGroup = boneGroups.Get(int(boneGroup_index));
-					SpringComponent component;
+					SpringComponent component{};
```
turanszkij, line 2230: Was this added because there was uninitialized member in springcomponent? I added missing initialization to member, can this now be removed if that was the reason?

brakhane: it complained about entity not being set. I've updated the PR, it now defaults entity to -1 in the struct and removes the `{}`

brakhane:
whoops, sorry, didn't notice you had pushed changes to the branch. They are now back

turanszkij (approved):

PR #1170, opened 2025-07-13
github builds: on linux, use ninja and compile PRs with clang as well

brakhane:
PR builds will be compiled with both GCC and Clang to more easily spot errors.

There is still only one Linux artifact, the GCC one.

Also, the build system was changed from make to ninja, as it is faster.

build.yml and build-nightly.yml will still only use GCC. The only changes there are switching to ninja and adjusting the name of the github cache to be consitent with build-pr.

Also the Windows CCache binary was updated to the latest version


brakhane:
since the targets have been renamed, the "Require status checks to pass" setting for master would need to be updated after this PR is merged.

brakhane:
> I'd like if the name of builds don't change so that other settings and links don't need to be edited, just add a new linux (clang). And I don't see the ninja being faster so change back to default pls so it's consistent with the readme.

I've reverted the rename of the other targets, but for simplicity and consistency the linux targets are now `linux (gcc)` and `linux (clang)`. My thought process was since the clang target needs to be added to the "status checks" anyway (otherwise, what's the point?) you could change `linux` for `linux (gcc)` at the same time.

I can name the targets `linux` and `linux (clang)` if absolutely necessary, but I find them confusing. Let me know.

As for ninja, it is slightly faster (though I admit there is huge variation on how fast an action runs between runs, so a lot more tests would be needed to exclude randomness):

image

But it's not a hill I'm willing to die on, and the difference is not that huge, especially since windows builds still take up the most time, therefore I changed it back to make.

As a heads up, I have rebased the branch onto current master, so your merge is now gone and if you have the branch checked out locally, you'll need to do a `git reset --hard`

turanszkij (approved):

turanszkij:
Ok, I merged it like this. I don't want to update to a new builder if we don't have to, the build speeds are varying and I'd like to stick to the basics.

brakhane:
Thanks. One advantage of ninja would be that each compilation output is buffered and printed out together, unlike make where a parallel build can cause error/warning messages to interleave between files. But in practice that's never happened with Wicked, so I'm ok with sticking with make.

PR #1171, opened 2025-07-14
lua: update to 5.4.8

brakhane:
Previous version was 5.3.3 from 2016, so after 9 years an update doesn't hurt.

I don't see anything in [the incompabilities](https://lua.org/manual/5.4/manual.html#8) that would apply to Wicked, and at least character_controller and FPS shooter (Patreon version) seem to work like they did before.

[Main changes can be found here](https://lua.org/manual/5.4/readme.html#changes)


turanszkij (approved):

turanszkij:
I checked some scripts too, seems like they are ok.

PR #1175, opened 2025-07-17
github actions: cleanup runner cache on closed pull requests

brakhane:
The script was taken directly from [the official GitHub docs](https://docs.github.com/en/actions/how-tos/managing-workflow-runs-and-deployments/managing-workflow-runs/manage-caches#force-deleting-cache-entries), it deletes caches from PR that have been closed and are not useful anymore.

While GitHub will delete older caches automatically, if there are a lot of parallel PRs open, those could be enough to cause the "master" cache to be evicted, which is the only cross-PR cache, leading to longer build times. Also, it's nicer if we clean up garbage ourselves instead of relying on others to do it.

But this PR is completely optional, nothing will break without it (except maybe some longer compilation time in rare cases)

### DIFF
--- a/.github/workflows/cleanup-cache.yml
+++ b/.github/workflows/cleanup-cache.yml
@@ -1,0 +1,29 @@
+name: Cleanup github runner caches on closed pull requests
+on:
+  pull_request:
+    types:
+      - closed
+
+jobs:
+  cleanup:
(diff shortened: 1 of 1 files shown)


turanszkij:
I'd rather not add it, to reduce the number of things that can break.

PR #1185, opened 2025-07-26
cmake: only symlink when supported

brakhane:
Windows only supports symlinking as admin or in developer mode. While most devs will have the latter enabled, not everyone does, so it's better to handle that case instead of just failing.


turanszkij:
If there are two methods that can be chosen by cmake it will lead to confusing situation when it works one way for someone and a different way for someone else. I think we can rely on symlink.

brakhane:
*tl;dr: read the last paragraph*

if we always symlink people will not be able to compile it on windows without enabling developer mode, even worse, cmake will run successfully, but the build will fail mid way through.

CMake should check everything that's needed and configure the build files so that the build will succeed. If it can't do that, it should just outright fail in all cases where it can tell it advance (and we can in this case)

Also CMake projects often are compiled slightly differently depending on the environment, especially with all various linux flavours around, it's pretty normal for CMake to do thing A on one env, and thing B on another.

That being said, I understand your point that CMake shouldn't automatically fall back to copying.

I can add a setting `USE_SYMLINKS` or something that is on by default but can be overwritten by the user. If it is on, CMake will test if symlinks are working and if not, fail with a message to either enable dev mode or disable symlinking. This way, it's explicit. Would this be acceptable to you?

brakhane:
I've updated the PR to use a flag instead, I'm going to just quote the commit message here

>Windows only supports symlinking as admin or in developer mode.
    This commit adds a cmake flag to use symlinking instead of copying.
    When enabled, it will check for support and output a hint to enable
    developer mode when it fails.
>    
>By default, it is disabled on Windows, and enabled otherwise.

brakhane:
seriously github? fixing merge conflicts removes the pull approval?

brakhane:
@turanszkij you'll need to approve it again, github did a funny. (That's the last time I try to use the web interface to solve merge conflicts)

Or you can just merge it. Nothing has changed from the previous version, just rebased on top of master and resolved the weird merge conflict.

brakhane:
thanks

PR #1186, opened 2025-07-26
cmake: add support for unity builds

brakhane:
Not related to the game engine, unity builds are a cmake concept to combine the cpp source files into a few huge amalgamation, this increases memory usage but can speed up compilation time significantly, and potentially runtime as well since more functions might get inlined by the compiler.

On my laptop, without ccache, the compilation time went from 4:19 to 2:39, a 40% speed increase

It is best practise to let the person running cmake decide whether to use it or not, so this commit doesn't enable unity builds by defaults. It however tells CMake to compile a few files seperately that would run foul of the ODR if combined.


turanszkij:
Will this require extra maintenance when I add cpp include to some files in the future?

brakhane:
I think there are two parts to that question: "do I need to modify CMakeLists when I add an include-only cpp" and "will this PR increase my workload in the future"

The answer to the first is "you already do, but often it works even if you forget". Originally, the cmake files required every file to be compiled to be explicitly listed. That often broken when you added files and forced you to also modify the cmake files to get linux builds working again. Since in 99% of cases, a newly added file should just be another file to be compiled, we changed it so that all files are compiled by default and the exceptions are listed; this also made the exceptions a lot clearer, in the old version I never realized that volk.c was not compiled seperately, so all in all I think the new way is better.

If a new cpp file is included and you forget to add an exception, it will most likely still work, the file will be compiled in twice but it doesn't seem be that big of a problem in practice (for example currently D3D12MemAlloc.cpp is such an example due to an oversight, it gets included and also compiled; which this PR also fixes btw)

The answer to the second part is no. While including a new cpp file might indeed break unity builds, some other random changes might as well. Currently unity builds are broken without the PR because for example wiPrimitives_BindLUA refers to `Sphere` which when compile alone is unambigious, but in a unity build could refer to Jolt's or wicked's `Sphere`, that's why those files are excepted.

This PR adds some exceptions to the Unity Build process to make it work with the current code base. Unity builds might fail again in the future, but that's ok, they are a convenience, and I can fix them if needed; if they are broken for a few days, it's not a big deal. People can always deactivate them and recompile if necessary.


**tl;dr: this PR will not increase your workload. On the contrary, it might make it easier for me to spot that you've changed something that should be taken into account for "normal" non-unity cmake builds as well (the D3D12MemAlloc.cpp is a good example)**

turanszkij (approved):

PR #1187, opened 2025-07-28
configure DXMath via compile defs when using CMake, F16C support for Jolt

brakhane:
*see current patch for commit message*

### DIFF
--- a/WickedEngine/CMakeLists.txt
+++ b/WickedEngine/CMakeLists.txt
@@ -12,7 +12,7 @@ set(INSTALL_CONFIGDIR "${CMAKE_INSTALL_LIBDIR}/cmake/WickedEngine")
 option(USE_SSE4_1 "Enable SSE4.1" ON)
 option(USE_SSE4_2 "Enable SSE4.2" ON)
 option(USE_AVX "Enable AVX" ON)
-option(USE_AVX2 "Enable AVX2" ON)
+option(USE_AVX2 "Enable AVX2" OFF)
 option(USE_AVX512 "Enable AVX512" OFF)
 option(USE_LZCNT "Enable LZCNT" ON)
 option(USE_TZCNT "Enable TZCNT" ON)
@@ -64,7 +64,6 @@ endif()
 add_subdirectory(LUA)
 add_subdirectory(Utility)
 
-add_compile_definitions(JPH_DEBUG_RENDERER=1)
 set(PHYSICS_REPO_ROOT ${CMAKE_CURRENT_SOURCE_DIR})
 include(${PHYSICS_REPO_ROOT}/Jolt/Jolt.cmake)
 if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
@@ -75,6 +74,7 @@ endif()
 set_target_properties(Jolt PROPERTIES
 	POSITION_INDEPENDENT_CODE ${WICKED_PIC}
 )
+add_compile_definitions(Jolt JPH_DEBUG_RENDERER)
 
 file(GLOB HEADER_FILES CONFIGURE_DEPENDS *.h)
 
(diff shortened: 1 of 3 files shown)


WickedEngine/wiPhysics_Jolt.cpp
```diff
@@ -7,6 +7,34 @@
 #include "wiRenderer.h"
 #include "wiTimer.h"
 
+#ifdef WICKED_USE_AVX512
+  #define JPH_USE_AVX512 1
```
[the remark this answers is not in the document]
brakhane: Good point.

I think we'll have to remove those lines then and define the JPH_USE macros alongside the wicked ones. That's automatically done in CMake, but for VS it would have to be done manually.

Either that, or we modify Jolt/Core.h to add that stuff there; or reuse the JPH defines for Wicked as well. Neither does sound particular appealing

WickedEngine/WickedEngine_Windows.vcxproj
```diff
@@ -62,7 +62,7 @@
       </PrecompiledHeader>
       <WarningLevel>Level3</WarningLevel>
       <Optimization>Disabled</Optimization>
-      <PreprocessorDefinitions>_DEBUG;_LIB;%(PreprocessorDefinitions);JPH_DEBUG_RENDERER</PreprocessorDefinitions>
+      <PreprocessorDefinitions>_DEBUG;_LIB;%(PreprocessorDefinitions);JPH_DEBUG_RENDERER;WICKED_USE_F16C;JPH_USE_F16C;WICKED_USE_FMADD;JPH_USE_FMADD</PreprocessorDefinitions>
```
turanszkij, line 65: I want to remove the WICKED_USE_ defines, the JPH defines can stay here.

WickedEngine/wiMath.h
```diff
@@ -5,8 +5,21 @@
 #include <algorithm>
 #include <limits>
 
-#define _XM_FMA3_INTRINSICS_
```
turanszkij, line 8: I want to keep these switchable only here like it was before and not use the WICKED_USE defines.

WickedEngine/CMakeLists.txt
```diff
@@ -103,6 +103,34 @@ set(WICKEDENGINE_STATIC_LIBRARIES
 	Utility
 )
 
+if (USE_AVX512)
```
turanszkij, line 106: I'd like to have JPH defines here hard coded instead of defines depending on other defines.

brakhane: there's little need, as the JPH will already be defined when compiling Jolt.

It would only be needed if wiMath sets _XM_ depending on JPH_ flags, which as mentioned above is weird; alternatively, these can define the _XM_ flags, this way, nothing is needed in wiMath

brakhane:
> I'd like to keep the defines in wiMath.h like it was before and only set JPH defines in project/cmake.

I dislike that solution because it prevents people from compiling the application to their needs. I'm fine with setting those by default (in fact, that's how it is currently), but people should be able to disable F16C or FMA if they want to without needing to change the code, especially since Jolt already allows to configure it when using CMake and those flags are exposed to the main project.

If you want to keep the defines, how about a compromise? CMake based builds will define a macro like `CMAKE_BUILD` or something, and wiMath will look something like

```cpp
#ifndef CMAKE_BUILD
#define _XM_F16C_INTRINSICS_
#define _XM_FMA3_INTRINSICS_
#endif
```

I'm not sure if I understand the second paragraph. Currently, Jolt will not use F16C because of the missing defines, do you not want to set the defines in the solution file?

turanszkij:
What I want:

- XM defines should be defined just in code in wiMath.h because it's our code, defining it outside is unnecessary
- JPH defines has to be defined in solution/cmake unfortunately because we don't modify jolt code

Whatever we can we define in just the code, what we can't then it can be defined in project/cmake

brakhane:
The problem with hardcoding the defines is that people cannot turn F16C off without modifying the code, which kinda goes against CMake's philosophy.

Especially on Linux different distributions have different rules on what instruction sets they require, so it's good practice to allow people do configure that. Unfortunately, there is no way on the command line to override a `#define`, at least none that I know of, so the only other way is patching the file at compile time, which is kinda ugly.

Right now CMake offers the user some flags like `USE_F16C`. Those originally come from Jolt, I just recently modified our CMake to make use of them as well.

The problem in current master is that I can set `USE_F16C=OFF`, but DXMath will still use F16C.

I understand that you don't want to put all defines in the VS solution, that's why I proposed adding an `ifdef` for CMake builds so it can be overridden there. This way VS doesn't get littered with compile time definitions and CMake based builds still get the flexibility they need. Since CMake automatically sets the preprocessor defines, not `#defining` stuff in header files a lot cleaner than in VS.

The only other "solution" I can think of is modifying our CMake files to tell people that we don't respect those flags, and we will always compile AVX/F16C/FMA3, and if they want to change that, they'll have to modify the sources and basically fork the code. Which doesn't feel good to me. The official builds could still define what instruction sets they use, and we can even add a warning that other settings are untested and unsupported (who knows, maybe there's some bug that only triggers when using AVX2), but outright denying it is not something I want to do.

But if you insist, I will add the aforementioned warning. I want the USE_F16C etc. to not be misleading; preferably by doing what they advertise, but if I have to disable them, that's what I will do.

There is actually a third option: CMake could in theory modify the wiMath.h file accordingly before compiling, but that's a road I really don't want to travel.

turanszkij:
(Discussed outside of gihub)

PR #1188, opened 2025-07-29
remove pragma gccs

turanszkij:


brakhane:
just to clear some things up: those pragmas work! They do exactly what they are supposed to do. There is just a bug in GCC that only triggers when it's part of a PCH that the "maybe uninitialized" warning cannot be turned off at that part. Working around that specific instance is easily done by just doing "Entity entity = 0" or any other value.

I think in general the "maybe-uninitialized" warning is useful, so I'd prefer if it stays activated. IIRC there were some subtle bugs in wiGraphicsDevice_Vulkan that I noticed because of this warning.

brakhane:
I should also add that it seems using PCH on GCC doesn't really give that much of a speed advantage; I have to run some more tests, but it could very well be that enabling them for GCC is pointless anyway. Clang does give a noticable improvement and doesn't seem to have this bug (haven't tested it yet, because currently Clang doesn't have the warning set, GCC and Clang use different warnings atm)

brakhane:
I should also add that it seems using PCH on GCC doesn't really give that much of a speed advantage; I have to run some more tests, but it could very well be that enabling them for GCC is pointless anyway. Clang does give a noticable improvement and doesn't seem to have this bug (haven't tested it yet, because currently Clang doesn't have the warning set, GCC and Clang use different warnings atm)

brakhane:
I should also add that it seems using PCH on GCC doesn't really give that much of a speed advantage; I have to run some more tests, but it could very well be that enabling them for GCC is pointless anyway. Clang does give a noticable improvement and doesn't seem to have this bug (haven't tested it yet, because currently Clang doesn't have the warning set, GCC and Clang use different warnings atm).

So IMHO these changes aren't necessary; I'm happy fighting with GCC and Clang so you can focus on working on the engine. (That doesn't mean I don't *want* you to work on stuff like this, I just don't *expect* you to)

turanszkij:
For now I change the warning levels to be more closely matching what I use in visual studio. We don't need pragma gcc any more with this in engine code.

brakhane:
That also means that some subtle bugs will not trigger a warning anymore, though. 

Is there a list of warnings you use in MSVC (maybe by default) I can use? I can then try to configure Clang similarly. (For example, right now Clang does warn about a switch statement not handling all cases, while MSVC does not)

turanszkij:
I use warning level 3. The switch statement warning is now fixed.

turanszkij:
This pull request changed into enabling more warnings with -Wall and silencing some. Many warnings are now fixed at the call location. The pragma GCC is kept in ModelImporter_OBJ.cpp and wiECS.h.

WickedEngine/wiScene.cpp
```diff
@@ -5594,7 +5594,7 @@ namespace wi::scene
 				XMVECTOR movement = XMLoadFloat3(&character.movement);
 				XMVECTOR position = XMLoadFloat3(&character.position);
 				XMVECTOR height = XMVectorSet(0, character.height, 0, 0);
-				const bool any_movement = character.movement.x != 0 || character.movement.y != 0 && character.movement.z != 0;
+				const bool any_movement = character.movement.x != 0 || character.movement.y != 0 || character.movement.z != 0;
```
brakhane, line 5597: I assume this is a bugfix, or did this get in by mistake?

CMakeLists.txt
```diff
@@ -71,25 +71,27 @@ if (WIN32)
...
+    
+    # Common compiler options and warning level for CLANG and GCC:
+    add_compile_options(
+        -Wall
+
+        -Wno-unused-variable
+        -Wno-unused-function
+        -Wno-unused-but-set-variable
+        -Wno-sign-compare
+        -Wno-switch
```
brakhane, line 83: I'd prefer having -Wswitch because it helps spotting when a case was missed. But I guess one can always manually activate it when working on unfamiliar code...

Samples/Example_ImGui_Docking/Example_ImGui_Docking.cpp
```diff
@@ -1237,8 +1237,8 @@ void Example_ImGuiRenderer::Update(float dt)
 
 		//PE: stop flickering when scrollbar goes on/off.
 		if (ImGui::GetCurrentWindow()->ScrollbarSizes.x > 0) {
-			ImGui::Text("");
-			ImGui::Text("");
+			ImGui::Text("%s", "");
```
brakhane, line 1240: personally I would just go with `-Wno-format-zero-length` as I don't see the point, but this also works.

brakhane (commented):
If the last change was indeed intended, I don't see any problems with it except personal preferences.

brakhane (dismissed):
sorry, the previous comment was supposed to be approval, not comment

WickedEngine/wiScene.cpp
turanszkij: This was fix for parenthesis warning between && and || but actually && was incorrect to have there.

CMakeLists.txt
turanszkij: I can enable that

PR #1190, opened 2025-07-30
improve Ccache hit rate

brakhane:
Sorry, this will hopefully be the last PR regarding the build system for quite some time.

I missed that Ccache requires specific settings when PCH are used, otherwise the hit rate is pretty close to zero. Those settings can be done via environment variables. For Clang, an additional setting is required, but that setting can always be set even when Ccache is not in use, like in the nightly build.

Note that the first run after this is merged will still be slow, it's only the runs after that that will be fast again.

----

Change the Ccache config so that it has a much higher hit rate during compilation, see https://ccache.dev/manual/latest.html#_precompiled_headers for details.

For Clang, always setting -fno-pch-timestamp seems to be ok, so no need to special case this depending on whether or not Ccache is used.

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -8,7 +8,8 @@ env:
   CCACHE_MAXSIZE: 500Mi
   # needed because wiRenderer uses conditional include via __hasinclude
   CCACHE_NODIRECT: 1
-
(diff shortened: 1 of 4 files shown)


turanszkij (approved):

brakhane:
will be included in #1195

PR #1194, opened 2025-08-01
use correct __has_include

brakhane:
__has_include should use the same form as the #include, otherwise search paths might differ. It probably doesn't matter in practice, but better safe than sorry.


turanszkij (approved):
Huh, I didn't know this

brakhane:
Me neither, it just came up when I researched what would need to be done to add support for it into CCache

PR #1197, opened 2025-08-04
gh actions: no LTO for PR verification

brakhane:
IPO/LTO increases link time significantly; we want PR checks to finish as quickly as possible and we don't really care about the performance of the result, so disable it.

Nightly builds will still use IPO

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -114,7 +114,7 @@ jobs:
       run: |
         mkdir build
         cd build
-        cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -114,7 +114,7 @@ jobs:
       run: |
         mkdir build
         cd build
-        cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
+        cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DWICKED_USE_IPO=NO
```
turanszkij, line 117: why also for non-pr?

brakhane: we need it for the master branch because PRs cannot share caches with each other, they can only take the cache from master

turanszkij: I usually take the latest build artifact when updating releases because that can be more up to date than the nightly. Why do prs have to share cache between each other? Within one PR would the ccache work if the not-pr still use the link time optimization?

turanszkij (approved):

PR #1198, opened 2025-08-06
gltf: fix saving for current tinygltf versions, tinygltf back to v2.9.6

brakhane:
`TINYGLTF_NO_FS` now disables *all* filesystem operations, including writing. Remove the define and move the callbacks into a separate namespace to avoid collisions.

### DIFF
--- a/Editor/ModelImporter_GLTF.cpp
+++ b/Editor/ModelImporter_GLTF.cpp
@@ -9,7 +9,6 @@
 #include <wiUnorderedSet.h>
 
 #define TINYGLTF_IMPLEMENTATION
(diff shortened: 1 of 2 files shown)


brakhane:
of course, if the NO_FS is required to get it to compile on PS5, we're out of luck and can't easily upgrade except by conditionally defining NO_FS and effectively disabling saving on PS5, which I assume is not a big deal.

turanszkij:
It will need proper testing, I will delay this after current release. We don't need gltf on ps5, and probably don't need no_fs any longer that was only for uwp.

turanszkij:
merging for now, testing will be carried out until next release sometime

PR #1200, opened 2025-08-07
Fix CCache build on windows

brakhane:
This PR fixes CCache under Windows. While CCache does speed up compilation on Windows also, it's not as big of a difference as on Linux (4-5 minutes saved vs. 10+); if you don't think the added complexity is worth it, just close this PR.

----

This reverts commit 8251c2a0259ff150bea6b93bec7dab7e2396af84 and adds the necessary fix: setting the environment variable CCACHE_NODIRECT on Windows.


turanszkij:
Ok let's try for one more time until it breaks again.

PR #1206, opened 2025-08-12
improved shadow biasing

turanszkij:


turanszkij:
Experimented again, disappointed again in some cases.

PR #1207, opened 2025-08-12
Don't use by reference capture by default

brakhane:
replace `[&]` with `[this]` or `[]` where possible to avoid people accidentally capturing something by reference that needs to be copied, like strings.

Almost all changes are either to `[this]` or `[]` (to avoid a clang warning), the only exception are in TerrainWindow, where `[&, wnd]` and `[&, i]` got replaced with `[this, wnd]` and `[this, i]`, respectively.


turanszkij:
I think `this` will be good, thanks.

PR #1208, opened 2025-08-13
Prevent accidental reference captures in Editor

ricejasonf:
First off, this deletes the copy/move constructor for `Widget` since they often capture `this` in their event handlers.

I also attempted to eliminate all reference captures and benign uses of `[&]` to prevent undefined behavior when some references point to objects on the stack or objects that live in containers where certain operations can invalidate their addresses. Please consider these changes. Thank you.

Oh, and I also added a `FIXME` note that needs special attention. I think that was the only non-benign case because the terrain component belongs to a `vector` and it isn't obvious if it would not be invalidated.

### DIFF
--- a/Editor/AnimationWindow.cpp
+++ b/Editor/AnimationWindow.cpp
@@ -43,7 +43,7 @@ void AnimationWindow::Create(EditorComponent* _editor)
 	modeComboBox.AddItem("Step", AnimationComponent::AnimationSampler::Mode::STEP);
 	modeComboBox.AddItem("Linear", AnimationComponent::AnimationSampler::Mode::LINEAR);
 	modeComboBox.AddItem("Cubic spline", AnimationComponent::AnimationSampler::Mode::CUBICSPLINE);
-	modeComboBox.OnSelect([&](wi::gui::EventArgs args) {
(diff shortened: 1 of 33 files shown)


turanszkij:
Thanks for the work but I noticed that there is an earlier PR that has lambda capture changes like this: https://github.com/turanszkij/WickedEngine/pull/1207
Since that doesn't contain the problems I mentioned I will be testing that instead.

brakhane:
#1207 and this solve different problems. A reference to this becomes invalid once the widget is copied, and that happens in vector when it resizes, or eg. when emplace_back is used.

Edit: actually I didnt see this PR also does some of the stuff my PR did, but it goes further than that.

turanszkij:
That stale reference of widget is not handled here how I would like to. And the biggest part of this was similar to what you did which could cause conflicts.

ricejasonf:
The use of `std::deque` was required for holding objects with no move constructor (ie the widgets). `std::deque` is good when you want stable iterators and references. It allocates in slabs so it is not that much worse than `vector` especially if it's just holding a few menu buttons.

I will remove the overlapping commit and push here (if that is possible) in case you change your mind, but it wasn't much to change.

PR #1209, opened 2025-08-13
Delete Widget copy/move constructors to prevent invalidating this

ricejasonf:
This deletes the copy/move constructors for `Widget`. The changes here are only to fix what this change would break. ie use of `vector` and calls to `reserve`. As was discussed, this is a breaking change that could affect downstream users of any kind of object whose class derives from `Widget`. However, it does create a safer, less error prone interface that could possibly fix difficult to reproduce crashes caused by the use of invalid references in the event handlers.

### DIFF
--- a/WickedEngine/wiGUI.h
+++ b/WickedEngine/wiGUI.h
@@ -291,6 +291,9 @@ namespace wi::gui
 		Widget();
 		virtual ~Widget() = default;
 
+    // Delete copy/move to keep internal references stable.
+    Widget(const Widget &) = delete;
+
 		const std::string& GetName() const;
 		void SetName(const std::string& value);
 		std::string GetText() const;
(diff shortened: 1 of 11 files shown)


WickedEngine/wiGUI.h
```diff
@@ -291,6 +291,10 @@ namespace wi::gui
 		Widget();
 		virtual ~Widget() = default;
 
+    // Delete copy/move to prevent invalidating `this`
+    // since it is often captured by event handlers.
```
brakhane, line 295: You could add that It's not just that, but current Wicked Widgets also have some internal implementation that doesn't support being copied; otherwise it sounds like we're patronizing the programmers

ricejasonf: I don't think it was patronizing, but I changed it as requested. Plus, I undid that newline in NewContentBrowser.

turanszkij:
Thanks, I think after discussion this is the better way of making it safer to use.

PR #1222, opened 2025-09-24
rename utility_common.cpp to utility_common.c

brakhane:
compiling it as C++ source will result in missing linker symbols when using address sanitizer.

an alternative would be to keep it a .cpp file, but add `extern "C" {` at the beginning.


brakhane:
guess MSVC is not a fan. I'll make a new PR

PR #1226, opened 2025-09-26
example_imgui_docking: rename confusingly named DIVTWOPI to DEG_PER_RAD

brakhane:


turanszkij:
I don't find any of them less confusing than the other. For displaying angles on gui I'd use the wi::math::RadiansToDegrees() instead, and the DegreesToRadians() to convert back from gui to engine.

brakhane:
I'm open to alternate names, I don't really like the current one either, but the current one is very confusing. Naming 180/π (or 360/2π) as DIVTWOPI is weird; for me it reads as 1/2π. And when it's used as "divide by DIVTWOPI" it becomes even more unclear.

turanszkij:
The wi::math::RadiansToDegrees() should be used when displaying angles on gui. The DegreesToRadians() should be used when converting gui display value before passing to system.

brakhane:
guess we'll keep the confusing naming for now

turanszkij:
I don't get it, you could have changed to the RadiansToDegrees()

brakhane:
Sorry for the passive-aggressive tone, last few days were a bit rough.

I closed this PR because it's a small refactor instead of a rename, and I don't like changing the scope of a PR. I might make a new one once I find the time to replace it with function calls and test it.

turanszkij:
I changed it to radian degree conversion

PR #1233, opened 2025-10-04
sdl: fix number keys

### LINKED ISSUE #1232: Incorrect keyboard keys for switching between editor modes (0.71.822 [Linux] [Vulkan])
Wicked Engine version 0.71.822 [Linux] [Vulkan]
Keyboard shortcuts for Movement, Rotation, and scale are mixed and moved incorrectly. For example, the move mode has the keyboard shortcut set as 1, however it's only activated when pressing 2. Rotation only works when 3 is pressed when it's set to 2, etc.

brakhane:
SDL numbers the keys as they are organized on the keyboard, so the first number key is 1, not 0.

Fixes #1232


turanszkij (approved):

turanszkij:
Thanks!

PR #1238, opened 2025-10-09
Improve the logic for auto adding the weather component to the terrain

nxrighthere:
Prevents adding a weather component to the terrain entity upon loading a scene that already contains weather.


turanszkij:
Thanks

PR #1239, opened 2025-10-10
Add editor options to disable rounded corners and gradients

nxrighthere:
This is more of an aesthetic option for square fanatics like myself. 😅


turanszkij:
Looks good now, thanks

PR #1240, opened 2025-10-12
Add parameters to show\hide mesh and internal materials

nxrighthere:
The terrain generator creates many chunks with materials that appear in the material picker window and material combo boxes, making it harder to manage the actual materials.


turanszkij:
I'm not sure about this. Maybe to have an option to filter out terrain generated materials from material picker would be better.

nxrighthere:
I added two new checkboxes to the material picker window: one to show or hide transform materials (useful for hiding materials that come from imported models), and another to hide internal materials generated by terrain chunks.

nxrighthere:
I added two new checkboxes to the material picker window: one to show or hide mesh materials (useful for hiding materials that come from imported models), and another to hide internal materials generated by terrain chunks.

turanszkij (approved):

turanszkij:
This is great, thanks.

PR #1242, opened 2025-10-16
Implement smooth light shaft fading

nxrighthere:
At the moment, depending on the camera angle, the light shaft effect can sometimes cover too much of the screen and too suddenly, which doesn’t look good visually. Instead, we smoothly fade it in and out to make it more visually appealing.

### DIFF
--- a/Editor/config.ini
+++ b/Editor/config.ini
@@ -139,4 +139,4 @@ options.width = 338.000000
 options.height = 500.000000
 components.width = 338.000000
 components.height = 500.000000
-entities.height = 299.994781
+entities.height = 300.000000


turanszkij:
Could we have just one simplified parameter for this for example "fade in smoothness"? These 3 extra sliders are hard to understand how to use them.

nxrighthere:
I unified the parameters controlling the fade speed. The fade threshold is now a fixed value, currently in a sweet spot. After experimenting with making it adaptive to the fade speed, I found that keeping it fixed produces the best results.

turanszkij:
Thanks, it's much easier to use it now.

PR #1244, opened 2025-10-16
Implement throttling for the camera preview

nxrighthere:
In heavy scenes with many entities and high graphics settings, selecting a camera in the editor can reduce performance by half, causing the editor to stutter. One solution is to throttle the preview of the selected camera. By default, the throttle is set to 10 FPS.


turanszkij:
Thanks for the work but I don't want to do this throttling. I added an update to disable rendering the preview when camera window is collapsed, I think that should be enough.

PR #1245, opened 2025-10-17
Fix edge sampling in sharpen shader

nxrighthere:
Ensure that sampling at the screen edges properly clamps to valid texture coordinates; otherwise, it produces line artifacts at the edges of the screen.


turanszkij (approved):

turanszkij:
Good catch, thanks for the fix.

PR #1247, opened 2025-10-19
Fixed Slider wiGUI.cpp

MaxLegend:
Remove rounding for step
[Fixed ](fixed: When adding a slider inside the window widget, the window intercepted the mouse capture of the slider. Now it works correctly when adding a slider to the window widget.)


turanszkij:
Can you show an example of what it fixes? (code example for example) I don't really understand. There are already many sliders on window widgets in the editor and I don't see the problem currently.

turanszkij:
Closing for inactivity

PR #1248, opened 2025-10-19
sdl: use keyboard layout for letters

### LINKED ISSUE #1246: Undo is on Ctrl+Y on QWERTZ layouts when using SDL
I live in germany, therefore I got a different keyboard layout. When I press STRG+Z the undo function does nothing. When I press STRG+Y everything works.

brakhane:
The windows version already uses the local layout, so that eg. on QWERTZ keyboard, undo is still CTRL-Z even though Z is not the key next to X.

SDL didn't do this, and forced people to press CTRL-Y instead.

This change also means that AZERTY people will have a hard time navigating now, but it's consistent with the Windows version.

Fixes #1246

### DIFF
--- a/WickedEngine/wiSDLInput.cpp
+++ b/WickedEngine/wiSDLInput.cpp
@@ -24,7 +24,7 @@ namespace wi::input::sdlinput
 
     wi::vector<SDL_Event> events;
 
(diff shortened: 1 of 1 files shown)


tccountus:
Yes, this commit fixes the issue

turanszkij (approved):

PR #1253, opened 2025-10-23
CMake overhaul

CitroenGames:
the cmake structure from before was just a PITA it wasnt really easily scalable and was hard to use


turanszkij:
This is a very bad explanation of what your problem is and what you are trying to do.

PR #1254, opened 2025-10-24
Fix the chromatic aberration sampling

nxrighthere:
Ensure that all per-channel samples remain within bounds, eliminating stretched edge artifacts, especially at higher strengths.


turanszkij:
A good improvement, thanks.

PR #1256, opened 2025-10-27
Supply a manifest file to the editor

nxrighthere:
Specify OS compatibility, DPI awareness, modern common controls (for dialog windows, etc.), and the requested execution level.

### DIFF
--- a/Editor/Editor.cpp
+++ b/Editor/Editor.cpp
@@ -362,6 +362,21 @@ void Editor::Initialize()
 		}
 	}
 }
+
+void Editor::Exit()
+{
+	// Check all scenes for unsaved changes
+	for (int i = 0; i < (int)renderComponent.scenes.size(); ++i)
+	{
+		if (!renderComponent.CheckUnsavedChanges(i))
+		{
+			return;
+		}
+	}
+
+	Application::Exit();
+}
+
 void Editor::HotReload()
 {
 	if (IsScriptReplacement())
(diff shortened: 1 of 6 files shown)


turanszkij:
What exactly is this and can we do it in code instead?

nxrighthere:
To my knowledge, supplying the file is the only proper way to enable support for modern Windows OS features and prevent it from going through compatibility shims.

For instance, the editor is currently using the legacy controls:
CaptureXX

With the manifest file, we tell the OS to use modern controls:
CaptureX

The editor also properly supports high-DPI displays, so we don't want the OS to use bitmap scaling.

The security level is standard, as with most applications. We run the editor with the same privileges as the user, without a UAC elevation prompt.

turanszkij:
Is this gonna be built into the exe somehow or does it need to be released as a separate file? if it needs to be a separate file I don't want to include it in the release.

nxrighthere:
Yes, in this case, it will be embedded into the exe.

turanszkij:
All right, looks good to me then, thanks.

brakhane:
Well, the CMake stuff is still missing, but I can fix that in the future, shouldn't be that hard

nxrighthere:
@brakhane You think `Editor.manifest` won’t be picked up by CMake automatically? We should test that, I think it should work as is.

brakhane:
> You think Editor.manifest won’t be picked up by CMake automatically? 

Haven't tested it, but I don't think it will, as source files in Editor's makefiles are configured to be GLOB `*.cpp`

PR #1258, opened 2025-10-27
Add question to SDL to mirror windows behaviour

tccountus:
When the user closes the window, he should be asked on linux too.

### DIFF
--- a/Editor/Editor.cpp
+++ b/Editor/Editor.cpp
@@ -5258,10 +5258,10 @@ void EditorComponent::Open(std::string filename)
 }
 void EditorComponent::Save(const std::string& filename)
 {
-	struct ClearFlagOnExit {
+	struct SetAndClearFlagOnExit {
 		bool& flag;
-		ClearFlagOnExit(bool& flag) : flag(flag) {}
-		~ClearFlagOnExit() {flag = false;}
+		SetAndClearFlagOnExit(bool& flag) : flag(flag) { flag = true; }
+		~SetAndClearFlagOnExit() { flag = false; }
 	} scoped(save_in_progress);
 
 	std::string extension = wi::helper::toUpper(wi::helper::GetExtensionFromFileName(filename));
--- a/Editor/main_Windows.cpp
+++ b/Editor/main_Windows.cpp
@@ -8,6 +8,8 @@
 
 Editor editor;
 
+bool quit = false;
+
 int APIENTRY wWinMain(_In_ HINSTANCE hInstance,
                      _In_opt_ HINSTANCE hPrevInstance,
                      _In_ LPWSTR    lpCmdLine,
@@ -86,6 +88,7 @@ int APIENTRY wWinMain(_In_ HINSTANCE hInstance,
 		break;
 		case WM_CLOSE:
 			editor.Exit();
+			quit = true;
 			break;
 		case WM_DESTROY:
 			PostQuitMessage(0);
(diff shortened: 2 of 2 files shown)


brakhane:
Not sure if the Windows version has the same problem, but it only seems to work correctly if the scene has already been saved before. If I start from an empty scene, add something and quit, it will ask me if I want to save the untitled scene. Clicking Yes will allow me to input the filename, but the scene is not saved, the program just exists after that

nxrighthere:
@brakhane  On Windows, the untitled scene saves just fine for me, both the first save and subsequent ones.

The dialog invokes `EditorComponent::SaveAs()` and then checks for `editorscene.has_unsaved_changes` [here](https://github.com/turanszkij/WickedEngine/blob/02e27e3644480ee5a851ccfc96b68149eb54fc58/Editor/Editor.cpp#L6156).

tccountus:
I just mirrored the windows commit to the SDL version. The dialog doesn't save for me too.

tccountus:
It was PR #1255

nxrighthere:
Could you set some breakpoints [here](https://github.com/turanszkij/WickedEngine/blob/02e27e3644480ee5a851ccfc96b68149eb54fc58/Editor/Editor.cpp#L6156) and check what’s going on?

tccountus:
The breakpoints gets triggered when I did the following:
Add an object, Close the window, Click yes

nxrighthere:
Try to step into `EditorComponent::SaveAs()` to see what’s happening during scene saving.

turanszkij:
I noticed that SDL_ShowMessageBox return value was checked incorrectly in wiHelper.cpp, that's now fixed, though I don't think that would cause this problem described here.

brakhane:
> I noticed that SDL_ShowMessageBox return value was checked incorrectly in wiHelper.cpp, that's now fixed, though I don't think that would cause this problem described here.

@turanszkij That fix is actually wrong. in SDL3 `SDL_ShowMessageBox` indeed returns a bool, but in SDL2 a negative value is an error and zero is success.

I'll add a revert to it in this PR once I find the reason why saving unchanged doesn't work.

tccountus:
I think the problem lies inside the FileDialog method in wiHelper.cpp; I added an debug output to the SaveAs method inside the callback, but it doesn't seem to show anything

brakhane:
To avoid duplicate work: I'm currently working on it. The underlying problem is that Save() actually finishes saving before returning, but SaveAs will not. SaveAs will call Save in a callback, but not actually wait for that callback to finish. ~~What makes it a bit tricky is that the callback can not finish executing without the SDL main loop running.~~

brakhane:
OK, I think I solved it. Save and SaveAs now set a flag that gets reset once Save completes and the main loops have been changed to only quit if both the quit flag has been set and the editor says it can close (eg. save_in_progress is false). Haven't tested it on windows yet, though.

brakhane:
OK, after some fixes works on Windows as well. @turanszkij Please have a look. Obviously the changes in main_Windows are not strictly necessary since it seemed to work before (my guess is Windows somehow processes the callbacks differently so it works out there), but I changed it nonetheless for consistency with the SDL version.

turanszkij:
Thanks @brakhane

nxrighthere:
image

This breaks the exit button on Windows.

brakhane:
Hmm, I'm pretty sure I tested that and it worked on my machine. I'll check again later

PR #1268, opened 2025-10-31
fix alignment of colliders_{c,g}pu

### LINKED ISSUE #1265: [CRASH] Commit 932c90b breaks the engine
Hey, the last commit breaks the engine on Linux (Kubuntu 24.04)

Stack:
Thread 31 "wi::job_18" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fff8f7ee6c0 (LWP 22355)]
0x00005555561f76bd in wi::scene::Scene::RunProceduralAnimationUpdateSystem(wi::jobsystem::context&)::{lambda(wi::jobsystem::JobArgs)#5}::operator()(wi::jobsystem::JobArgs) const [clone .isra.0] ()
(gdb) bt
#0  0x00005555561f76bd in wi::scene::Scene::RunProceduralAnimationUpdateSystem(wi::jobsystem::context&)::{lambda(wi::jobsystem::JobArgs)#5}::operator()(wi::jobsystem::JobArgs) const [clone .isra.0] ()
#1  0x0000555555c17d24 in wi::jobsystem::Job::execute() ()
#2  0x0000555555c17f0c in std::thread::_State_impl<std::thread::_Invoker<std::tuple<wi::jobsystem::Initialize(unsigned int)::{lambda()#1}> > >::_M_run() ()
#3  0x00007ffff7aecdb4 in std::execute_native_thread_routine (__p=0x55555bec53d0) at ../../../../../src/libstdc++-v3/src/c++11/thread.cc:104
#4  0x00007ffff769caa4 in start_thread (arg=<optimized out>) at ./nptl/pthread_create.c:447
#5  0x00007ffff7729c6c in clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:78

I get a SEGFAULT

To reproduce: Use latest commit (after 932c90b), build with cmake and gcc (debug build) (it worked before the commit), run the Editor and add Terrain -> CRASH

brakhane:
They are now 32 byte aligned, but since we store them manually in big contiguous memory region, we have to take care of the alignment ourselves.

Fixes #1265

### DIFF
--- a/WickedEngine/wiScene.cpp
+++ b/WickedEngine/wiScene.cpp
@@ -1028,12 +1028,27 @@ namespace wi::scene
 			sizeof(ColliderComponent) * collider_count_cpu +
 			sizeof(ColliderComponent) * collider_count_gpu
 		;
-		collider_deinterleaved_data.reserve(size);
-		ASAN_UNPOISON_MEMORY_REGION(collider_deinterleaved_data.data(), size);
-		aabb_colliders_cpu = (wi::primitive::AABB*)collider_deinterleaved_data.data();
-		aabb_colliders_gpu = aabb_colliders_cpu + collider_count_cpu;
-		colliders_cpu = (ColliderComponent*)(aabb_colliders_gpu + collider_count_gpu);
+		// we're going to store AABB and colliders in one big array, for
+		// this to work with their alignment, ColliderComponents needs
+		// to be allocated first at a 32-byte aligned address, then
+		// AABB, as the latter have a size multiple of 16, so if
+		// collider_count_cpu and collider_count_gpu are not both odd or even,
+		// we would end up at a multiple of 16.
+
+		// First, we need to make sure our current assumptions are correct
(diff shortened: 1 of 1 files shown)


turanszkij:
Is the problem here the different alignemnt of AABB and collider? What if AABB is also 32 byte aligned?

brakhane:
alignment of AABB doesn't matter per se, the size of the struct does, it needs to be a multiple of 32. But I've worked around that by just putting the colliders first, then sizeof(AABB) doesn't matter. That still leaves the problem that `collider_deinterleaved_data.data()` needs to be 32-byte aligned. On Linux it's currently 16-byte aligned, but probably mostly by chance anyway, so it worked before the changes to the alignment only by chance. I don't see a way around manually aligning the start of the array, but also I don't see a problem with it, except that it wastes up to 31 bytes.

turanszkij (approved):

PR #1277, opened 2025-11-02
archive: don't use undefined behaviour

brakhane:
unaligned stores are technically undefinied behaviour, even though they are allowed on x86. But replacing those with memcpy should result in the same code on x86 anyway, as it's optimized away by the compiler, and will not cause issues on other architectures that don't support it.

It also allows us to use `-fsanitize=undefined` to find other, potentially more problematic, issues without getting distracted by errors messages cause by this.


turanszkij:
Thanks!

PR #1282, opened 2025-11-04
Implement wireframe overlay

nxrighthere:
In complex scenes, it's hard to see mesh topology using wireframe alone. Implement a wireframe overlay option that draws the wireframe on top of the geometry.

image


turanszkij:
Very nice, thanks.

PR #1291, opened 2025-11-07
Implement rigidbody physics shape cache

nxrighthere:
Creating physics shapes at large scale for triangle meshes and convex hulls is expensive; cache them instead of recreating for the same meshes:

- Implemented `PhysicsShapeCache`, stored in `PhysicsScene`
- Added `ScaledShapeSettings` to handle scale variations across meshes
- Massively improves scene loading and runtime performance (e.g., many terrain props with rigidbodies)
- Caching is disabled for the terrain

### DIFF
--- a/Editor/Editor.cpp
+++ b/Editor/Editor.cpp
@@ -669,8 +669,7 @@ void EditorComponent::Load()
 		Scene& scene = GetCurrentScene();
 		PickResult pick;
 
-		XMFLOAT3 in_front_of_camera;
-		XMStoreFloat3(&in_front_of_camera, XMVectorAdd(camera.GetEye(), camera.GetAt() * 4));
+		XMFLOAT3 in_front_of_camera = GetPositionInFrontOfCamera();
 
 		switch (args.userdata)
 		{
@@ -882,6 +881,27 @@ void EditorComponent::Load()
 		}
 		if (pick.entity != INVALID_ENTITY)
 		{
+			// Place the entity in front of camera, if enabled
+			if (generalWnd.placeInFrontOfCameraCheckBox.GetCheck())
+			{
+				TransformComponent* transform = scene.transforms.GetComponent(pick.entity);
+				if (transform != nullptr)
+				{
+					// Check if this entity type should spawn in front of camera
+					bool should_reposition = true;
+					if (args.userdata == NEW_DIRECTIONALLIGHT || args.userdata == NEW_CAMERA)
+					{
+						should_reposition = false;
+					}
+
+					if (should_reposition)
+					{
+						transform->translation_local = in_front_of_camera;
+						transform->SetDirty();
+					}
+				}
+			}
+
 			wi::Archive& archive = AdvanceHistory();
 			archive << EditorComponent::HISTORYOP_ADD;
 			RecordSelection(archive);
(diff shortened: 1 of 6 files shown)


turanszkij:
There is already a way to cache rigid body shape for a mesh, you can use the MeshComponent::precomputed_rigidbody_physics_shape for that. If this is created, it will be reused when creating rigidbodies from this mesh. Can you not make use of that instead?

You can check for example usage in wiTerrain.cpp where the shape caching is used on the chunk generation background thread at this call:
wi::physics::CreateRigidBodyShape(newrigidbody, transform.scale_local, &mesh);

nxrighthere:
I could potentially use `MeshComponent::precomputed_rigidbody_physics_shape` in a similar way since `physicsobject` in `RigidBodyPhysicsComponent` is a shared pointer, but that would make it dependent on the terrain generation logic specifically handling terrain props. The code in this PR makes rigidbody shape caching scene-wide, regardless of its source, for example, an entity spawned dynamically from user code.

I feel like I want to refactor `MeshComponent::precomputed_rigidbody_physics_shape` to make it more lightweight instead of using the entire `RigidBodyPhysicsComponent`, decoupling it from `MeshComponent` and have it specifically handle `RigidBodyPhysicsComponent::HEIGHTFIELD` for terrain, since that’s the only place it’s used.

nxrighthere:
I can add an optional checkbox in this PR to enable rigidbody shape caching per component, if you’re concerned.

turanszkij:
What do you mean it would be dependent on terrain generation logic? That's just one example where it is used, it could be used probably in other situations as well.

But the bigger issue is that this breaks physics, there are so many configuration of mesh scaling and such that it needs to be handles extremely carefully. One example where it breaks is the FPS sample island level:

Correct (before):
sc_09-11-2025 08-22-56

Incorrect (after):
sc_09-11-2025 08-20-49

From the first look, only the rocks are broken (blown up to huge) but there might be others, this is just the first thing I saw. Unfortunately subtle changes logic at setup and how physics shape is synced to TransformComponent could easily break existing assets.

If you want to test it, this level is in Wicked Shooter paid version here: https://turanszkij.itch.io/wicked-shooter
Located in assets/levels/island/island.wiscene
Because you are already a supporter on Patreon, you have access to this, let me know if you can't access it.

nxrighthere:
> What do you mean it would be dependent on terrain generation logic? That's just one example where it is used, it could be used probably in other situations as well.

Of course, but that’s exactly what this PR already does. It covers all cases where rigidbodies are used with complex shapes. I must have misunderstood you earlier.

I’ll look into what’s going on in the shooter.

nxrighthere:
It should be fixed now. I forgot to apply scaling to physics shapes that are not in the cache.

I haven’t noticed this in our game because we don’t use source objects in the scene, only their instances.

nxrighthere:
Note that the yellow physics shapes in the visualizer represent scaled shapes, unlike the red ones.

nxrighthere:
I have refactored `MeshComponent::precomputed_rigidbody_physics_shape` internally, as planned per https://github.com/turanszkij/WickedEngine/pull/1291#issuecomment-3506545568, to specifically serve terrain heightfields and decoupled it from `MeshComponent`, since it’s merely a vessel for the physics shape of the terrain.

Let me know how you’d like to proceed with this. I’m still open to wrapping the code in this PR under an optional checkbox on the rigidbody component if you prefer, or we can close this PR and move on. It’s not a problem for me, it just makes it a bit harder to resolve conflicts with the changes upstream.

turanszkij:
Hi, I have refactored this branch to not change physics interface, basically moved the whole caching logic into the AddRigidBody. This works together with the terrain chunk precomputation, no changes required for that. The CreateShape function is now without changes, but in case of caching it will simply receive a unit scaling and scaling will be applied at physice instance creation. Please test that it still works for your case.

nxrighthere:
This is a much cleaner way! I tested it internally, and everything seems to work correctly. Thanks. 👍🏻

PR #1293, opened 2025-11-09
Implement adaptive granularity-based scrolling

nxrighthere:
The current scrolling implementation doesn’t respect granularity, causing the mouse wheel to scroll through the content in large steps.


turanszkij (approved):
Yes, very good I wanted to have this for a while. Thanks.

PR #1295, opened 2025-11-10
Add an option to bypass the system key 

nxrighthere:
On Windows, pressing the Alt key activates the menu bar, which we don’t use. Consume `WM_SYSKEYDOWN` and `WM_SYSKEYUP` to prevent the editor from freezing caused by activated menu bar.

### DIFF
--- a/WickedEngine/wiPhysics_Jolt.cpp
+++ b/WickedEngine/wiPhysics_Jolt.cpp
@@ -6,6 +6,7 @@
 #include "wiJobSystem.h"
 #include "wiRenderer.h"
 #include "wiTimer.h"
+#include "wiSpinLock.h"
 
 #include <Jolt/Jolt.h>
 #include <Jolt/RegisterTypes.h>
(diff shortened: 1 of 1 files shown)


turanszkij:
Actually I use this behaviour, good to pause application.

zpl-zak:
Annoyingly, this still emits a beep sound when running windowed and you combine Alt with other keys, like WASD.

Adding an extra event to block fixes that: 
```c
		case WM_SYSCHAR:
			return 0;
```

EDIT: SYSKEYDOWN and SYSKEYUP can break alt+f4 shortcut, etc. Here's a complete version that preserves alt+f4, suppresses beep and menu bar activation:
```c
		case WM_SYSCHAR:
		case WM_SYSKEYDOWN:
		case WM_SYSKEYUP:
			if (wParam == VK_F4) {
				return DefWindowProc(hWnd, message, wParam, lParam);
			}
			// Handle system key messages (like Alt key) to prevent menu bar activation freeze
			// Return 0 to indicate we processed the message and prevent default behavior
			return 0;
```

To be fair this should probably be locked behind a compile-time switch as this behavior is quite subjective.

nxrighthere:
Let’s make this optional as a config variable?

nxrighthere:
It can now be configured using the `bypass_system_key` parameter.

turanszkij (changes requested):

turanszkij (approved):

PR #1299, opened 2025-11-14
get rid of some undefined behaviour

brakhane:
### overloaded function call in `wiGraphicsDevice_*.cpp`

A call of the form `to_internal((const Bar*)foo)` is undefined behaviour if `foo` is not an instance of a subclass of Bar; it also took me a while to understand that the code is using it to call the correct overloaded function. I've rewritten those calls using templates to achieve the same thing while not relying on undef behavior. I also find it more readable, but I'm obviously biased ;)

### adding offset to a nullptr in `wiScene.cpp`

I don't think this is would ever be a de facto problem on any C++ compiler, but technically the compiler is allowed to do anything, including completely omitting the code; moving the calculation inside an `if` should result in the same code due to optimization (haven't verified it though), so I just changed it to be on the safe side.

### casting NaN to int in `wiScene.cpp`  and `wiScene_Components.h`

There are probably more instances, I've just changed those that UBSAN noticed when running an old version of the FPS demo and character_controller. There are probably more instances and I feel like the functions should never be called with NaN, but to keep it consistent, I changed it so that instead adding an assert (and therefore crashing), it will explicitly set the values to zero, which is what clang seems to do.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -105,10 +105,15 @@ endif()
 if (WIN32)
     set(PLATFORM "Windows")
     add_compile_definitions(WIN32=1)
+    add_compile_definitions(_HAS_EXCEPTIONS=0)
+
     if (MSVC)
         add_compile_options(
-            /W3
-            /MP
+            /W3 # warning level 3
+            /MP # multi-processor compilation
+            /EHsc- # exceptions disabled
+            /GR- # runtime type information disabled
+            $<$<CONFIG:Release>:/GS-> # security check disabled in Release
         )
     endif()
 
@@ -116,15 +121,27 @@ if (WIN32)
 elseif(UNIX)
     set(PLATFORM "SDL2")
     add_compile_definitions(SDL2=1)
+    add_compile_definitions(_GLIBCXX_USE_CXX11_ABI=1)
 
     # Common compiler options and warning level for CLANG and GCC:
     add_compile_options(
-        -Wall
+        -Wall # warning level: all
 
+        # some warnings are disabled to better match MSVC warning level 3:
         -Wno-unused-variable
         -Wno-unused-function
         -Wno-unused-but-set-variable
         -Wno-sign-compare
+
+        $<$<COMPILE_LANGUAGE:CXX>:-fno-exceptions> # exceptions disabled
+        $<$<COMPILE_LANGUAGE:CXX>:-fno-rtti> # runtime type information disabled
+
+        # security checks disabled in Release:
+        $<$<CONFIG:Release>:-fno-stack-protector>
+        $<$<CONFIG:Release>:-fcf-protection=none>
+        $<$<CONFIG:Release>:-fno-stack-clash-protection>
+        $<$<CONFIG:Release>:-fno-stack-check>
+        $<$<CONFIG:Release>:-fno-asynchronous-unwind-tables>
     )
 endif()
 
@@ -155,7 +172,6 @@ elseif (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
         -Wno-strict-aliasing
     )
 endif()
-
 
 add_subdirectory(WickedEngine)
 add_custom_target(Content
--- a/Editor/ContentBrowserWindow.cpp
+++ b/Editor/ContentBrowserWindow.cpp
@@ -404,6 +404,13 @@ void ContentBrowserWindow::AddItem(const std::string& filename, const std::strin
 			});
 		this->SetVisible(false);
 		});
+	button.OnRightClick([this, filename](wi::gui::EventArgs args) {
+		wi::eventhandler::Subscribe_Once(wi::eventhandler::EVENT_THREAD_SAFE_POINT, [=](uint64_t userdata) {
+			editor->NewScene();
+			editor->Open(filename);
+			});
+		this->SetVisible(false);
+		});
 	button.font_description.params.h_align = wi::font::WIFALIGN_CENTER;
 	button.font_description.params.v_align = wi::font::WIFALIGN_TOP;
 	button.font.params.size = 42;
@@ -438,4 +445,5 @@ void ContentBrowserWindow::AddItem(const std::string& filename, const std::strin
 			}
 		}
 	}
+	button.SetTooltip(button.GetTooltip() + "\n\nLeft-click: Merge into current scene\nRight-click: Open in new scene tab");
 }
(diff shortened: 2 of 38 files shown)


turanszkij:
I don't want to merge these changes sorry.

1.) I think I am using this kind of casting correctly, I never had issue with it. I would rather not change things in device that is absolutely necessary.
2.) I don't understand the problem
3.) Adds runtime checks to common path which will be used in mesh uploading. If there is incorrect data, that should be fixed in the model file.

brakhane:
1. Would you reconsider if I verify that the generated assembly in both MSVC and GCC is (basically) the same? I'm also open to getting rid of the undefined behaviour in other ways. First, apparently dereferencing after casting  `A*` to `B*` is still undefined even when `A` and `B` share a parent and the only thing accessed via the pointer is an element of the parent. Casting to `B*` is allowed, dereferencing it isn't. It works on the current compilers, but strictly speaking is not required to stay that way (and maybe on ARM it will not work, haven't tested it). Secondly, casting that pointer solely to call a different function is really confusing, it took me a minute to understand what was going on. Others might have the same problem, so I feel if we can avoid undefined behaviour while *also* making the intention more clear, we should.

2. It's admittably a silly case of undef behaviour. Adding integers to pointers is only allowed when the pointer points to an object. The nullptr does not do so, so it's undefined. Since it's easy to avoid this UBSAN error by slightly rewriting the code, I thought it a good idea to do so; optimization should ensure that it's the same at runtime. I could live without this change though.

3. If the function is not supposed to be called with NaN, I will undo the change. Would it be ok to replace it with asserts? Because apparently when that happens something went wrong. It would mean that currently some example models are wrong and need to be fixed to work in debug mode, though.

turanszkij:
1. I see this as a kind of refactor, but I still don't understand how this fixes something. This is still casting, and that should be valid, but accessing a wrong cast will be invalid. Refactor is usually fine for cpp files but there are several people now with modified graphics device so it's good to avoid bigger changes here unless there is a big benefit.
2. I don't think there is an issue here, just offsetting a pointer. Access is still guarded by the if. Your changes make it compute the pointer offset for every bone instead of once for whole skeleton. But the pointer will not change while iterating the bones. If this would be a problem then the whole engine wouldn't work, we use pointer offsets a lot.
3. There shouldn't be validation for it inside the function. But there can be a validator helper function as part of the scene if you want that can be called by user. There is an example Scene::FixupNans().

brakhane:
1. The difference is that the original code casts eg. a GPUResource into a Texture; while they do share a GraphicsDeviceChild as a parent, a GPUResource is not a Texture, and since GraphicsDeviceChild has a virtual destructor and hence a vtab, that's undefined behaviour. My change doesn't cast a GPUResource into a texture, it casts the `internal_state`. The current implementation probably will keep working on windows and linux ABI (but I haven't checked), but AIUI on other platforms, if GPUResource* p is 0x1234, then (const Texture*)p is allowed to be say 0x1250 to adjust for vtab stuff. It *should* probably only matter if there are pointers to member functions, though. So in practise this UB will probably not cause problems, but if there's an easy fix, why not.

2. I'm aware I've moved it into the loop, but since the members accessed are not volatile, the compiler should optimize it and only assign it once at the start of the loop. I haven't checked that tbf, but can do so.

3. What if the validation is an assert though? Those will only be in debug mode. If the function is not supposed to be called with NaNs, an assert doesn't hurt to drive that point home and it helps to spot bugs. If for some reason MSVC changes how a nan gets cast into an integer, it could suddenly change how some models look or behave and it will be time consuming to figure out the reason is that the original model caused some NaN somewhere.

brakhane:
Also regarding 1 I should also mention that removing the copy pasted `to_internal` is not necessary, I could leave that part as is and just add a `template<typename T> T* to_internal_as(const GraphicsDeviceChild*)` or something and then replace the later `to_internal` calls that cast the parameter into `to_internal_as`, if you'd prefer that.

turanszkij:
For (1) I was not aware that the virtual pointer can be put to other offsets. Is there any compiler that does it? I guess we can put in those device changes then.

For (2) I think this is not needed, I can't imagine this is a problem, it's adding numbers.

For (3) I don't want assert in the vertextype conversion structure. We assume that model is provided correctly.

brakhane:
> For (1) I was not aware that the virtual pointer can be put to other offsets. Is there any compiler that does it? I guess we can put in those device changes then.

Great. I'll rework this PR including it.

> For (2) I think this is not needed, I can't imagine this is a problem, it's adding numbers.

Technically the compiler is free to format the hard drive in that case, but I agree it won't be a problem in practice. We'll have to add a sanitizer override if we keep it though, otherwise it will potentially hide other issues since people get used to the error messages when they activate UBSAN. I'll figure something out.

> For (3) I don't want assert in the vertextype conversion structure. We assume that model is provided correctly.

I don't see the problem with it as it's removed in release mode, but ok, I'll try to add exceptions for these occurrences as well. Should I try to find out where the NaNs are coming from, or do you already know what the problem in the character controller and FPS demos are?

turanszkij:
This looks good to me now. 👍

As for the vertex assert, I don't want this to assert really, if the normals are wrong in the model it will be very annoying for development, but I don't really care usually, I don't even make the models.

turanszkij:
By the way, what if you do this, does it still complain for you?
```
ShaderTransform* gpu_dst = skinningDataMapped == nullptr ? nullptr : (ShaderTransform*)((uint8_t*)skinningDataMapped + armature.gpuBoneOffset);
```

brakhane:
I can check later, but from what I can see that would fix the first "adding something to nullptr", but still leaves the second in `uint8_t* gpu_bone_dst = (uint8_t*)(gpu_dst + boneIndex);` since now gpu_dst is nullptr. Of course, we could add another check but I think at this point it will be less performant than just moving the calculation of gpu_dst into the loop (which I still think should be optimized to be outside of the loop)

brakhane:
Also since clang and gcc allow overriding specific errors on lambda scope I'm fine with the current solution. Of course it might hide other "overflowing pointer arithmetic" stuff in the same lambda in the future, but how likely is that? I also agree with you that it's not a real problem. Sure, a compiler would technically be correct if it just threw a runtime exception or generate completely broken code, but why would it? What I think is more likely is that something like `(uint32_t*)nullptr + 1` would result in 1 for some compilers and 4 on others; but we don't use the value anyway.

brakhane:
(no changes from previous version, just rebased)

PR #1301, opened 2025-11-15
compile optimizations

turanszkij:
- [x] remove exceptions
- [x] remove runtime security checks from release builds
- [x] disable runtime type information

These features will not be used, also better matching default console project settings


CMakeLists.txt
```diff
@@ -156,6 +162,7 @@ elseif (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
     )
 endif()
 
+add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-fno-exceptions>)
```
brakhane: this would also add the flag for MSVC

PR #1304, opened 2025-11-16
Fixing DPI Scaling on KDE Wayland Linux when display scale isn't 100%

Kemny:
Hi, this is a small fix for DPI scaling on linux.

I've downloaded the engine today, and quickly needed to track down the DPI settings, because the text was unreadably small on my monitor. I did my best to follow conding standards, but if there is something wrong, I'll do my best to fix it. 

One thing that comes to mind is whether the DPI should be set to the default 96 on SDL error, or whether there should be an exception thrown in that case, I opted for just defaulting to the previous value for the sake of not throwing random exceptions around if things break, since that seems to be the standard.

**Environment**
OS: Linux Fedora 43
DE: KDE Wayland (I'm assuming the same issue will be present on other desktop environments)
Compiler: Clang 21.1.5
Monitors: Kamvas 16 gen 3 (150% scale) & Framework13 2880x1920 screen (200% scale)

**How to repro the issue**
1. Increase the scaling of a monitor in KDE settings (I'm assuming the exact same repro applies for GNOME)
2. Start the editor
3. Notice the editor UI doesn't respect the DPI scale

**Tested edge cases**
Changing the scale and confirming that the DPI was getting updated accordingly.
Changing the scale while the editor was running.
Dragging the application window between 2 monitors with different DPI.

### DIFF
--- a/WickedEngine/wiPlatform.h
+++ b/WickedEngine/wiPlatform.h
@@ -88,7 +88,14 @@ namespace wi::platform
 		int window_width, window_height;
 		SDL_GetWindowSize(window, &window_width, &window_height);
 		SDL_Vulkan_GetDrawableSize(window, &dest->width, &dest->height);
-		dest->dpi = ((float)dest->width / (float)window_width) * 96.f;
+
+		float dpi;
+		int window_index = SDL_GetWindowDisplayIndex(window);
+		if (SDL_GetDisplayDPI(window_index, &dpi, nullptr, nullptr) != 0)
+		{
+			dpi = 96.f;
+		}
+		dest->dpi = ((float)dest->width / (float)window_width) * dpi;
 #endif // PLATFORM_LINUX
 	}
 }


brakhane:
I noticed the SDL docs have this

> WARNING: This reports the DPI that the hardware reports, and it is not always reliable! It is almost always better to use [SDL_GetWindowSize](https://wiki.libsdl.org/SDL2/SDL_GetWindowSize)() to find the window size, which might be in logical points instead of pixels, and then [SDL_GL_GetDrawableSize](https://wiki.libsdl.org/SDL2/SDL_GL_GetDrawableSize)(), [SDL_Vulkan_GetDrawableSize](https://wiki.libsdl.org/SDL2/SDL_Vulkan_GetDrawableSize)(), [SDL_Metal_GetDrawableSize](https://wiki.libsdl.org/SDL2/SDL_Metal_GetDrawableSize)(), or [SDL_GetRendererOutputSize](https://wiki.libsdl.org/SDL2/SDL_GetRendererOutputSize)(), and compare the two values to get an actual scaling value between the two. We will be rethinking how high-dpi details should be managed in SDL3 to make things more consistent, reliable, and clear.

But I think it's better than nothing.

brakhane:
Hmm, something seems wrong. If I set my monitor scaling to 100% (I usually have it at 200%), it's way too big:

grafik

Kemny:
Uh oh. I'm not sure what could be the issue in that case. I'm guessing that the SDL docs were right about this being unrelibable?

The GetDisplayDPI returns 96 when scale is set to 100, and 192 when set to 200, etc. I assumed that would be universal, but can you please check if you have other values? Or maybe a different UI scale with the bottom left scaler?

For reference, this is what it looks like for me with the fix:
image

And this is what it looks like without it:
image

brakhane:
With 100% scaling I get back a dpi of 407.053, with 200% (my default) it's 101.763. I have a 2256x1504 laptop display

PAMinerva:
@Kemny Are you using sdl2-compat?
Because I have the same problem and, as discussed on discord, it seems that this issue only affects users that have sdl2-compat installed.

Kemny:
@brakhane That is definitely not what I expected to hear when it comes to numbers. 
But if what @PAMinerva is saying is true, then I suppose that would explain it. I've never even heard about this lib before, it looks like it's something that fedora just does all by itself. I'm not sure if there's a way to switch it out for regular sdl2 without breaking 10 more things elsewhere in the desktop.

Sadly, my knowledge of SDL ended the moment we started talking about compatibility layers. I'll close this PR, and just keep the change locally on my computer, since I can't test this more without doing ungodly things to my computer, or installing another distro.

Sorry for the confusion, and thanks for the help.

brakhane:
@Kemny sdl2-compat is the sdl2 compatiblity layer for sdl3. We'll probably update the code to use SDL3 in the future, but that's only reasonable if most people use up to date distros that come with SDL3.

PR #1317, opened 2025-11-21
Expose the character collision tolerance option

nxrighthere:
It’s important to have this option adjusted from the application logic to handle cases where the character might fall through large triangles or complex geometry.


zpl-zak:
It also makes sense to expose this to Lua bindings. What do you think?

nxrighthere:
Sure, that makes sense. I don’t use Lua for game logic, so please check it.

zpl-zak:
Yeah looks good, thanks for adding it!

PR #1328, opened 2025-11-24
Bugfix: C++ to Lua Type conversion fails in some cases for the RenderPaths

### LINKED ISSUE #1327: Bug: C++ to Lua Type conversion fails in some cases for the RenderPaths
[Error] [Lua Error] Line 4: GetActivePath() Warning: type of active component not registered!
SuballocateGPUBuffer created buffer block with size: 256.0 MB, with page size: 64.0 KB, page count: 4096
Scene::Serialize took 0.11 seconds
[Error] [Lua Error] [string "..."]:150: [Lua Error] [string "--[[/home/martin/Downloads/wicked-shooter-wind..."]:52: attempt to index a nil value (local 'path')

This is because the engine compares the pointers. This happens when built with gcc

tccountus:
Resolves #1327

### DIFF
--- a/WickedEngine/wiRenderPath3D_PathTracing.h
+++ b/WickedEngine/wiRenderPath3D_PathTracing.h
@@ -47,7 +47,7 @@ namespace wi
 
 		// This is an identifier of RenderPath subtype that is used for lua binding.
 		static constexpr std::string_view script_check_identifier = relative_path(__FILE__);
-		virtual std::string_view GetScriptBindingID() const { return script_check_identifier; }
+		virtual std::string_view GetScriptBindingID() const override { return script_check_identifier; }
 	};
 
 }


tccountus:
This may not be the „best“ solution, but it fixes the issue by using string views

brakhane:
I can't see how this would solve it except by accident. The relative_path should already be a constant pointer throughout the runtime of the program

tccountus:
On gcc it's somehow not. You can try it out: build it with gcc on ubuntu 24.04 LTS and try to run the demo script at https://turanszkij.itch.io/wicked-shooter . It somehow fails because the components are nil (in the lua). With the commit it works.

Test it if you don't trust me

tccountus:
(At least it fixes the bug for gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0)

brakhane:
I wasn't saying you're making stuff up. What gcc version does ubuntu 24.04 use? IIRC the official binaries are also built using ubuntu 24.04, so it's weird it doesn't work for you.

But apparently constexpr does not guarantee the result will always be the same pointer given the same argument, but it's still weird, since `__FILE__` gets replaced by the preprocessor and should be treated like any other fixed string.

I'll try to get ubuntu running in a VM later.

tccountus:
thanks for checking, building with clang seems to resolve the issue without this commit

turanszkij:
Can you try this branch or build? https://github.com/turanszkij/WickedEngine/pull/1329

tccountus:
Your commit fixes it somehow

turanszkij:
Thank you for testing, I will go with my approach, I want to avoid string comparison for these objects.

PR #1339, opened 2025-11-29
Fix duplicated texture paths during scene save/load

nxrighthere:
Texture paths are getting duplicated during scene save/load cycles. For example, a path like `C:/project/textures/diffuse.png` would become `C:/project/C:/project/textures/diffuse.png`  after saving and loading, eventually growing into extremely long corrupted paths with many duplications.

When loading scenes, texture paths are unconditionally prepended with the source directory `path = dir + path`, without checking if the path was already absolute or already contained the directory prefix.

In release builds, this fix will **not** automatically correct corrupted file paths in existing scenes, but in debug builds the logic will automatically resolve them.

I have tested this change internally and with the Wicked Shooter scenes by repeatedly saving and loading scenes and adding new textures, so it should be stable in its current state.

Please make backups of your scenes before testing this change.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -140,9 +140,9 @@ else()
         # security checks disabled in Release:
         $<$<CONFIG:Release>:-fno-stack-protector>
         $<$<CONFIG:Release>:-fcf-protection=none>
-        $<$<AND:$<CONFIG:Release>,$<NOT:$<PLATFORM_ID:Windows>>>:-fno-stack-clash-protection>
+        $<$<AND:$<CONFIG:Release>,$<NOT:$<PLATFORM_ID:Windows>>>:-fno-stack-clash-protection> # not supported on Windows
         $<$<CONFIG:Release>:-fno-stack-check>
-        $<$<CONFIG:Release>:-fno-asynchronous-unwind-tables>
+        $<$<AND:$<CONFIG:Release>,$<NOT:$<PLATFORM_ID:Windows>>>:-fno-asynchronous-unwind-tables>  # seems to crash LUA evaluation on Windows
     )
 endif()
 
--- a/Editor/MaterialWindow.cpp
+++ b/Editor/MaterialWindow.cpp
@@ -1254,6 +1254,7 @@ void MaterialWindow::ResizeLayout()
 		{
 			windowTitle += " (Entity: " + name->name + ")";
 		}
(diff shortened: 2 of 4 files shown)


turanszkij:
How exactly to repro the broken behaviour? Because in my short tests I couldn't (opening and saving the sample Sponza.wiscene multiple times)

My intention with paths in the scene was like this:
- material.textureName and such always contain absolute path normally, matching the entry in wiResourceManager
- when saving, the path is made relative to the serialization directory
- when loading, path coming from archive is assumed to be relative to serialization directory, so seri path is added to make it absolute

nxrighthere:
Thank you for clarifying. This fix aligns with your original intent. The only important difference is that we now first check whether the path is already absolute.

Steps to reproduce:
1. Create a new scene and create a cube entity
2. Load a texture from a file for the cube's material base color map texture slot [C:/textures/diffuse.jpg]
3. Save the scene to a file on another disk [D:/project/scene.wiscene]
4. Close the scene and open it again:
```
[DEBUG] Texture names in loaded scene:
  Material[cube] slot[0]: D:/project/C:/textures/diffuse.jpg
```
5. Save the scene, close it, and open it again:
```
[DEBUG] Texture names in loaded scene:
  Material[cube] slot[0]: D:/project/D:/project/C:/textures/diffuse.jpg
```

turanszkij:
You're right, damn this is a pretty serious issue. But this attempt looks too complicated when I look what it is doing. I am seeing that this part in MakePathRelative is not entered when saving to a different drive letter meaning the relative path resolution failed:

```cpp
if (!relative.empty())
{
	StringConvert(relative.generic_wstring(), path);
}
```

Do you think it could be fixed here instead by removing the directories instead from the filename and keeping it only a filename in  the inout path string? My suggestion:

```cpp
if (!relative.empty())
{
	StringConvert(relative.generic_wstring(), path);
}
else
{
	path = GetFileNameFromPath(path); // force it to be in relative format, if this happened then it could mean path is absolute but relative couldn't be resolved (drive letter change)
}
```

The problem with you solution is that it still keeps old absolute path in the filename which will be saved into asset (and it will reveal asset the developer's computer's file structure)

nxrighthere:
I agree, I hadn’t considered the privacy side of things. Let me check out your option.

nxrighthere:
Let's use your solution for this. Feel free to push it. 👍🏻

PR #1344, opened 2025-12-03
Implement the force unlit option

nxrighthere:
Useful for calibrating color maps and for color-matching albedo textures in the scene by disabling lighting and shadows.


turanszkij:
It would have been nicer to use the existing unlit shader, but this will do as temporary solution.

PR #1346, opened 2025-12-04
Mac OS support

turanszkij:


turanszkij:
Issue tracking will be continued here: https://github.com/turanszkij/WickedEngine/issues/1479

PR #1348, opened 2025-12-06
vulkan: increase timeout to 3s

brakhane:
2s timeout is a bit low on my igpu, 3s seems to work better.

Also, use chrono literals to make it a bit more readable.


turanszkij:
In what way does it work better? Also I think including a header to make one line more readable is not good.

brakhane:
With 2s during initialization and loading the first scene I almost always get an error logged about a timeout in graphics and compute queue, even though it later works normally. With 3s the error does not appear.

Personally, I think readability is worth adding another header, but that's personal preference. I've removed the header and replaced it with digit separators for a bit more readability in the new version.

PR #1350, opened 2025-12-08
Implement a thumbnail renderer for prefab scenes

nxrighthere:
For scenes that contain only a single root entity and effectively serve as prefab scenes, use a dedicated renderer to generate thumbnails for the content browser:
- Consistent, clear dimensions and angles of meshes
- No post-processing effects from the editor’s render path
- Independent from viewport dimensions
- Complex scene thumbnails are rendered as before, without any changes

thumbnails


nxrighthere:
This PR also adds a utility function that we use extensively in our project. It iterates over each child entity and applies a lambda function, with optional early returns.

turanszkij (approved):

PR #1356, opened 2025-12-10
cmake: add option to compile and embed shaders during the build process

brakhane:
This adds an option `WICKED_EMBED_SHADERS` (off by default). When on, `offlineshadercompiler` will be compiled first and run every time a `*.hlsl?` file changes, the resulting `wiShaderDump.h` will be put in the build directory, and then wiRenderer will be compiled again with the build directory in the include path; this will cause the `__hasinclude` to trigger embedded shaders mode.

A nice side effect is that the hacky solution to handle the `__hasinclude` is not needed anymore, all dependencies are now handled correctly.

`offlineshadercompiler` was also extended with a `quiet` option to suppress messages during compilation.

The build scripts were also adjusted to use it where applicable.


turanszkij:
Looks like it compiled in the pull request but not on the master branch, could you please check this?

PR #1362, opened 2025-12-12
Improve camera speed mouse scroll

nxrighthere:
Improve the camera scroll speed by using finer 0.1 increments below the 1.0 threshold, making navigation through very small spaces easier. Alice in Wonderland would appreciate the change!


turanszkij (approved):

turanszkij:
Thanks, we will find out how deep the rabbit hole goes...

PR #1368, opened 2025-12-13
gh actions: force lld linker on clang, re-enable IPO on nightly/master

brakhane:
for some reason, github actions seem to try to link clang code with the GNU linker instead of LLD, which doesn't work when LTO/IPO is enabled. Explicitly specifying LLD on Clang builds seems to solve this.

Hence, set the linker explicitly and re-enable IPO as it produces slightly better code.

IPO stays disabled for PR because it increases compilation time and we want fast compilation feedback for those

### DIFF
--- a/.github/workflows/build-nightly.yml
+++ b/.github/workflows/build-nightly.yml
@@ -64,7 +64,7 @@ jobs:
 
     - name: Compile with embedded shaders
       run: |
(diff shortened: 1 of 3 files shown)


brakhane:
Closing this because compiling with LTO would make the cache useless for PR builds.

PR #1375, opened 2025-12-19
Implement spline fill and recenter options

nxrighthere:
Implement spline fill and recenter options:
- Centroid-based triangle fan generation
- Supports subdivision
- Automatic UV and normals calculation
- Works only in combination with the looped spline option

image


turanszkij:
Wow this is cool, thanks for the work!
image

turanszkij (approved):

PR #1492, opened 2026-01-13
components: reduce code duplication when setting flags

brakhane:
Just some small refactoring to reduce repetition and to make it easier to see when setters have side effects besides changing the given flag.

I'm not sure if SetFlag should be in a separate namespace, I just put it there because generally component flags should be changed via the component's setter and I consider SetFlag to be an implementation detail, not a public API.

### DIFF
--- a/WickedEngine/CommonInclude.h
+++ b/WickedEngine/CommonInclude.h
@@ -384,7 +384,18 @@ constexpr bool has_flag(E lhs, E rhs)
 {
 	return (lhs & rhs) == rhs;
 }
-
+template<typename T, typename U>
+constexpr void set_flag(T& flags, const U flag, const bool set)
+{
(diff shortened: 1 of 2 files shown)


turanszkij (changes requested):
If we do this I'd prefer to add a more generic function to CommonInclude.h instead where there is already flag manipulation helper `has_flag`. And not use the `this` class, but instead give it the destination variable directly, this way it could be used anywhere. For example:
```
template<typename T>
constexpr void set_flag(T& dest, T mask, bool enabled)
```

brakhane:
sounds good. I've made those changes, have a look

turanszkij (approved):

turanszkij:
Thx, I think this way is better.

PR #1493, opened 2026-01-13
ecs: add asserts to catch lookups using invalid value

brakhane:
Got bitten by this by not checking the return value "because it surely exists" and then got a segmentation fault.


brakhane:
Hmm, good point. Those assert is a bit more obscure when they happen, but it's probably not worth it adding these asserts.

Issue #1456, opened 2026-01-01
[feature] Use local space


Anthony-Gaudino:
An option to switch to use local space would be nice when transforming rotated objects.

Godot example:

https://github.com/user-attachments/assets/6d6893d6-fb17-46d6-8e3f-4fd9e14c5497

PR #1494, opened 2026-01-16
Implement local space mode for the translator

nxrighthere:
Toggle between global and local space via hotkey <kbd>4</kbd> or the button beneath the _Scale_. This implementation follows standard UI/UX patterns from other popular 3D packages.

localspace

Implements #1456


turanszkij:
This is great, a minor thing is that the ui doesn't look good for it in default settings:
image

I'll fix this

turanszkij:
The local mode scaling also doesn't seem to work correctly, however normal scaling also has issues when rotation is applied. I think this is just the problem with the usage of decompose in my implementation and not specifically your local mode implementation.

nxrighthere:
Yes, there is an issue with the original decomposition logic when using non-uniform scaling with rotation. I’m not sure what the optimal solution is, but for me it’s not a critical problem.

Thanks for fixing the rounded corners, I always forget to test the default theme.

PR #1496, opened 2026-01-16
Improve the depth indicator for the hierarchy component

nxrighthere:
I realized that my original approach for depth indication might get out of hand in complex hierarchies like character rigs. So, instead of displaying `>>>` infinitely, limit it to three levels and use numbers thereafter e.g., `> 4 >` for better readability.

### DIFF
--- a/Editor/HierarchyWindow.cpp
+++ b/Editor/HierarchyWindow.cpp
@@ -120,7 +120,7 @@ void HierarchyWindow::SetEntity(Entity entity)
 			std::string depth_indicator;
 			if (depth > 3)
 			{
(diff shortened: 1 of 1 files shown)


brakhane:
I would suggest formatting it as `>%2i`, because otherwise the levels look like

    > first
    >> second
    >>> third
    >4 fourth

which looks odd. `> 4` would look nicer IMO. Or even use 3i, so that `>   4` is actually 4 long

nxrighthere:
I was playing around with it and I decided to stick to the `> 4 >` format.

turanszkij (approved):

PR #1498, opened 2026-01-16
Change bone mapping order in humanoid rig creation

nxrighthere:
I was lurking around that code and randomly noticed that the mapping don’t match the pattern used for the other bones. I haven’t tested this, it’s just an assumption.


turanszkij:
It just matches the order of HumanoidBone enum values, I don't think there is a problem. The order could be changed here because it's a map but I don't think it would improve anything.

PR #1507, opened 2026-01-19
allocate application on heap instead of stack and delete it

brakhane:
This prevents problems with Vulkan debug layers that don't like Vulkan calls to happen during C++ app shutdown.

Not pretty, but does the job


brakhane:
Windows is not tested yet, just a quick proof of concept; since I have a feeling you don't want to merge this anyway ;)

turanszkij:
I don't want to merge these, I recommend to add application.graphicsdevice.reset() on your end when dealing with this issue of debug device.

PR #1530, opened 2026-01-31
Editor: Linux file dialogs: remember last-used dir

metaleap:
Thanks for giving consideration to this minor Editor-on-Linux proposal:

- Behavior stays like it was previously for users who never change dirs in Open/Save dialogs.
- But for those who previously had to keep selecting other folders than `$PWD` every single time in an open/save-file dialog, now the last-used directory (in this session; ie. no persistence) is being defaulted to instead.
- Open and Save dialog each keep their own last-used dir path. If in a user's workflow both stay the same anyway, then no problem. But some workflows keep importing assets from one dir and saving `.wiscene`s to another, and for those, that separation can be beneficial.

### DIFF
--- a/WickedEngine/wiHelper.cpp
+++ b/WickedEngine/wiHelper.cpp
@@ -1545,11 +1545,15 @@ namespace wi::helper
 				{
 					options = options | pfd::opt::multiselect;
 				}
-				std::vector<std::string> selection = pfd::open_file("Open file", "", extensions, options).result();
+				static std::string recent_dir_path = "";
+				std::vector<std::string> selection = pfd::open_file("Open file", recent_dir_path, extensions, options).result();
 				if (!selection.empty())
 				{
 					for (auto& x : selection)
 					{
+#ifdef PLATFORM_LINUX
+						recent_dir_path = GetDirectoryFromPath(x);
+#endif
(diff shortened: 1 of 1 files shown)


turanszkij:
Can you try this commit instead: https://github.com/turanszkij/WickedEngine/commit/223e982a085e4cdaabe04aeb4c2417f8554c444b
This uses system-default path to match windows behaviour. This retains last directory across sessions. At least on Mac OS that works now.

metaleap:
Just tried it, doesn't work like that on Linux (ie. Zenity) unfortunately.

Looking at my PR here, how about this: I could initialize the 2 `recent_dir_path` vars to "" and then would `#ifdef PLATFORM_LINUX` the 2 `recent_dir_path` variable-setting lines: this way, for Mac users it's like in your commit, and yet Linux users would get a functioning recent-dir UX too.

What you think?

turanszkij:
You can make ifdef for linux then, but leave it with blank default dir for mac os.

metaleap:
> You can make ifdef for linux then, but leave it with blank default dir for mac os.

~~Will do so, dunno why GH closed this PR but will reopen it shortly with that change.~~ 
**Done:**

PR #1549, opened 2026-02-11
Implement moon in terrain/weather

Anthony-Gaudino:
Adds a moon when a terrain is added.

The moon is a second `Directional light`, so it has it's own light and casts shadows.
Users can load a texture for the moon.
Moon properties are accessible in the Weather component.

Moon, sun and earth lighting interactions simulations are implemented, for example, the earth shadow is cast on the moon.

Tested on Linux with Vulkan.

-----

Current issues:

* The moon glow is always a circle, which is not accurate, the glow should come from lighted areas in the moon.
* Stars are visible trough the moon
* Solar eclipse is not accurate, the moon is not visible in front of the sun and the eclipse happens too abruptly.
* Lunar eclipse needs work too
* Default settings, like moon size are not accurate, the moon should have the same size as the Sun in the sky, also automatic eclipses should be the default.
* [Moon color](https://www.littlepassports.com/blog/space/moon-colors/) should change depending on some factors
* Code cleanup and improvement

-----

Help would be appreciated.


WickedEngine/wiScene.cpp
```diff
@@ -31,6 +33,111 @@ namespace wi::scene
 {
 	static constexpr uint32_t small_subtask_groupsize = 256u;
 
+	static float ComputeMoonPhaseVisibility(const XMFLOAT3& sun_dir, const XMFLOAT3& moon_dir)
+	{
+		XMVECTOR sun = XMLoadFloat3(&sun_dir);
+		XMVECTOR moon = XMLoadFloat3(&moon_dir);
+		const float sun_len_sq = XMVectorGetX(XMVector3LengthSq(sun));
+		const float moon_len_sq = XMVectorGetX(XMVector3LengthSq(moon));
+		if (!std::isfinite(sun_len_sq) || !std::isfinite(moon_len_sq) || sun_len_sq < 1e-6f || moon_len_sq < 1e-6f)
```
turanszkij: We should avoid all these std::isfinite calls, instead make it to never be infinite (check that source vector is not zero if you must). Then we also don't need to include cmath.

turanszkij (commented):
Haven't checked everything, just a quick try and some observations.
- When I place a terrain, now there are two directional lights, sun and moon. Moon cannot be rotated like the sun
- Moon can be rotated only inside weather settings, but I think it should be working like the sun. It should be active only if someone puts a directional light, and for example tags it as moon
- It looks like moon is lit on the other side of where it should be relative to the sun-moon directions
- I think the moon halo effect is not necessary, since it is not the same shape as the moon it looks out of place. Maybe we can just rely on the bloom effect picking up a bright moon, or it could be also handled by the lensflare system
- Moon should be optional, just like the sun. Especially moon shadows, we don't want to render them usually, only when there is no sun. Moon parameters on a directional light could work better, because that can be then placed like the sun.

Good work so far!

socialtwisty:
I love that you are adding the moon!
Is there any way to make it more visible? its very faint (texture or no texture), and I don't know how to control the phase:

image

Anthony-Gaudino:
Hi @socialtwisty

Answering your questions:

You can make the moon more visible by changing it's color:
Screenshot From 2026-02-22 13-38-03
If you change it to white then it will be very visible.

------------------

The moon phase is calculated automatically based on the sun position. If you rotate the sun you can have different moon phases. Currently it's not accurate.

socialtwisty:
I found a problem where playing with the rain amount can cause a crash

Anthony-Gaudino:
> I found a problem where playing with the rain amount can cause a crash

@socialtwisty could you please try to replicate the crash in the main branch to check if this is crash is related to this branch or not.

Anthony-Gaudino:
@turanszkij I believe I have addressed the issues you pointed. Also other bugs were solved, more features like lunar eclipse were implemented and I cleaned and refactored the code.

Many moon parameters aren't exposed in the UI, if you would like some to be exposed, please let me know. I decided to not expose them for now to avoid cluttering the UI and avoid having a PR that is too large.

Also, I added a Sun size multiplier just like the moon has, this way users can control size of both. The default multiplier approximately matches the previous sun size.

PR #1552, opened 2026-02-12
Fix pinning of editor component windows

nxrighthere:
Fix bug where collapsing `HumanoidWindow`, `ArmatureWindow`, and `AnimationWindow` on child entities without the component would release the component window making it impossible to use.

### DIFF
--- a/Editor/AnimationWindow.cpp
+++ b/Editor/AnimationWindow.cpp
@@ -1071,7 +1071,7 @@ void AnimationWindow::SetEntity(Entity entity)
 		wi::scene::Scene& scene = editor->GetCurrentScene();
 
 		AnimationComponent* animation = scene.animations.GetComponent(entity);
-		if (animation != nullptr || IsCollapsed())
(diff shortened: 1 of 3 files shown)


turanszkij (approved):

nxrighthere:
@turanszkij I realized that with this change, those windows now remain pinned to entities that are not part of the hierarchy with the component. For example, if I select an entity with a humanoid component window and then select a directional light entity, the humanoid window remains pinned.

Should we replace
```cpp
if (animation != nullptr || (IsCollapsed() && entity == INVALID_ENTITY))
```
with
```cpp
if (animation != nullptr || !scene.Entity_IsDescendant(entity, this->entity))
```
instead?

Or perhaps you have a better idea?

turanszkij:
Ok, I changed the logic to this:
```
if (humanoid != nullptr || (IsCollapsed() && !scene.Entity_IsDescendant(entity, this->entity)))
```
- remains active always while opened
- remains active when collapsed but selecting a child

Let me know if this works for you better.

PR #1560, opened 2026-02-19
Improve the transform tool

### LINKED ISSUE #1454: Rotation gizmo takes too much screen space
I think that the rotation gizmo takes too much screen space and covers the element being rotated. This happens because it uses flat shapes.

Image

Another issue with using flat shapes is that on a certain angle you can't manipulate the rotation gizmo of a given axis. For example, here the red gizmo is too thin:

Image

I think that the gizmo should be thick lines like the other two.

I checked how Godot does it, and it does use thick lines for it's rotation Gizmo:

Image

Also it's gizmos are smaller.

nxrighthere:
Improve the appearance of the transform tool, expose and organize its parameters in the general options for tweaking. I used Blender as a reference when refining the transform tool’s appearance.

Closes #1454

params
rotation_tool

### DIFF
--- a/WickedEngine/offlineshadercompiler.cpp
+++ b/WickedEngine/offlineshadercompiler.cpp
@@ -224,8 +224,9 @@ wi::vector<ShaderEntry> shaders = {
 	{"lightmap_expandCS", wi::graphics::ShaderStage::CS },
 
 
-	{"emittedparticlePS_soft", wi::graphics::ShaderStage::PS },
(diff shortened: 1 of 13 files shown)


turanszkij:
Thanks, I want the translator to remain the same as before by default, but the options to modify are nice to have.

nxrighthere:
I’ve restored its default visual look.

turanszkij:
Thanks!

PR #1563, opened 2026-02-21
Fix building with PIC

tccountus:
This fixes building with PIC Enabled


brakhane:
Haven't checked, but it looks ok. Though IIRC there were other issues that currently prevent compiling Wicked as a dynamic lib.

tccountus:
For me it currently works, i'm currently experimenting with java bindings, these two patches make it work

brakhane (approved):

PR #1564, opened 2026-02-21
Add toggle for RTTI

tccountus:
When using the engine in combination with the JVM (Java virtual machine), you need RTTI. This patch adds an flag to enable the rtti


brakhane:
LGTM

brakhane (approved):

PR #1566, opened 2026-02-23
Stable Diffusion AI image generation

turanszkij:
Generate images in editor, apply them to anything.


megumumpkin:
Turanski, I think it would be better to have online asset integration like with sites like polyhaven or ambientcg than this.
I'm currently working on this.

turanszkij:
They are not mutually exclusive, although after more testing, I'm not satisfied with this and I'm going to experiment with it separately from the engine.

PR #1567, opened 2026-02-26
Implement optimization of the physics broadphase for terrain props

nxrighthere:
Jolt’s broadphase uses a fixed-size node pool for physics bodies. In scenarios such as terrain generation, where many props are likely to have collision shapes, we should periodically compact the pool to free stale nodes. This allows new bodies to be allocated and prevents the engine from forcefully terminating when the pool is exhausted.

### DIFF
--- a/WickedEngine/wiRenderPath3D.cpp
+++ b/WickedEngine/wiRenderPath3D.cpp
@@ -1274,6 +1274,7 @@ namespace wi
 		{
 			cmd_shadowmap = device->BeginCommandList();
 			cmd = cmd_shadowmap;
+			device->WaitCommandList(cmd, cmd_prepareframe_async); // shadow map waits for UpdateRenderDataAsync (particle-shadowmap interaction)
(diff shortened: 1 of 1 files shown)


WickedEngine/wiPhysics_BindLua.cpp
```diff
@@ -513,6 +514,23 @@ namespace wi::lua
...
+			scene::Scene_BindLua* scene = Luna<scene::Scene_BindLua>::lightcheck(L, 1);
+			if (scene == nullptr)
+			{
+				wi::lua::SError(L, "OptimizeBroadPhase(Scene scene) first argument is not a Scene!");
+				return 0;
+			}
+			wi::physics::OptimizeBroadPhase(*scene->scene);
+		}
+		else
+			wi::lua::SError(L, "OptimizeBroadPhase(Scene scene, [opt bool force]) not enough arguments!");
```
nxrighthere: @turanszkij I forgot to remove `[opt bool force]` from the comment here. 😅

PR #1576, opened 2026-03-05
update dxc to 2026.02 (linux and win only)

brakhane:
@turanszkij As mentioned in Discord, this version seems to create invalid SPIR-V for some shaders.
Also, since I don't have a Mac, the dylib is missing, but homebrew might have the current version. Alternatively, the github repo should compile on MacOS


JonahGoldsmith:
All Mac Developers automatically get Metal Shader Converter through XCode most likely, which includes a dxc dylib and executable. on Mac would it have to be gated by Wicked's versioning or by Metal Shader Converters?

brakhane:
I think we should use the same DXC version for everything. DXC changes details between versions and it sometimes has bugs (like eg. this version creates incorrect SPIR-V code). Ignoring the different APIs for now, it would make debugging between platforms more difficult since you have to take the different behaviour of the DXC versions into account.

PR #1595, opened 2026-04-02
Metal: Implement Draw*IndirectCount emulation via CPU based loop

JonahGoldsmith:
Implements DrawInstancedIndirectCount and DrawIndexedInstancedIndirectCount for Metal by emulating ExecuteIndirectCount behavior.
Adds a helper compute sanitize pass that copies indirect records to scratch, clamps active count, and zeroes instanceCount for inactive records.
Adds minimal render-pass snapshot/split/resume plumbing required to run sanitize compute between render draws without mutating source argument buffers.

Rationale:

Metal render encoders do not provide a native indirect-count draw equivalent to DX12/Vulkan ExecuteIndirectCount semantics.
This patch restores cross-backend behavior by emulating count-based multi-draw safely and deterministically.

Known Limitations:
Emulation iterates max_count records on CPU-side command submission; inactive records are made no-op by zeroing instanceCount.
Requires an active render-pass snapshot (captured in RenderPassBegin) to split for compute sanitize and resume correctly.

(Sanitization version causes weird rendering artifacts and flickers, switched to cpu loop below )

### DIFF
--- a/WickedEngine/wiGraphicsDevice_Metal.cpp
+++ b/WickedEngine/wiGraphicsDevice_Metal.cpp
@@ -9,9 +9,53 @@
 #include "wiBacklog.h"
 
 #include <Metal/MTL4AccelerationStructure.hpp>
+#include <cstddef>
 
 namespace wi::graphics
 {
+
+namespace
+{
+	static constexpr uint32_t INDIRECT_COUNT_HELPER_THREADGROUP_SIZE = 64;
+	static const char* INDIRECT_COUNT_HELPER_MSL = R"(
+		#include <metal_stdlib>
+		using namespace metal;
+
+		struct WIIndirectCountHelperParams
+		{
+			uint stride;
+			uint argument_size;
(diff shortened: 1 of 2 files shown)


JonahGoldsmith:
I have changed the implementation to use a CPU loop on Metal. If we were to make it use a full on Indirect Command Buffer, it would require changing a lot of the Metal internals to add specific scenario tracking when someone fills an Indirect Buffer from the GPU and not the CPU. But when filled from the CPU this implementation works correctly

turanszkij:
It kind of defeats the purpose if we have to set the count from cpu, at that point it's the same as regular drawindirect. I think it would be better to have metal-specific implementation in render code (for specific effect) to try how it works first. But we don't even use drawindirect count right now anywhere to test it.

The other thing you mentioned that paused the encoder and switched to an other one in the middle is also not too good, an other reason it would be needed to have some kind of metal specific pass before the render encoder is started.

PR #1596, opened 2026-04-05
Make cmake build work on MacOS

SiebrenW:
So I tried building the MacOS version of the engine, since I just got a MacBook, but I did notice that XCode is a massive POS and I couldn't even figure out what setting to change to make it work. (because its error messages are 20 clicks away and cryptic as they can be)

After messing quite a bit with it I gave up and I spent significantly less time to fix/hack it in CMake, which is my preference anyway (which is a low bar).

Since it's not production ready or anything I figured it may be helpful to have it. I still notice some wonkiness, but at least it builds.

Apple clang was crashing when building, so I switched to llvm clang. I also suppressed/solved some warnings.

Cheers.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,16 +27,25 @@ option(USE_LIBCXX "Link WickedEngine to llvm libc++ library - only available wit
 option(WICKED_EDITOR "Build WickedEngine editor" ON)
 option(WICKED_TESTS "Build WickedEngine tests" ON)
 option(WICKED_IMGUI_EXAMPLE "Build WickedEngine imgui example" ON)
-option(WICKED_ENABLE_IPO "Enable IPO/LTO in non-debug builds" NO)
-option(WICKED_EMBED_SHADERS "Embed shaders into the library" NO)
+option(WICKED_ENABLE_IPO "Enable IPO/LTO in non-debug builds" OFF)
+option(WICKED_EMBED_SHADERS "Embed shaders into the library" OFF)
 option(WICKED_ENABLE_RTTI "Enable RTTI" NO)
-if(UNIX)
+if (APPLE)
+    set(WICKED_TESTS OFF) # no main for macos
+    set(WICKED_IMGUI_EXAMPLE OFF) # no main for macos
+    set(WICKED_ENABLE_IPO OFF)
+endif()
+
+if(UNIX AND NOT APPLE)
     option(WICKED_ENABLE_ASAN "Enable AddressSanitizer in debug builds" OFF)
     option(WICKED_ENABLE_UBSAN "Enable UndefinedBehaviourSanitizer in debug builds" OFF)
     option(WICKED_ENABLE_TSAN "Enable ThreadSanitizer in debug builds" OFF)
     option(WICKED_ENABLE_SAN_ALWAYS "Enable the selected sanitizers in all builds, not just debug" OFF)
 endif()
-if(UNIX)
+
+if(APPLE)
+    option(WICKED_MACOS_TEMPLATE "Build WickedEngine MacOS template" ON)
+elseif(UNIX)
     option(WICKED_LINUX_TEMPLATE "Build WickedEngine Linux template" ON)
 elseif(WIN32)
     option(WICKED_WINDOWS_TEMPLATE "Build WickedEngine Windows template" ON)
@@ -113,6 +122,15 @@ if (WIN32)
     set(PLATFORM "Windows")
     add_compile_definitions(WIN32=1)
     add_compile_definitions(_HAS_EXCEPTIONS=0)
+elseif (APPLE)
+    # set(CMAKE_C_COMPILER /opt/homebrew/opt/llvm/bin/clang)
+    # set(CMAKE_CXX_COMPILER /opt/homebrew/opt/llvm/bin/clang++)
+    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -fuse-ld=lld")
+    set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -fuse-ld=lld")
+    set(CMAKE_MODULE_LINKER_FLAGS "${CMAKE_MODULE_LINKER_FLAGS} -fuse-ld=lld")
+    set(CMAKE_LINKER ld64.lld)
+	set(PLATFORM "MacOS")
+    add_compile_definitions(MACOS=1)
 elseif (UNIX)
     set(PLATFORM "SDL2")
     add_compile_definitions(SDL2=1)
@@ -147,7 +165,7 @@ else()
         # security checks disabled in Release:
         $<$<CONFIG:Release>:-fno-stack-protector>
         $<$<CONFIG:Release>:-fcf-protection=none>
-        $<$<AND:$<CONFIG:Release>,$<NOT:$<PLATFORM_ID:Windows>>>:-fno-stack-clash-protection> # not supported on Windows
+        $<$<AND:$<CONFIG:Release>,$<NOT:$<OR:$<PLATFORM_ID:Windows>,$<PLATFORM_ID:Darwin>>>>:-fno-stack-clash-protection> # not supported on Windows or macOS
         $<$<CONFIG:Release>:-fno-stack-check>
         $<$<AND:$<CONFIG:Release>,$<NOT:$<PLATFORM_ID:Windows>>>:-fno-asynchronous-unwind-tables>  # seems to crash LUA evaluation on Windows
     )
@@ -204,6 +222,10 @@ endif()
 
 if (WICKED_LINUX_TEMPLATE)
     add_subdirectory(Samples/Template_Linux)
+endif()
+
+if (WICKED_MACOS_TEMPLATE)
+    add_subdirectory(Samples/Template_MacOS)
 endif()
 
 if (WICKED_WINDOWS_TEMPLATE)
--- a/Editor/CMakeLists.txt
+++ b/Editor/CMakeLists.txt
@@ -6,8 +6,14 @@ string(REGEX REPLACE "([.+?*])" "\\\\\\1" SDIR "${CMAKE_CURRENT_SOURCE_DIR}")
 file(GLOB SOURCE_FILES CONFIGURE_DEPENDS *.cpp)
 list(FILTER SOURCE_FILES EXCLUDE REGEX ${SDIR}/main_.*)
 list(FILTER SOURCE_FILES EXCLUDE REGEX ${SDIR}/stdafx.*)
-list(APPEND SOURCE_FILES main_${PLATFORM}.cpp)
 
+if (APPLE)
+    list(APPEND SOURCE_FILES main_MacOS.mm)
+    # The project PCH doesn't work with objective C, so disable it for the mm files.
+    set_source_files_properties(main_MacOS.mm PROPERTIES SKIP_PRECOMPILE_HEADERS ON)
+else()
+    list(APPEND SOURCE_FILES main_${PLATFORM}.cpp)
+endif()
 
 if (WIN32)
     list (APPEND SOURCE_FILES
@@ -30,6 +36,22 @@ if (WIN32)
 
     set_property(TARGET Editor PROPERTY VS_DEBUGGER_WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}")
     set(LIB_DXCOMPILER "dxcompiler.dll")
+elseif (APPLE)
+    add_executable(Editor ${SOURCE_FILES})
+
+    find_library(APPKIT_FRAMEWORK AppKit REQUIRED)
+
+    target_link_libraries(Editor PUBLIC
+        WickedEngine
+        ${APPKIT_FRAMEWORK}
+    )
+    set(LIB_DXCOMPILER "libdxcompiler.dylib")
+
+    # needed for proper names in crash stacktrace
+    set_target_properties(Editor
+        PROPERTIES
+        ENABLE_EXPORTS ON
+    )
 else ()
     add_executable(Editor ${SOURCE_FILES})
 
--- a/WickedEngine/CMakeLists.txt
+++ b/WickedEngine/CMakeLists.txt
@@ -6,21 +6,36 @@ if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.31)
 endif()
 
 set(MIN_OpenImageDenoise_VERSION "2.0")
+option(CPU_IS_X86_ARCH "Whether the CPU architecture is (a subset of) x86, implying potential instructions are supported" OFF)
 
 include(CMakePackageConfigHelpers)
 include(GNUInstallDirs)
 set(INSTALL_LOCAL_CONFIGDIR "${CMAKE_BINARY_DIR}/cmake")
 set(INSTALL_CONFIGDIR "${CMAKE_INSTALL_LIBDIR}/cmake/WickedEngine")
 
+if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
+    if ("${CMAKE_VS_PLATFORM_NAME}" STREQUAL "x86" OR "${CMAKE_VS_PLATFORM_NAME}" STREQUAL "x64")
+        set(CPU_IS_X86_ARCH ON)
+    else()
+        set(CPU_IS_X86_ARCH OFF)
+    endif()
+elseif ("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "x86_64" OR "${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "AMD64" OR "${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "amd64" OR "${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "x86" OR "${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "i386")
+    set(CPU_IS_X86_ARCH ON)
+else()
+    set(CPU_IS_X86_ARCH OFF)
+endif()
+
 # Use same options as Jolt
 # Select X86 processor features to use (if everything is off it will be SSE2 compatible)
-option(USE_SSE4_1 "Enable SSE4.1" ON)
-option(USE_SSE4_2 "Enable SSE4.2" ON)
-option(USE_AVX "Enable AVX" ON)
-option(USE_AVX2 "Enable AVX2" OFF)
-option(USE_AVX512 "Enable AVX512" OFF)
-option(USE_LZCNT "Enable LZCNT" ON)
-option(USE_TZCNT "Enable TZCNT" ON)
+if (CPU_IS_X86_ARCH)
+    option(USE_SSE4_1 "Enable SSE4.1" ON)
+    option(USE_SSE4_2 "Enable SSE4.2" ON)
+    option(USE_AVX "Enable AVX" ON)
+    option(USE_AVX2 "Enable AVX2" OFF)
+    option(USE_AVX512 "Enable AVX512" OFF)
+    option(USE_LZCNT "Enable LZCNT" ON)
+    option(USE_TZCNT "Enable TZCNT" ON)
+endif()
 option(USE_F16C "Enable F16C" ON)
 option(USE_FMADD "Enable FMADD" ON)
 
@@ -37,6 +52,15 @@ if (WIN32)
         add_compile_options(
             /bigobj
         )
+    endif()
+elseif (APPLE)
+	# find_package(Metal )
+    find_package(OpenImageDenoise "${MIN_OpenImageDenoise_VERSION}" QUIET)
+    find_package(Threads REQUIRED)
+    if(NOT ${OpenImageDenoise_FOUND})
+        message("OpenImageDenoise>=${MIN_OpenImageDenoise_VERSION} not found, it will be disabled.")
+    else()
+        message("OpenImageDenoise ${OpenImageDenoise_VERSION} Found.")
     endif()
 else ()
     find_package(SDL2 REQUIRED)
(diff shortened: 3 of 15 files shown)


Editor/CMakeLists.txt
```diff
@@ -4,10 +4,14 @@ cmake_minimum_required(VERSION 3.19)
 string(REGEX REPLACE "([.+?*])" "\\\\\\1" SDIR "${CMAKE_CURRENT_SOURCE_DIR}")
 
 file(GLOB SOURCE_FILES CONFIGURE_DEPENDS *.cpp)
-list(FILTER SOURCE_FILES EXCLUDE REGEX ${SDIR}/main_.*)
+list(FILTER SOURCE_FILES EXCLUDE REGEX ${SDIR}/main_*.*)
```
brakhane: That would match `main______foo.cpp` as well, is this change really necessary? It's a regex, not a glob

WickedEngine/CMakeLists.txt
```diff
@@ -12,17 +12,42 @@ include(GNUInstallDirs)
...
+    option(USE_AVX "Enable AVX" ON)
+    option(USE_AVX2 "Enable AVX2" OFF)
+    option(USE_AVX512 "Enable AVX512" OFF)
+    option(USE_LZCNT "Enable LZCNT" ON)
+    option(USE_TZCNT "Enable TZCNT" ON)
+    option(USE_F16C "Enable F16C" ON)
+    option(USE_FMADD "Enable FMADD" ON)
+else()
+    # If it's not x86, then don't enable x86 features
+    option(USE_SSE4_1 "Enable SSE4.1" OFF)
```
brakhane: I believe the CMake best practice in this case would be to not define the options at all.

WickedEngine/Utility/CMakeLists.txt
```diff
@@ -59,3 +59,9 @@ set_target_properties("Utility" PROPERTIES
 	FOLDER "ThirdParty"
 	POSITION_INDEPENDENT_CODE ${WICKED_PIC}
 )
+
+target_include_directories(Utility PUBLIC
```
brakhane: Wouldn't PRIVATE suffice? PUBLIC means that everything that depends on Utility would also need to include those directories. And `.` should always be included anyways?

WickedEngine/CMakeLists.txt
```diff
@@ -132,6 +176,8 @@ add_library(WickedEngine ALIAS ${TARGET_NAME})
 # aliases for backwards compatibility
 if(WIN32)
     add_library(WickedEngine_Windows ALIAS ${TARGET_NAME})
+elif(APPLE)
```
brakhane: Strictly speaking we don't need that, since there will be no CMake files in the wild which refer to WickedEngine_MacOS. I think leaving that out would be less confusing.

WickedEngine/CMakeLists.txt
```diff
@@ -358,11 +428,11 @@ export(EXPORT Engine-Targets
         FILE "${CMAKE_BINARY_DIR}/cmake/WickedEngineTargets.cmake"
         NAMESPACE WickedEngine::
         )
-
-install(EXPORT Engine-Targets
-        FILE WickedEngineTargets.cmake
-        NAMESPACE WickedEngine::
-        DESTINATION ${INSTALL_CONFIGDIR})
+# I don't quite get why, but this conflicts with specifying the include dirs in Editor
```
brakhane, line 437: IIRC install is currently broken anyway, we can just comment it out for now

CMakeLists.txt
```diff
@@ -27,16 +27,23 @@ option(USE_LIBCXX "Link WickedEngine to llvm libc++ library - only available wit
 option(WICKED_EDITOR "Build WickedEngine editor" ON)
 option(WICKED_TESTS "Build WickedEngine tests" ON)
 option(WICKED_IMGUI_EXAMPLE "Build WickedEngine imgui example" ON)
-option(WICKED_ENABLE_IPO "Enable IPO/LTO in non-debug builds" NO)
+option(WICKED_ENABLE_IPO "Enable IPO/LTO in non-debug builds" ON)
```
brakhane: this was deliberately set to off by default since it increases compile time significantly and was a common gotcha for people trying the engine out for the first time.

CMakeLists.txt
```diff
@@ -27,16 +27,23 @@ option(USE_LIBCXX "Link WickedEngine to llvm libc++ library - only available wit
...
 option(WICKED_EMBED_SHADERS "Embed shaders into the library" NO)
 option(WICKED_ENABLE_RTTI "Enable RTTI" NO)
+if (APPLE)
+    set(WICKED_TESTS OFF) # no main for macos
+    set(WICKED_IMGUI_EXAMPLE OFF)
+    set(WICKED_ENABLE_IPO OFF)
+endif()
 if(UNIX)
```
brakhane: UNIX is true for Apple as well, I think ASAN and the others only work on Linux

WickedEngine/CommonInclude.h
```diff
@@ -422,15 +422,15 @@ constexpr const char* relative_path(const char* path)
 
 constexpr auto relative_path_storage(const char* path)
 {
-	StackString ret;
+	StackString<> ret;
```
brakhane, line 425: is this required for macos or an accidental change?

build_macos.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
```
brakhane: I've heard that modern MacOS ships with zsh. Can we assume that each MacOS also has bash installed? I would assume so, but would like to double check.

Editor/CMakeLists.txt
brakhane: oh, and also mainfoo.cpp, which is the potentially more "gotcha" case

WickedEngine/Utility/CMakeLists.txt
SiebrenW: It's public because this is a library. There is no apparent distinction between public and private headers, so they're all public. In particular I got problems finding metal in common, so I added it here.

WickedEngine/CMakeLists.txt
SiebrenW: I have no idea about that to be honest. I think best practice is to fuse all this off into a separate toolchain file and not have it in the main CMakeLists.txt at all.

WickedEngine/CMakeLists.txt
SiebrenW: I liked the consistency. I searched for all instances of all these and just added the APPLE option wherever I could find it.

CMakeLists.txt
SiebrenW: I also noticed that MacOS really doesn't like this flag either.

CMakeLists.txt
SiebrenW: Ah, forgot another APPLE check and then put the UNIX in the else.

Maybe UNIX is the wrong check to add, since GNU's Not Unix.

WickedEngine/CommonInclude.h
SiebrenW: It didn't compile with Apple clang. I got another error later on switching to llvm clang. I haven't tried llvm clang with this removed again. I don't think it's bad to be explicit that it's a template realisation.

build_macos.sh
SiebrenW: My laptop is fresh as a baby, coming from Linux and at work Windows, and I don't remember installing bash.

Editor/CMakeLists.txt
SiebrenW: I don't know what regex flavour CMake uses. Does just `*` mean "anything" or `.*`? I figured that `main_.<anything>` wouldn't exclude main_Windows.cpp, but I'm probably thinking in the wrong flavour. Will revert

brakhane: In regex, `*` means "repeat the previous char/group 0 or more times". The whole point of the regex is to exclude main_Windows/Linux ;)

WickedEngine/CMakeLists.txt
brakhane: It technically doesn't hurt, I just find it weird to add an alias that was never used and isn't supposed to be used (in the future, we might get rid of the WickedEngine_Windows/Linux targets entirely, but that will be mid/long term)

WickedEngine/CommonInclude.h
brakhane: If it's required for Mac, there's no harm in leaving it in, I just wanted to make sure

WickedEngine/CMakeLists.txt
brakhane: What will happen if someone does `cmake -DUSE_SSE4_1=ON -B build` on MacOS? We shouldn't add options that aren't supported.
I'm open to move the options to somewhere else, but we shouldn't offer options we know are invalid. I'll try to experiment with this later.

WickedEngine/Utility/CMakeLists.txt
brakhane: PUBLIC in CMake means that the options will also added to targets that depend on the specific target, while PRIVATE will only be used to compile that specific target. So if we can't compile targets that use Utility without also including those directories, it would have to be PUBLIC, but I don't think that's the case? At least not under Linux/Windows, as it worked fine without them before.

build_macos.sh
brakhane: hmm, ok. but AFAICT there are no bashisms in the script anyway, so we should just be able to use `#!/bin/sh`. then it wouldn't matter if sh is symlinked to bash, zsh, dash or whatever

Editor/CMakeLists.txt
SiebrenW: I mean in the flavour whether the `.` is required or not. I tested it a bit and now I know what flavour it is, the `.` is required and is not included in the matching text.

build_macos.sh
```diff
@@ -0,0 +1,17 @@
...
+set -e
+
+CURRENT_DIR="$(dirname $0)"
```
brakhane, line 4: cmake will use the current directory by default anyway, so I don't see a particular reason to do this. I don't really mind it being in there, it just feels weird and needlessly complex

WickedEngine/CMakeLists.txt
SiebrenW: Agreed, but even if I remove these options this would still be possible. So I still think adding a separate toolchain file would be best, so even injecting the flags would not be present, if you don't use the wrong toolchain file. You can use cmake presets to automate the right selection, but in my experience they're wayy too verbose.

WickedEngine/Utility/CMakeLists.txt
SiebrenW: The metal include is used in the common library ( WickedEngine/wiGraphicsDevice_Metal.cpp:11 ). Maybe it's not required, or we can move the actual dependencies to another header that is accessible, but that's why I added it. I rather change the build script than the code itself.

WickedEngine/CommonInclude.h
SiebrenW: LLVM clang is fine with it. I think Apple is messing up again.

build_macos.sh
SiebrenW: Agreed. Bash is my default, coming from Linux, but sh would be more correct.

This script was mostly to show what commands I was using, I don't know if it should be in the repo.

build_macos.sh
SiebrenW: This is so I can run the script from anywhere without having to move to that dir. I don't like moving in my terminal. I would be in WickedEngine/WickedEngine (because that's where the shaders are relative to) and compile with `bash ../build_macos.sh --target Editor --target WickedEngine_ext_shaders` and run the Editor like `../build/Editor/Editor`.

Pure laziness.

WickedEngine/Utility/CMakeLists.txt
brakhane: If it's used by wiGraphicsDevice, we should make sure to specify the metal include path (also) in the WickedEngine_MacOS target (as `PRIVATE`). Relying on a side effect that Utility will cause it to get added because it's marked as PUBLIC is confusing and makes maintenance harder, IMO. Especially for people like me who don't have a mac and want to maintain/adjust the CMake files without breaking it for others

build_macos.sh
brakhane: yeah, sorry, I had a brain fart and interpreted dirname $0 as current dir, not the script path. `cd $(dirname $0)` would have been another way. Either way I don't really mind.

WickedEngine/CommonInclude.h
brakhane: XCode clang must also be fine with it, otherwise it wouldn't work there. I don't mind it being there if it's necessary, but in the end, it's János' call to make.

WickedEngine/CMakeLists.txt
SiebrenW: I think the generated files should all be put into build (aswell as the compiled shaders) so you don't have to run Editor from inside the WickedEngine dir, but the build dir. (if that makes sense)

The installation can be a pain with CMake. Maybe I can look at it sometime, since I have some experience with that, but mostly for linux specifically and statically linked libraries.

WickedEngine/CMakeLists.txt
brakhane: Im my experience, everything in CMake is way too verbose ;)

brakhane:
@SiebrenW BTW, check out the build failures, there currently are some problems on Linux GCC (invalid flag gets added now) and on Windows.

(And since my communication style often catches people off guard: your work on this PR is appreciated)

WickedEngine/Utility/CMakeLists.txt
SiebrenW: I looked how the engine includes vulkan/d3d and it's just with `Utility/metal/`, so I changed it accordingly (though I wouldn't call it correct....). Good that you mentioned it.

CMakeLists.txt
SiebrenW: For me on Linux it didn't matter much for compile times. But I use mold as a linker.

`mold -run cmake --build build --target Engine`

The default linker on MacOS is decently fast, I'm impressed, but I didn't even try and immediately used lld when switching to llvm.

WickedEngine/CMakeLists.txt
```diff
@@ -132,6 +164,8 @@ add_library(WickedEngine ALIAS ${TARGET_NAME})
 # aliases for backwards compatibility
 if(WIN32)
     add_library(WickedEngine_Windows ALIAS ${TARGET_NAME})
+elif(APPLE)
```
brakhane:
```suggestion
elseif(APPLE)
```

Editor/CMakeLists.txt
```diff
@@ -6,8 +6,14 @@ string(REGEX REPLACE "([.+?*])" "\\\\\\1" SDIR "${CMAKE_CURRENT_SOURCE_DIR}")
 file(GLOB SOURCE_FILES CONFIGURE_DEPENDS *.cpp)
 list(FILTER SOURCE_FILES EXCLUDE REGEX ${SDIR}/main_.*)
 list(FILTER SOURCE_FILES EXCLUDE REGEX ${SDIR}/stdafx.*)
-list(APPEND SOURCE_FILES main_${PLATFORM}.cpp)
 
+if (APPLE)
+    list(APPEND SOURCE_FILES main_MacOS.mm)
+    # PCH has disabled objective C, so don't use PCH
```
brakhane: btw, what do you mean by "has disabled objective C", do you mean they are broken in obj C?

WickedEngine/CMakeLists.txt
```diff
@@ -172,60 +206,62 @@ set(WICKEDENGINE_STATIC_LIBRARIES
...
-if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
-    if ("${CMAKE_VS_PLATFORM_NAME}" STREQUAL "x86" OR "${CMAKE_VS_PLATFORM_NAME}" STREQUAL "x64")
-        if (USE_AVX512)
-            target_compile_options(WickedEngine_common PUBLIC /arch:AVX512)
-        elseif (USE_AVX2)
-            target_compile_options(WickedEngine_common PUBLIC /arch:AVX2)
-        elseif (USE_AVX)
-            target_compile_options(WickedEngine_common PUBLIC /arch:AVX)
-        endif()
+if (CPU_ARCH STREQUAL "x84")
```
brakhane:
```suggestion
if (CPU_ARCH STREQUAL "x86")
```

WickedEngine/CMakeLists.txt
```diff
@@ -12,17 +12,31 @@ include(GNUInstallDirs)
 set(INSTALL_LOCAL_CONFIGDIR "${CMAKE_BINARY_DIR}/cmake")
 set(INSTALL_CONFIGDIR "${CMAKE_INSTALL_LIBDIR}/cmake/WickedEngine")
 
+if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
+    if ("${CMAKE_VS_PLATFORM_NAME}" STREQUAL "x86" OR "${CMAKE_VS_PLATFORM_NAME}" STREQUAL "x64")
+        set(CPU_ARCH "x86")
```
brakhane: I'm not a fan of changing x64/x86_64 to x86, they are seperate for a reason.

Since CPU_ARCH only seems to be used to check if we provide the SSE etc. options, make just having a variable `IS_X86_ARCH` instead, or even better, name it `CPU_ARCH_CLASS` or something, to make it clear that it's a potentially simplified value.

WickedEngine/CMakeLists.txt
```diff
@@ -172,60 +206,62 @@ set(WICKEDENGINE_STATIC_LIBRARIES
...
-        target_compile_options(WickedEngine_common PUBLIC -mlzcnt)
-    endif()
-    if (USE_TZCNT)
-        target_compile_options(WickedEngine_common PUBLIC -mbmi)
+    if (USE_SSE4_1 OR USE_SSE4_2)
+        target_compile_definitions(WickedEngine_common PUBLIC _XM_SSE4_INTRINSICS_)
     endif()
     if (USE_F16C)
-        target_compile_options(WickedEngine_common PUBLIC -mf16c)
+        target_compile_definitions(WickedEngine_common PUBLIC _XM_F16C_INTRINSICS_)
```
brakhane, line 225: why remove the option? The idea is that if we want to use F16C, we want to use it everywhere, not just in the directx math implementation. Without the `-mf16c` option, the compiler will not create F16C code.

The define should not be necessary in that case. It's better to have another "if not mac" guard for this.

WickedEngine/CMakeLists.txt
```diff
@@ -172,60 +206,62 @@ set(WICKEDENGINE_STATIC_LIBRARIES
...
     endif()
     if (USE_FMADD)
-        target_compile_options(WickedEngine_common PUBLIC -mfma)
```
brakhane, line 225: same as above.

Editor/CMakeLists.txt
SiebrenW: There was some warning about saying "disabled ojective C" related to the precompiled headers. I didn't really find it necessary to find out how to re-enable it, so I just disabled PCH for the objective C files.

brakhane: My point is that the comment is very confusing, and I have no idea what it's supposed to tell me. "We don't use PCH on Mac because PCH disables objective C". Could you maybe recheck what the warning says specifically? According to my (admittedly very short) research objective C does support PCH, so I'm a bit confused why it needs to be disabled.
I'm fine with disabling it if it doesn't work in our case for some reason, but we should document the exact reason (just so we/somebody else can re-enable it in the future when that reason doesn't exist anymore)

For context: I've worked with quite a few projects that had workarounds like this that stopped being needed for 5 years, but everybody had assumed they still are. That's why I'm a bit anal about carefully documenting special cases.

SiebrenW: Yes, sorry. I'm at work and will be busy/not have access to my Mac until the afternoon (CET) tomorrow, but I was hoping this would be at least a bit informative.

If I had my mac on me I would have elaborated better with snippets and all. Probably just fixed it 😐 .

Thank you for taking the time to review by the way. My intent was to have it as a throwaway hack to get it working, but I'll take it a bit more seriously (hopefully) tomorrow during my break.

SiebrenW: The error in particular is error:

```
Objective-C was disabled in precompiled file '.../WickedEngine/build/Editor/CMakeFiles/Editor.dir/cmake_pch.hxx.pch' but is currently enabled
1 error generated.
```

I think the real issue is down to the fact that the precompiled header is a pure C++ header, not objective C. I couldn't find a way to enable it for both, so I think disabling it has the least impact. Then we have 1 file built without PCH, which isn't such a big deal I think.

WickedEngine/CMakeLists.txt
SiebrenW: It was already in the x86 block though. I didn't look too closely to be honest and just disabled everything I thought had to do with x86. I'll re-enable it if I see it working on arm.

SiebrenW: ~~I see it's not working at all, so I'll keep it in that block. I think it has to do with the CPU arch and not so much the OS. I presume you'd get the same error on arm on Linux, but I can't test that out for now (someone with a raspberry pi or some other SBC can check maybe?)~~

It works. It's really just directXmath not working

Editor/CMakeLists.txt
brakhane: According to [this thread](https://discourse.cmake.org/t/target-precompile-headers-with-macos-and-objective-c-mm-files/3205/3) it could be all that's needed is adding `LANGUAGES C CXX OBJC OBJCXX` to the `target` line(s)

brakhane: We can probably omit `C` but better first test it with to make sure it's not required when using OBJC

WickedEngine/CMakeLists.txt
brakhane: -mf16c (as well as all the others) are X86 only flags, they are not supported on other architectures.

(All GCC `-m` flags are architecture specific)

Editor/CMakeLists.txt
SiebrenW: Interesting. That did work.

Now the question is: do we really want this? I don' know what the impact is on other targets. I can simply throw it over the fence and see what the pipeline does I suppose.

WickedEngine/CMakeLists.txt
SiebrenW: It didn't catch my elseif case for some reason, but I fixed it. I can confirm that indeed the mf16c flag does not work. Apparently f16 and fltmultadd are part of the base instruction set of ARM, so no additional flags needed.

WickedEngine/CMakeLists.txt
SiebrenW: I'll go for the boolean for now. If we want to add specific ARM flags, then it can be changed to BASE_CPU_ARCH or something.

Later I think the project should just use toolchain files for these things. It will be a lot less hassle with if-else etc. but something needs to be thought of for Windows perhaps.

Editor/CMakeLists.txt
brakhane: I think we would need `LANGUAGES CXX` on Windows/Linux (maybe `LANGUAGES C CXX`, not sure) and add objective C for mac, something like
```cmake
if(APPLE)
set(OBJ_C_LANGS OBJC OBJCXX)
endif()
project(... LANGUAGES C CXX ${OBJ_C_LANGS})
```
(I don't quite remember from the top of my head if CMake treats undefined vars as empty string, if not, we'd just use a `set(LANGUAGES C CXX)` for linux and the 4 for mac)

SiebrenW: It's not working for all other platforms, I will stick with the exclusion.

PR #1598, opened 2026-04-16
Add UTF-8 compile option for MSVC (to prevent compilation breaks depending on the building computer locale)

rapha-s:
Hi!~

Tried to build and was getting a lot of **"newline in constant"** inside **wiInput.cpp**, complaining specifically about the lines with arrows (→) and other uncommon characters inside strings, so added this (**/utf-8**) to make sure **MSVC** will treat the source file as UTF-8, doesn't matter what locale the user's computer is running at, so compilation can be always successful.


brakhane:
Thanks for the PR.

Can you also add the flag to the VS solution files? The "official" way to build Wicked on Windows is to use the provided solution files, CMake is currently mostly for Linux (though I try to keep it working on Windows as well)

rapha-s:
Absolutely ^^
Done~

brakhane:
Didn't have time to check the MSVC stuff yet, but the CMake stuff LGTM.

turanszkij (approved):

skinkairewalker:
how can i fix it ?

`Erro	C2001	newline constant	WickedEngine_Windows	C:\Users\Waldemar Lima\Downloads\WickedEngine-0.72.33\WickedEngine\wiInput.cpp	1302	`

PR #1604, opened 2026-05-16
Add arithmetic expression evaluation to TextInputField

### LINKED ISSUE #1476: Allow aritmentic operations on numeric inputs
For numeric inputs in the Editor, it would be cool if the user could insert arithmetic operations, for example, if the input has value `10` and the user types `* 2` (`10 * 2`) then the value would be `20` when the user commits the change (presses enter).

srktnc:
Numeric input fields in the editor now evaluate arithmetic expressions when the user presses Enter. For example, if the field shows '10' and the user appends '* 2' to get '10 * 2', the committed value becomes 20.

Supported operators: +, -, *, /, parentheses, unary +/-, floating-point literals and scientific notation.

Non-numeric / plain-text inputs (e.g. 'reset', 'FLT_MAX', object names) fall back to the original raw-string behaviour, so all existing special- case handling in sliders and other widgets is fully preserved.

Closes #1476

### DIFF
--- a/WickedEngine/wiGUI.cpp
+++ b/WickedEngine/wiGUI.cpp
@@ -1647,6 +1647,77 @@ namespace wi::gui
 		}
 		return false;
 	}
+
+	// Simple recursive-descent arithmetic expression evaluator.
+	// Supports: +, -, *, /, parentheses, unary +/-, floating-point literals.
+	// Returns true and sets 'result' if 'str' is a fully-consumed valid expression.
+	// Returns false for non-numeric input (e.g. plain text strings) so callers fall back.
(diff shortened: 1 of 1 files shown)


turanszkij:
This is pretty cool, thanks.

turanszkij (approved):

srktnc:
> This is pretty cool, thanks.

I absolutely love Wicked Engine. Exploring the codebase and adding small things to it has been really enjoyable for me. The more I learn about its structure, the more I want to contribute and build on top of it.

Thank you for creating and maintaining such a great project. 💯🚀

PR #1608, opened 2026-05-20
Editor: add drag-and-drop entity reordering in hierarchy tree (#1084)

### LINKED ISSUE #1084: Editor QOL: Re-order scene elements
A somewhat minor request, but for organizational sake, I think it would be nice to be able to move and re-order scene entities in the list.

srktnc:
- Extend wi::gui::TreeList with drag-and-drop support:
  - Track drag start on item click with 4px activation threshold
  - Compute drop target position and render a horizontal indicator line
  - Fire OnReorder callback with source/target indices on mouse release

- ComponentsWindow: maintain per-parent custom entity display order
  - entity_user_order stores custom ordering keyed by parent entity
  - ApplyUserOrder() applies stored order when sorting mode is None (0)
  - OnReorder callback reorders siblings and refreshes the tree
  - Auto-sorting modes (by name / by type) are unaffected

Fixes #1084

### DIFF
--- a/Editor/ComponentsWindow.cpp
+++ b/Editor/ComponentsWindow.cpp
@@ -122,6 +122,100 @@ void ComponentsWindow::Create(EditorComponent* _editor)
 	entityTree.OnDoubleClick([this](wi::gui::EventArgs args) {
 		editor->FocusCameraOnSelected();
 		});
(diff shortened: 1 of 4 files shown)


turanszkij (approved):

turanszkij:
This is a nice feature, thanks for the work.

PR #1612, opened 2026-05-21
Fix/issue 1464 drag drop entities

### LINKED ISSUE #1464: [feature] Drag and drop to organize entities
It would be very useful to be able to organize entities in the entities panel by drag and dropping them.

Currently to nest entities the user has to add a `Hierarchy` component and then select the parent. It would be easier and more intuitive if this component didn't exist and the possibility of nesting entities was builtin in all entities.

srktnc:
Extend wi::gui::TreeList drag-and-drop with hierarchy re-parenting:

Track drop-on item targets separately from between-item reorder targets
Use a three-zone item drop area:
- top: insert before item
- middle: re-parent as child of item
- bottom: insert after item
Track target hierarchy level for between-item drops and draw the indicator at that level
Allow child entities to be moved up one level by dragging to the indicated outdent line
Avoid treating level-changing drops as no-op reorders

ComponentsWindow: apply hierarchy changes from tree drag-and-drop:

Drop-on attaches the dragged entity to the target entity
Between-item drops reorder within the resolved parent/sibling group
Between-item outdent drops detach or attach to the resolved parent level
Prevent circular hierarchy drops
Skip no-op direct-child reattachments
Record hierarchy changes in undo/redo history
Auto-expand newly assigned parent nodes after drop

HierarchyWindow: refresh parent display after drag-and-drop:

Rebuild the parent combo even when the selected entity stays the same
Keep the Hierarchy component parent field in sync after tree re-parenting

Fixes #1464

### DIFF
--- a/Editor/ComponentsWindow.cpp
+++ b/Editor/ComponentsWindow.cpp
@@ -123,6 +123,58 @@ void ComponentsWindow::Create(EditorComponent* _editor)
 		editor->FocusCameraOnSelected();
 		});
 	entityTree.OnReorder([this](wi::gui::EventArgs args) {
+		// Re-parent: drop ON another entity (bValue == true)
+		if (args.bValue)
+		{
+			int source_idx = args.iValue;
+			int target_idx = args.iValue2;
+			if (source_idx < 0 || source_idx >= entityTree.GetItemCount())
+				return;
+			if (target_idx < 0 || target_idx >= entityTree.GetItemCount())
+				return;
+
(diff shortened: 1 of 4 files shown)


turanszkij (approved):

turanszkij:
Thank you for the work, this was a much requested feature!

srktnc:
> Thank you for the work, this was a much requested feature!

Thank you so much! 😊 I’m really happy to have contributed, even in a small way, to such an awesome project. Hopefully I can make bigger contributions in the future! 🙌

PR #1613, opened 2026-05-21
add deprecated attribute

brakhane:
Add `[[ deprecated ]]` to deprecated stuff so the IDE will warn.

The `WI_DISABLE_DEPRECATED` macro is not strictly necessary, as we only need it in one place, but it might be used more than once in the future; without it, we'd still need the `#ifdef` and `#pragma`s at that one location to prevent a warning, so we don't save any LOC.

### DIFF
--- a/WickedEngine/CommonInclude.h
+++ b/WickedEngine/CommonInclude.h
@@ -21,6 +21,14 @@
 #define NO_SANITIZE(x)
 #endif // defined(__GNUC__) || defined(__clang__)
 
+#ifdef _MSC_VER
+#define WI_DISABLE_DEPRECATED_BEGIN __pragma(warning(push)) __pragma(warning(disable: 4996))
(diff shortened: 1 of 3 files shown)


turanszkij (approved):

turanszkij:
This causes a lot of warnings now in engine build because of helper functions that convert from deprecated to new structures. Any way to fix it without removing deprecated tag?

brakhane:
Hmm, those didn't show up for me, but it might be a MSVC thing. You can enclose them within `WI_DISABLE_DEPRECATED_BEGIN`/`WI_DISABLE_DEPRECATED_END` blocks like it's done in RenderPassDesc

brakhane:
Actually now that I think of it, `WI_ALLOW_DEPRECATED` would be a better name. Feel free to rename it alongside your changes

PR #1615, opened 2026-05-22
cleanup: use NoHeapAlloc mixin instead of copy pasting

brakhane:
No functional change, just avoids copy-paste and the potential problems that come with it if it needs to get changed in the future

### DIFF
--- a/WickedEngine/wiGraphics.h
+++ b/WickedEngine/wiGraphics.h
@@ -860,6 +860,21 @@ namespace wi::graphics
 		ShaderStage stage = ShaderStage::Count;
 	};
 
+	// Mixin: dynamic allocation and destruction of this object is not allowed because virtual table is not used. Placement new is allowed.
+	struct NoHeapAlloc
+	{
+		static void* operator new (size_t) = delete;
+		static void* operator new[](size_t) = delete;
+		static void  operator delete (void*) = delete;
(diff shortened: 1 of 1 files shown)


brakhane:
Weird, that MacOS failure feels like a compiler bug. Why is `*video = {}` trying to instantiate NoHeapAlloc? I'll have to analyze that one a bit more.

(Edit: because of the Texture member, but still this doesn't seem right)

turanszkij:
I prefer if there is no double inheritance.

brakhane:
Hmm, I don't get why, but I'll have to accept that. What if I create a GPUResourceNoHeapAlloc class instead?

turanszkij:
I think I tried adding them to GPUResource but it didn't work the same way as now

turanszkij:
But I tried quickly now and it seems to work if I just add these to GPUResource struct. (It will be compile error to make GPUBuffer with "new")

brakhane:
This would prevent all GPUResources from being heap allocated though, I thought it was only needed for some?

turanszkij:
There are only Buffers, Textures and RaytracingAccelerationStructures and all of these disabled the heap allocations. I can't remember why I didn't put it in GPUResource but I think there was a reason. Seems to work that way now, so maybe it's fine.

PR #1649, opened 2026-06-17
Documentation + tooling: a type-checked Lua scripting API reference that doubles as editor IntelliSense

Anthony-Gaudino:
This reworks the Lua scripting documentation so that the reference **is** valid, LuaLS-annotated Lua. The same files humans read can be turned, by a tiny included script, into autocomplete + type-checking stubs for any Wicked Engine Lua project, no separate, drifting stub set to maintain.

## The problem

`ScriptingAPI-Documentation.md` was one large file describing the API as prose lists (`Method(Type p, opt Type q = d) : Ret`). That format has two costs:

* **It drifts.** Because it isn't checked against anything, small errors accumulate — wrong return types, parameters that are actually optional, enum constants without real values, bindings that exist in C++ but were never documented (and vice versa).
* **It can't help your editor.** Script authors get no autocomplete, no signature help, and no type checking, so the same mistakes get rediscovered at runtime.

## What this PR does

**1. The reference is now annotated Lua.** Each section keeps its prose intro, but the API itself lives in ```lua blocks using LuaLS/EmmyLua annotations (`---@param`, `---@return`, `---@class`, `---@enum`, …). It reads at least as clearly as before and is now machine-verifiable.

**2. A full C++ audit, baked in.** Every binding was reconciled against its `*_BindLua.cpp` source. Along the way this corrected real inaccuracies: return types (e.g. dot product returning a scalar), optional parameters that were marked required, enum/constant values that were placeholders, entity handles now typed via an `Entity` alias, plus missing component constructors, global functions, and return values that were never documented. The reference is meaningfully more accurate than before, independent of the tooling angle.

**3. Navigable layout.** The single file becomes `scripting_api/index.md` (intro + table of contents) linking to one file per topic under `sections/`. Easier to read on GitHub, easier to diff, easier to review.

**4. Optional, zero-dependency tooling.** `generate_stubs.py` (pure Python 3) extracts the code blocks into LuaLS definition files, one stub per section. Point your `.luarc.json` at the output folder and the whole scripting API lights up with autocomplete, signatures, and type checking in VS Code / any LuaLS editor. `README.md` walks through it. Generated files are git-ignored: the Markdown stays the single source of truth, so docs and stubs can never disagree.

## Why merge it

* **Lower maintenance, not higher.** One source of truth that is validated. When a binding changes, you update one annotated block; the docs and the IntelliSense stubs update together.
* **Correctness is enforced.** `lua-language-server --check` passes with no problems and every block passes `luac -p`. Future edits get the same safety net for free.
* **Better day-one experience for script authors.** Autocomplete and type checking catch typos and wrong argument types before the game runs.
* **No workflow imposed.** The tooling is opt-in; the docs render fine on GitHub with or without running anything.

## Notes for reviewers

* The old `ScriptingAPI-Documentation.md` is removed and fully superseded by `scripting_api/`.
* Nothing in the engine or its bindings changes, this is documentation and a standalone helper script only.


turanszkij:
Thanks for the work, there are some good ideas here and nice that you made a lot of commenting on all the features.

1) Wouldn't it be better to have the documentation itself in lua instead so we don't need to generate anything with python just use it as-is? (seems like it is already mostly in lua)

2) I want to have it in one file for easy searching with Ctr+F, and no directory structure.

3) I want there to be no README for documentation, documentation must be simply useable without any other readme. However the documentation itself could contain comments about how to set it up in an editor as an intellisense.

Anthony-Gaudino:
I think we can reach a middle ground.

1. Markdown is better for documentation because it supports prose formatting, links, and future extensibility (diagrams, images, etc.). Raw Lua code alone loses these benefits. The distinction keeps each format doing what it does best.

2. Rather than maintaining a 10,000-line file, we keep the documentation organized in markdown sections and generate a single `wicked_engine_bindings.lua` file from them. You can Ctrl+F and use it directly as your stubs, while the source stays organized.

3. We can fold `index.md` into `README.md`, one file that serves as both the documentation entry point and setup guide.

This way: organized source that's pleasant to maintain + single searchable artifact for end users + no extra documentation overhead.

turanszkij:
1. I see, makes sense
2. I want to keep the same one file for documentation, the contents can change but the one file must stay
3. anything related to script doc must go into that script documentation file we already have ( https://github.com/turanszkij/WickedEngine/blob/master/Content/Documentation/ScriptingAPI-Documentation.md )

Anthony-Gaudino:
* Changed the script to generate a single `wicked_engine_bindings.lua` file.
* Folded `README.md` and `index.md` into `ScriptingAPI-Documentation.md`
* Added latest documentation changes from master.

turanszkij:
There are still separate files in the sections folder.

Anthony-Gaudino:
Now the documentation is consolidated into a single file.

turanszkij (approved):

turanszkij:
Thanks for the work, I think this looks great now.

PR #1659, opened 2026-06-25
perf(terrain): parallelize chunk generation across the job pool

Anthony-Gaudino:
Terrain generation was slow to populate and struggled with large radii because the whole spiral was generated one chunk at a time on a single background job, leaving the worker pool underused.

Two changes, both in `Generation_Update()`:

- Per-chunk tail parallelism: the grass build (CreateFromMesh) and the heightfield physics-shape build now run as job-pool tasks alongside mesh.CreateRenderData() instead of serially on the controlling worker. Both read the mesh as const and write disjoint targets, so they are safe to run concurrently.

- Batched, cross-chunk parallel generation: request_chunk is split into setup_chunk (serial scene/ECS component creation), build_chunk (the heavy vertex/mesh/texture/grass/physics work, run in parallel across a bounded batch), and place_chunk (serial grass/prop placement). All scene component creation is done serially up front per batch so the component managers stay stable while builds run; jobsystem Wait does work-while-waiting so the nested dispatches cannot deadlock. The batch is placed immediately after it builds, so props and grass stream in progressively alongside the terrain. Single-/few-chunk edits still use the existing fine-grained inner vertex dispatch, preserving edit latency. invalidated is cleared at (re)generation time so a budget-limited frame cannot leave a built chunk flagged for rebuild.

Only tested on Linux!


turanszkij:
Just tested it, even just looking at terrain gen it looks much faster. I'll take some time to test it more and to fully understand your optimizations, thanks for the work.

WickedEngine/wiTerrain.cpp
```diff
@@ -1415,62 +1463,190 @@ namespace wi::terrain
...
+			};
 
-				if (generated_something && timer.elapsed_milliseconds() > generation_time_budget_milliseconds)
+			// Generate chunks in batches so several build concurrently across
+			//  the job pool. setup_chunk() (which mutates the scene) runs
+			//  serially up front for the whole batch, so the component managers
+			//  stay stable while the build jobs run; the batch is then placed
+			//  (grass/props) before moving on, so props stream in progressively
+			//  alongside the terrain.
+			wi::vector<Chunk> build_batch;
```
turanszkij: This temporary vector creation will always heap allocate that should be avoided to happen every frame. To avoid this, one method is to make it static so its allocation can be reused the next time the generator runs. The static vector will just needed to be cleared() each time. Or there is also the StackVector utility that is a simplified vector that never allocates and has compile-time max size, or just using static array.

turanszkij (changes requested):

WickedEngine/wiTerrain.cpp
Anthony-Gaudino: Please let me know if 0eb406c80a0475fed2affeb74625f2291162e426 is satisfactory.

turanszkij: Yes it's perfect now, thanks.

turanszkij: There is some build issue with the windows-cmake builder around the flush_build_batch lambda. I don't know why, all the other builds are fine. I think it has a problem with calling the flush_build_batch from within flush_build_batch, if you wanna try fix it go ahead, or you can finalize the pull request too, I can take a look later it's not too important.

(Don't mind the macos and ios build those are failing due to some github issue only)

Anthony-Gaudino: I made a commit to try to fix the Windows build.

turanszkij (approved):

WickedEngine/wiTerrain.cpp
turanszkij: Thanks for the work.

PR #1662, opened 2026-06-27
Ocean performance improvements

Anthony-Gaudino:
Four independent optimizations reducing ocean GPU cost and memory use.

### FFT: pack 3 fields into 2 via two-for-one real packing
The ocean inverse-FFTs three Hermitian frequency fields to real spatial outputs.
Pack height + Dx into one complex field's real/imaginary parts and Dy into a
second, reducing the batched FFT from 3 slices to 2. (~33% FFT reduction)

### Gate displacement readback on actual CPU queries
Only copy the displacement map GPU->CPU while something queries ocean height;
skip the copy when idle. (~4 MB/frame when unused; ~2 MB VRAM savings)

### Drop per-frame constant-buffer barriers
Use BindDynamicConstantBuffer instead of persistent buffer with per-frame
Barrier/UpdateBuffer/Barrier. (Two barriers + one update eliminated per-frame)

### Store displacement map as RGBA16F instead of RGBA32F
Zero-mean displacement values only need 16-bit float precision for both rendering
and CPU buoyancy. Halves bandwidth and memory across all consumers.
(~50% bandwidth/VRAM for displacement; half-float readback decoding added)

### Skip foam noise where there is no foam
Gate 6 procedural noise evaluations behind [branch] on foam_combined, skipping
them across open water where foam_combined ~= 0.

**Only tested on Lunux**

### DIFF
--- a/WickedEngine/shaders/ShaderInterop_Ocean.h
+++ b/WickedEngine/shaders/ShaderInterop_Ocean.h
@@ -21,8 +21,10 @@ CBUFFER(OceanCB, CBSLOT_OTHER_OCEAN)
 	uint xOceanOutWidth;
 
 	uint xOceanOutHeight;
-	uint xOceanDtxAddressOffset;
-	uint xOceanDtyAddressOffset;
+	// Element offset to the second packed FFT field (Dy). The first field packs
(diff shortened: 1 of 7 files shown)


turanszkij:
These are good!

turanszkij (approved):

PR #1663, opened 2026-06-27
Particle system perf improvement

Anthony-Gaudino:
## Summary

Comprehensive performance optimization of the hair/grass particle system across CPU simulation, GPU shaders, and ray-tracing acceleration structures. These changes reduce per-frame GPU compute cost, CPU constant-buffer overhead, and ray-tracing traversal cost for large grass fields.

**Impact:** 60+ FPS maintained during fast camera movement through dense grass with RT diffuse enabled (previously tanked to ~8 FPS).

### GPU Simulation Shader Optimizations

**Skip geometry generation for distance-culled strands** — strands beyond `viewDistance` produce no visible indices and their render buffers are never read. Skip billboard/patch transforms and render-buffer writes for culled strands, keeping only RT positions valid.

**Skip raytracing buffer writes when no BLAS consumes them** — the raytracing position buffer is only read when a BLAS exists (raytracing active). Gate all RT writes behind a new `HAIR_FLAG_RAYTRACED` flag; when raytracing is off, skip the per-vertex float4 write entirely.

**Skip atlas-frame noise for single-rect systems** — the common case (default single texture, no sprite-sheet atlas) wastes a 3D gradient-noise evaluation whose modulo result is always 0. Guard the noise behind `if (xHairAtlasRectCount > 1)`.

**Skip identity matrix multiply for billboard 0** — billboard 0 always has size 1 and rotation 0, so its variation matrix is the identity; `mul(identity, TBN)` reduces to `TBN`. For `billboardCount == 1` (the default), this eliminates a full 3×3 matrix multiply per root and per segment.

**Hold distance-culled strands at rest pose instead of simulating** — culled strands skip their entire per-segment physics integration (inertia, wind sample, collider loop). Only the rest pose is stored, so re-entry is settled with no pop. The dither-fade ensures the simulated↔rest transition is never visible.

### CPU Constant-Buffer Optimization

**Skip per-frame CB upload when unchanged** — cache the bytes of the last uploaded `HairParticleCB` and only call `UpdateBuffer` + push the `COPY_DST→CONSTANT_BUFFER` barrier when contents differ. For static grass, this eliminates a per-hair GPU copy and barrier every frame. The buffer is `Usage::DEFAULT`, so contents persist when skipped.

### Ray-Tracing Acceleration Structure Optimization

**Build hair BLAS for fast traversal, not fast build** — grass is heavily traced by GI rays, so use `FLAG_PREFER_FAST_TRACE` (high-quality, fast-to-traverse BVH) instead of `FLAG_PREFER_FAST_BUILD` (cheap-to-build, slow-to-traverse).

**Periodic full rebuild + adaptive rebuild on camera motion** — refitting the BLAS every frame indefinitely degrades quality (refit can't fix topology). Do a full rebuild every 60 frames (staggered per system), with refits in between. Additionally, rebuild whenever the camera moves past a threshold (1.0 world units) since the last rebuild, so fast movement keeps the BVH fresh and prevents the FPS  tank during navigation.

**Spread culled strands across terrain instead of collapsing to one point** — distance-culled strands collapse their RT vertices to the object-space origin (a shared point), piling up a coincident-triangle cluster that rays must traverse. Collapse each culled strand to its own emitter position instead: triangles stay zero-area (far grass still doesn't contribute to GI) but spread across terrain, so the BVH stays well-distributed and stable across cull-boundary crossings.

**Include grass in RT diffuse (with escape hatch)** — grass can now stay in diffuse GI at stable cost thanks to the BLAS improvements. A `raytracing_inclusion_mask_diffuse` bit is available so grass can still be opted out per target if needed (one-line uncomment in `wiScene.cpp`).

**Only tested on Linux**


WickedEngine/wiHairParticle.cpp
```diff
@@ -437,7 +444,8 @@ namespace wi
 			{
 				desc = material.textures[MaterialComponent::BASECOLORMAP].resource.GetTexture().GetDesc();
 			}
-			HairParticleCB hcb = {};
+			HairParticleCB hcb;
```
turanszkij: Just use `hcb = {};` instead and no need to include the cstring as well.

WickedEngine/wiRenderer.cpp
```diff
@@ -5752,13 +5752,39 @@ void UpdateRaytracingAccelerationStructures(const Scene& scene, CommandList cmd)
...
-					if (hair.must_rebuild_blas)
+					// The grass BLAS is built for fast traversal, but refitting it
+					// every frame as strands animate and cull progressively
+					// degrades that quality. Do a full rebuild periodically
+					// (staggered per system so the cost is spread across frames)
+					// and refit on the frames in between.
+					constexpr uint64_t blas_rebuild_interval = 60;
+					const bool periodic_rebuild =
+						((device->GetFrameCount() + i) % blas_rebuild_interval) == 0;
+					if (hair.must_rebuild_blas || periodic_rebuild || motion_rebuild)
```
turanszkij: I think this interval of rebuilding the blas every 60 frames doesn't really help, because now the performance will just fluctuate every 60 frames. I think having it be rebuilt every frame is better, then we can better see what actual performance it has and we can tweak the grass density to the performance easier.

turanszkij (changes requested):

WickedEngine/wiRenderer.cpp
Anthony-Gaudino: This change is reverted.

WickedEngine/wiHairParticle.cpp
Anthony-Gaudino: Should be fixed.

WickedEngine/wiHairParticle.h
```diff
@@ -117,6 +117,9 @@ namespace wi
 		wi::graphics::Format position_format = wi::graphics::Format::R16G16B16A16_UNORM;
 		mutable bool must_rebuild_blas = true;
 		mutable bool gpu_initialized = false;
+		// Cached bytes of the constant buffer last uploaded to the GPU, used to
+		// skip redundant per-frame uploads while the contents are unchanged.
+		mutable wi::vector<uint8_t> constantBufferData;
```
turanszkij: Instead of a vector which will be a separate allocation, you can use `HairParticleCB constantBufferData = {};` here.

turanszkij (changes requested):

WickedEngine/wiHairParticle.h
Anthony-Gaudino: https://github.com/turanszkij/WickedEngine/pull/1663/commits/7d002a4909338a29120f2d3a607c2cbccf1e99ab should address this.

turanszkij (approved):

PR #1671, opened 2026-07-04
Surfel GI: performance overhaul, quality improvements, and stability fixes

Anthony-Gaudino:
## Summary

A broad pass over the Surfel GI subsystem: performance optimizations, quality and convergence improvements, a large set of newly exposed tuning parameters, and a batch of stability/correctness fixes. Touches the surfel shaders (update, raytrace, integrate, coverage, binning, denoise), the renderer plumbing, and the editor.

## Tunable parameters

Almost every hard-coded magic number in the pipeline has been pulled out into named, documented `static const` constants in `WickedEngine/shaders/ShaderInterop_SurfelGI.h`. This covers the cascaded grid, radius/spacing, recycling, coverage temporal filter, spawn budget/placement skips, ray budget/boost/amortization, and the denoiser, so the whole system is live-tunable from one place.

## Performance

- **Ray sorting** by surfel-origin locality (Morton key) for BVH-coherent tracing.
- **Temporal ray amortization**: distant, low-frequency surfels re-trace on a cadence that grows with distance (staggered per-surfel to avoid pulsing), extended past the top cascade level for bird's-eye views.
- **Temporal coverage accumulation** with history reprojection and disocclusion + distance guards, so the gather is amortized across frames without smearing under motion.
- **Per-subtile coverage gather-skip** and **cooperative LDS load** of the shared base-level cell in the coverage pass.
- **Ray count scaled down with cascade level** (far surfels get fewer rays).
- **Relevance-based recycler** bounding the live working set; skip surfel placement on transparent, fully-metallic, and strongly-emissive surfaces.
- Raytrace dispatch is bounded to the ray budget; per-pass GPU profiler ranges added for measurement.

## Quality / convergence

- **Cascaded multi-resolution grid**: distance-scaled surfel radius/spacing (constant screen-space size, sparser far / unchanged near), extended to sub-unit levels for near-field density.
- **Edge-aware à-trous denoiser** on the half-res GI output (normal+depth guided, output-side only so it can't feed back into the cache) to smooth per-surfel radiance variance on flat surfaces.
- **SH-guided bounce sampling** for faster convergence.
- **Pop-free spawns**: newborns seeded from the neighbor cache (clamped to the brightest neighbor to avoid runaway white) and faded in over frames, killing black/white spawn popping.
- **Tangent-plane anti-leak** in the coverage apply; spacing-driven placement with thinning so over-dense far regions relax back to target.

## Bug fixes

- **Black/white rectangles on foliage**: the shaded pixel's `surface.N` could be NaN (`normalize(0)` on degenerate/cancelling vertex normals—foliage cards, billboards). Guard the normal in the coverage gather with a camera-facing fallback.
- **Dark red/black blobs**: L1 SH irradiance can ring negative; clamp to `max(0, ...)` in both the visible gather and the raytrace multi-bounce so negative energy never subtracts light or feeds back into the cache.
- **Degenerate/garbage surfels**: reject non-finite face normals at creation, skip load-failed surfels in binning (kept grid counting and binning in sync, fixing cell corruption), and reject non-finite surfels in the gather.
- **Inf/NaN poisoning**: clamp ray radiance before packing so a sky/sun hit can't unpack as `Inf` and blow up the temporal estimator (`Inf - Inf = NaN`).
- **Colored blobs on viewport resize**: clear the ping-pong temporal history on the first frame after (re)creation. The reallocated textures were uninitialized and their finite garbage passed the reproject NaN/Inf guard.
- Recycle-counter repair, stop recycling still-visible surfels, persist off-screen surfels while the pool has headroom, and several flicker fixes (unwritten spawn-candidate pixels, ray-starved surfels).

## Known limitation / future work

**Mesh LOD transitions drop surfels.** Surfels are anchored to a `primitiveID` + barycentric, and `surface.load` fails when the underlying geometry is swapped by an LOD change (the same path as streamed-out/stale geometry). A surfel whose triangle no longer exists is treated as a load failure, zeroed and recycled, so **an LOD swap discards all surfels on that mesh and the GI has to re-converge** over the following frames. This is **not addressed in this PR** and is called out as a recommended future improvement: re-anchoring or re-projecting surfels across an LOD transition (or matching surfels to the new LOD's surface) would avoid the convergence hit.

----------------

**F6 toggle to pause/resume surfel spawning for debugging will be removed after PR is reviewed.**

**Only tested on Linux**


turanszkij (dismissed):
It looks great

Anthony-Gaudino:
Removed the F6 surfel debug.

turanszkij (approved):

PR #1674, opened 2026-07-08
DDGI: camera-centered multi-cascade clipmap with probe relocation and scalable ray budgeting

Anthony-Gaudino:
## Summary

Reworks DDGI from a single grid fitted to the scene AABB into a **camera-centered, toroidally-scrolling, N-cascade clipmap**. This decouples *probe density* (fine, near the camera) from *coverage* (extended outward by cascades), so GI works in both small scenes and large multi-km levels without the old failure modes: coarse probe density in big scenes, probes drifting through geometry when the scene bounds changed.

On top of the new grid it adds a **probe validity/relocation system** (so probes cope with enclosed and moving geometry), and a set of **performance/scalability** changes that keep 6 cascades affordable. The whole DDGI update chain already runs on the async compute queue.

## 1. Grid architecture

- **Camera-centered grid with toroidal scrolling.** The probe grid follows the main camera; when the camera crosses a cell boundary the grid snaps by whole cells and the buffers are addressed toroidally (a clipmap), so only the newly-entered slab of probes is reset while the rest keep their accumulated light. Signed modular addressing keeps this correct for any grid dimensions and scroll sign.
- **N concentric cascades** (default 6). Cascade 0 is the fine near-camera grid; each coarser cascade keeps the same probe count but doubles cell spacing and coverage. Probes live in one shared set of buffers (concatenated by `probe_offset`); sampling picks the finest cascade containing the point and blends into the next coarser one over the outer edge for a seamless density transition. Per-cascade teleport reset (each cascade has its own reset flag) avoids wiping other cascades' history.
- **Shader-interop refactor.** `ShaderInterop_DDGI.h` is now a leaf data header; `GetScene()`-based helpers moved to a new `ddgiHF.hlsli`. Debug rendering uses a dedicated DDGI-only low-poly icosphere header instead of the shared one.

## 2. Probe validity & relocation

- **Buried-probe detection.** Probes enclosed in solid geometry are detected from backface/escape ray ratios, temporally smoothed into a confidence value with a dead-band to stop the per-frame verdict blinking; invalid probes are down-weighted when sampled and hidden in the debug view.
- **Ejection.** Relocation now pushes a probe *out through the nearest face* when it is behind a surface, instead of deeper in, so a probe a moving object sweeps over is pushed back into open space rather than getting stuck dark inside.
- **Neighbor seeding + squeeze damping.** A buried probe seeds its color from the neighbor in its ejection direction (avoiding brighter edge probes), and a probe pinched between two near surfaces is damped so it stops flickering.
- **Over-bright transient mitigation.** Just-ejected probes stay down-weighted until their own rays land, and probes bordering a just-ejected one fast-reconverge, gated on ejection events, which only occur where a solid is moving, so static scenes are unaffected.

## 3. Performance & scalability

- **Staggered cascade updates.** Cascade 0 updates every frame; coarser cascades round-robin one per frame, so per-frame trace/update cost is ~constant regardless of cascade count.
- **Ray-memory decoupled from cascade count.** Rays are compacted into transient buffers sized for the probes refreshed in a single frame, not for every probe of every cascade, making ray VRAM independent of `DDGI_CASCADE_COUNT`.
- **Ray-budget shaping.** Variance-driven per-probe budget, with buried and open-sky probes capped to a small fixed count, and a **per-cascade budget** (coarser cascades get fewer rays) that trims trace cost and shrinks the ray buffers a further ~25%.
- **Per-active-cascade dispatch.** Ray allocation / update / update-depth dispatch only the two active cascades instead of all probes with an early-out, cutting ~2/3 of the group launches at 6 cascades.
- **Single-gather sampling** deep inside a cascade (only the boundary regions gather two cascades), and **frustum + distance culling** of debug probes.

## 4. Convergence quality

- **Blend-rate compensation** for the staggered schedule: a cascade refreshed 1/K as often scales its temporal blend so it converges at the same wall-clock rate as cascade 0, fixing directional lag/darkening and a cascade-boundary pop while moving (color and depth tuned separately to avoid shimmer).

## 5. Cleanup

- Stops serializing the DDGI probe/depth blobs, which the runtime always rebuilt on load, smaller scene files, with the stream layout kept compatible.

## Known limitation

Fast-moving occluders still show a brief GI transient on their newly-revealed faces.

-------

**Only tested on Linux**


turanszkij:
Didn't spend a lot of time to check yet, but few things I noticed:
- it is much slower than the old one
- if you discard depth data on serialize then it will not be useable as a static DDGI because depth needs to be rebuilt at runtime

I'd like DDGI to remain a more light-weight solution for fallback effects and especially useable as completely baked. I think instead of cascades it should be supplemented with a good screen space probe GI. SSGI rewrite for probe-based one is planned, however there could be a version of that with screen-probes + raytrace.

Anthony-Gaudino:
It's not slower on my system. Reducing the number of cascades would probably help, 6 cascades create a very large number of probes, enough to cover massive worlds. Using 3 or 4 cascades instead would likely provide a significant performance improvement.

You're also right that baking wouldn't work with the current cascade implementation, since the probe density is tied to the camera position. My goal was to move toward something similar to [[UE 5.8](https://www.youtube.com/watch?v=pJhWyxVBMX8)](https://www.youtube.com/watch?v=pJhWyxVBMX8): a fully dynamic solution. However, my implementation isn't complete yet, and I think there's still plenty of room for improvement. From what I can tell, Unreal uses clipmaps rather than cascades, and it only places or activates probes around nearby geometry, which greatly improves performance.

If the goal is baked DDGI, I think a better approach would be to let users define DDGI volumes and configure the DDGI settings independently for each volume. That would allow, for example, placing high-density volumes in interior rooms while using different settings elsewhere. The current approach of fitting a single DDGI volume to the entire scene AABB has a few drawbacks:

1. The maximum probe count is specified as X, Y, and Z dimensions in the UI. For large scenes, this often results in probe densities that are either too low to provide good lighting or unnecessarily high.
2. With procedural terrain, if automatic tile generation is enabled, moving through the world continuously expands the scene AABB as new tiles are created. This shifts the entire probe grid, causing all probes to move, which is undesirable.

PR #1676, opened 2026-07-10
Surfel GI: world-space gather for water and secondary cameras

Anthony-Gaudino:
## Summary

Surfel GI is applied as a screen-space result texture keyed to the **opaque pixels of the main camera**. That leaves two kinds of surface unable to read it, so they silently fell back to the flat weather ambient, staying lit while opaque geometry under surfel GI went dark in a scene with no light:

- **Transparent surfaces (water):** not present in the opaque G-buffer the coverage pass runs on, so the screen-space GI texture cannot be sampled for them. Result: lit ocean foam with no light source.
- **Opaque geometry in the planar reflection / refraction renders:** surfel coverage only runs for the main view, so these secondary cameras have no GI texture. Result: a visible, lit water reflection with no light source.

This is unlike DDGI, whose world-space probe volume is sampled in every camera on every surface, which is why DDGI stayed consistent (everything dark with no light, everything lit with a light) while surfel GI did not.

## Change

Add **`SampleSurfelGI(P, N)`**: a read-only, world-space gather of the surfel cache at an arbitrary shading point, using the same geometry weighting and `/PI` normalization as the coverage pass (the moment/visibility term is omitted, its texture isn't bound outside the surfel passes, and it's a minor anti-leak for these receivers). It is sampled the same way DDGI samples its probe volume.

Wired into `shadingHF.hlsli`:

- **Transparent path:** gather the cache directly at the water surface, before the DDGI block, so when both are enabled transparents use surfel GI (matching opaque).
- **Opaque, secondary cameras:** when the screen-space GI texture is absent (`texture_surfelgi_index < 0`, i.e. reflection/refraction), fall back to the world-space gather instead of flat ambient. **The primary-camera screen-space fast path is unchanged.**

To reach the cache from the forward pass, it is exposed the way DDGI exposes its probe buffer:

- `ShaderScene` gains a `SurfelGI` sub-struct of bindless SRV indices (surfel / grid / cell buffers), populated in `wiScene` (`-1` when surfel GI is inactive).
- Two typed bindless arrays (`Surfel`, `SurfelGridCell`) declared alongside the surfel structs; the cell-index list reuses `bindless_structured_uint`. The Windows-10 root signature gains SRV spaces 211/212 (the SM6.6 heap-directly-indexed signature needs no change).

## Result

- **No light:** water foam and its reflection now go dark, consistent with the dark main-view terrain, matching DDGI.
- **With a light:** the surfels around/under the water pick up the bounce, so water and its reflection receive real indirect GI plus direct light.
- Surfel GI no longer needs DDGI enabled alongside it to get a consistent image.

Water never carries surfels itself (placement already skips transparents and it isn't in the coverage input); it only *reads* the surrounding opaque-surface cache. Open water far from geometry returns no diffuse GI, as expected, its look comes from reflection/refraction and direct light.

## Notes / limitations

- Adds a world-space gather per transparent pixel and per opaque pixel in the reflection/refraction passes (only when surfel GI is enabled; those passes are reduced-resolution, same cost model as DDGI).
- No cross-pixel load sharing yet: unlike the coverage compute pass's cooperative LDS load, the pixel-shader gather re-fetches surfels per pixel. Fine for the water footprint; not intended to replace the opaque screen-space path. A future optimization could use wave intrinsics to dedupe loads.


turanszkij:
The surfel GI on transparents has a logical issue, that the transparents can't spawn surfels themselves, so they will not have correct surfels on their own surface so I think it's not a correct solution. In practice this looks perhaps better than completely disabled surfel GI on transparents. I think for transparents only the world-space probes should be used (DDGI). (Or there is voxel GI which is too slow for real world usage)

turanszkij:
Maybe it's still better to have some kind of les accurate gi for transparents, also this makes it available in envprobes too.

WickedEngine/shaders/ShaderInterop_SurfelGI.h
```diff
@@ -229,6 +229,26 @@ struct SurfelDebugPushConstants
...
+// 211/212, which the default root signature declares.
+#if defined(__PSSL__) || (defined(__hlsl_dx_compiler) && !defined(__spirv__) && __SHADER_TARGET_MAJOR >= 6 && __SHADER_TARGET_MINOR >= 6)
+static const BindlessResource<StructuredBuffer<Surfel> > bindless_structured_surfel;
+static const BindlessResource<StructuredBuffer<SurfelGridCell> > bindless_structured_surfelgridcell;
+#elif defined(__spirv__)
+[[vk::binding(0, DESCRIPTOR_SET_BINDLESS_STORAGE_BUFFER)]] StructuredBuffer<Surfel> bindless_structured_surfel[];
+[[vk::binding(0, DESCRIPTOR_SET_BINDLESS_STORAGE_BUFFER)]] StructuredBuffer<SurfelGridCell> bindless_structured_surfelgridcell[];
+#else
+StructuredBuffer<Surfel> bindless_structured_surfel[] : register(space211);
+StructuredBuffer<SurfelGridCell> bindless_structured_surfelgridcell[] : register(space212);
```
turanszkij: move the resource declarations to globals.hlsli please where all the others are.

turanszkij (changes requested):

WickedEngine/shaders/ShaderInterop_SurfelGI.h
Anthony-Gaudino: https://github.com/turanszkij/WickedEngine/pull/1676/commits/0460caf4aff77223356990bf29e11900b3af30d3 should addresses your request.

turanszkij: Thanks

PR #1683, opened 2026-07-23
updated shader compiler

turanszkij:


brakhane:
@turanszkij Does that mean that #1576 can be closed?

turanszkij:
Yes, I didn't remember that was still active, closed.

PR #1685, opened 2026-07-23
Fix/stable instance slots

Anthony-Gaudino:
Decouple GPU instance slots from ComponentManager array indices to fix surfel GI churn during terrain streaming, and add lazy compaction to reclaim per-frame overhead when objects are shed.

### Core fix: stable instance slots

Surfels blink out and re-converge every time terrain streams (adds/removes props) because surfel GI caches `instanceIndex`, validated by entity `uid`. Terrain removal via swap-with-last relocates objects, changing their index, so cached surfels point at wrong entities and die.

Now objects get persistent GPU instance slots that stay bound for their lifetime. Surfel caches survive unrelated object add/remove churn. Slots live at sparse positions in `[0, objectInstanceCapacity)` with gaps (invalid, `uid=0`); hairs/emitters/impostor/rain append after. Reconciliation before the parallel object job marks, sweeps, and assigns slots,  microseconds in steady state.

### Lazy incremental compaction

Capacity never shrinks (design choice for maximal stability). A scene that sheds a large object peak would pay to zero/process wasted trailing slots every frame. Incremental compaction reclaims them: two-pointer relocation moves top-slot objects into bottom holes, then trim. Conservative tuning (only when waste > 2× live count and ≥1024; max 128 relocations/frame) keeps churn tiny and spread out.

### DIFF
--- a/WickedEngine/shaders/impostor_prepareCS.hlsl
+++ b/WickedEngine/shaders/impostor_prepareCS.hlsl
@@ -29,6 +29,11 @@ void main(uint3 DTid : SV_DispatchThreadID)
 	if (DTid.x >= push.count)
 		return;
 
+	// The thread index is a stable instance slot, which spans the whole [0,
+	// objectInstanceCapacity) range and can contain gaps (slots whose object
+	// was removed). A gap is initialized invalid (geometryCount == 0); skip it
+	// so it is not mistaken for a real impostor object (load_geometry below
+	// would otherwise read geometry 0's slice offset for it).
 	const uint instanceIndex = DTid.x;
 	ShaderMeshInstance instance = load_instance(instanceIndex);
 	ShaderGeometry geometry = load_geometry(instance.geometryOffset);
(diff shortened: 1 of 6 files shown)


turanszkij:
I don't want to change scene representation just for surfel GI, this is too intrusive.

Anthony-Gaudino:
> I don't want to change scene representation just for surfel GI, this is too intrusive.

This PR isn't a scene representation change, the CPU-side ECS stays exactly as-is (swap-remove, component indices, everything). Only the **GPU-facing instance slot** becomes stable via a lightweight free-list allocator. The change is purely at the handful of sites that hand indices to the GPU.

The real problem is broader than surfels, I just happened to find this issue while testing surfels.
Today `instanceIndex == component array position`, which swap-remove reshuffles on any object removal. That's fine per-frame, but breaks the instant anything caches GPU state keyed to a scene object across frames. The engine already admits this: `preload_internal` carries a `uid_validate` check ([surfaceHF.hlsli:355](https://github.com/turanszkij/WickedEngine/blob/3a800b7134aafe58461093c8abb2e274d4e64033/WickedEngine/shaders/surfaceHF.hlsli#L355)) because indices are known unstable. But validation only converts silent corruption into **discarded work**.

Who this affects:

* **Surfel GI:** terrain streaming swap-relocates near objects; every relocated object's cached surfels fail validation and re-converge. Pure wasted recomputation + visible boiling.
* **Any future temporal/world-space cache:** ReSTIR reservoirs, radiance caches, temporal denoisers. They all use the same `PrimitiveID→load_instance` path and will hit the exact same fragility the instant they store an index across frames.
* **Silent correctness bugs:** anyone forgetting the `uid` guard gets the wrong geometry, not a clean miss.

Streaming scenes pay continuous re-convergence cost today. Stable slots cost microseconds reconciliation + allocator overhead (negligible in steady state, an improvement when objects shed).

turanszkij:
My design was intentionally to make the gpu scene have the same layout as the cpu scene so this modifies that base behaviour that all rendering uses, just for the surfel GI which is just an optional feature (honestly it was experimental feature, I didn't like it in the end). It also requires that additional book keeping in the scene that iterates multiple times through objects on single thread.  don't want to make this change.

PR #1689, opened 2026-08-10
Refactor, Harden, and Document CommonInclude.h

Anthony-Gaudino:
### Summary
Comprehensive refactor of `WickedEngine/CommonInclude.h`.

---

#### Bug Fixes & Correctness
- **StackString::size()**: Fixed to return `cnt` (active length) instead of `Capacity`
- **arraysize macro**: Added parentheses around `__VA_ARGS__` for correct precedence
- **align()**: Hybrid bitwise/division for arbitrary alignment (not just power-of-2)
- **is_aligned()**: Bitwise check for power-of-2, division fallback for arbitrary
- **StackVector::clear()**: Only zeros active elements (`last` items), not full capacity
- **StackVector::resize → set_size**: Renamed to avoid `std::vector` confusion
- **countbits(unsigned long)**: Return type fixed to `unsigned long`
- **lerp/bilinear**: Return component type `T` instead of vector type
- **firstbithigh/firstbitlow**: Return bit width (32/64) for input 0 (distinguishes from bit 0)
- **bilinear**: Template params `float4/float2` → generic `Vec4/Vec2` with SFINAE constraints
- **set_flag**: Removed redundant `const` on flag parameter
- **NOMINMAX**: Wrapped in `_WIN32` guard
- **StackString::push_back**: Added nullptr check
- **StackVector**: Added bounds checking asserts in debug builds
- **inverse_lerp**: Zero-range fallback documented (returns 0 for degenerate cases)
- **frac()**: Constrained to floating-point via `static_assert`, uses `std::floor`
- **has_flag**: Constrained to bitmask-enabled types via `enable_bitmask_operators`

#### API Additions
- **capacity()**: Added to `StackString` and `StackVector`
- **StackVector::at()**: Bounds-checked accessor
- **StackString::operator+=**: For `char` and `const char*`
- **to_stack_string()**: Replaces `extract_function_name()`, constexpr construct from literal
- **relative_path_storage()**: Owned copy alternative to `relative_path()`

#### Removals & Deduplication
- Removed unused includes

#### Attributes & Safety
- Added `[[nodiscard]]` and `noexcept` for hardening

#### Documentation
- Fully documented the file

#### Call Site Updates
- `wiGraphicsDevice_Vulkan.h`
- `wiGraphicsDevice_DX12.h`
- `wiAudio.cpp`
- Updated to use `to_stack_string()` instead of removed `extract_function_name()`

### DIFF
--- a/WickedEngine/CommonInclude.h
+++ b/WickedEngine/CommonInclude.h
@@ -5,6 +5,7 @@
 
 #include <cassert>
 #include <cfloat>
+#include <cmath>
 #include <cstdint>
 #include <type_traits>
 
@@ -66,9 +67,8 @@ constexpr T saturate(T x)
 template <typename T>
 constexpr T frac(T x)
 {
-	T f = x - T(int(x));
-	f = f < 0 ? (1 + f) : f;
-	return f;
+	static_assert(std::is_floating_point_v<T>, "frac only supports floating-point types");
+	return x - std::floor(x);
 }
 
 template <typename T>
--- a/WickedEngine/CommonInclude.h
+++ b/WickedEngine/CommonInclude.h
@@ -206,7 +206,7 @@ struct StackVector
 	{
 		return 31u - (unsigned int)bit_index;
 	}
-	return 0;
+	return 32u;
 }
 [[nodiscard]] inline unsigned long firstbithigh(unsigned long value) noexcept
 {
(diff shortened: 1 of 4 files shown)


turanszkij:
Please discuss with me before refactoring, I prefer to not have these changes.

PR #1700, opened 2026-09-12
Add lifecycle support for custom shaders

MonkeyFrogStudio:
Adds stable-ID update and unregister operations to Wicked Engine’s CustomShader registry.

Active storage remains compact, stale IDs never alias later registrations, render lookup retains PipelineState ownership safely, and invalid lifecycle operations fail without disturbing the active shader.

Internal renderer, material, sorting, and editor paths now treat custom-shader IDs as opaque handles.

Includes a focused CTest covering registration, PSO replacement, invalid operations, unregister behavior, retained copies, stale-ID safety, and 512 bounded-storage reload cycles.

The change is intended to support tools that need to recompile and replace custom shaders repeatedly at runtime while retaining a stable customShaderID.


WickedEngine/wiRenderer.cpp
```diff
@@ -3251,9 +3313,8 @@ void RenderMeshes(
 						break;
 					}
 				}
-				else if (material.customShaderID >= 0 && material.customShaderID < (int)customShaders.size())
+				else if (material.customShaderID >= 0 && GetCustomShader(material.customShaderID, customShader))
```
turanszkij: This new GetCustomShader does a mutex lock, but this is the RenderMeshes function that gets called inside render frame. This must not use mutex locking inside here. Have to ensure somehow else that at this point there will be no updates to custom shaders instead. This should be fine to just say that custom shaders must be registered on main thread while rendering is not happening, and just not use mutexes in it.

turanszkij (changes requested):

turanszkij (approved):

